PR #64, opened 2021-02-11
Rewrite ekomark using banana

giacomomagni:


benchmarks/ekomark/plots.py
```diff
@@ -71,7 +71,7 @@ def plot_dist(x, y, yerr, yref, title=None, oMx_min=1e-2, oMx_max=0.5):
     ax3 = plt.subplot(2, 3, 3)
     ax3.set_xscale("log", nonpositive="clip")
     ax3.set_yscale("log", nonpositive="clip")
-    oMx = 1.0 - x
+    oMx = [1.0 - i for i in x]
```
AleCandido, line 74: Actually it should work even as it was before, if `x` is in array (and so `oMx` will be as well). If for some reason it happens not to be an array you can add somewhere:
```py
x = np.array(x)
```

benchmarks/setup.py
```diff
@@ -4,12 +4,12 @@
 
 setup(
     name="ekomark",
-    author="F. Hekhorn, A.Candido",
+    author="F. Hekhorn, A.Candido, G.Magni",
```
AleCandido, line 7: Welcome :D

benchmarks/runners/sandbox.py
```diff
@@ -0,0 +1,81 @@
...
+
+if __name__ == "__main__":
+
+    # activate logging
+    logStdout = logging.StreamHandler(sys.stdout)
+    logStdout.setLevel(logging.INFO)
+    logStdout.setFormatter(logging.Formatter("%(message)s"))
+    logging.getLogger("eko").handlers = []
+    logging.getLogger("eko").addHandler(logStdout)
+    logging.getLogger("eko").setLevel(logging.INFO)
```
AleCandido, line 78: All these logging stuffs you can also keep on the top of the file. But since it is repeated in the exact same way on all the runners can it be lifted? or at least collected into a function/module inside ekomark, to invoke with a single function call?

AleCandido (commented):
Actually I had just a quick look, but I quite recognized the yadmark structure, so I believe you are enough close.
Have you tried by chance to run any runner fully with pytest? In case: how much did it take?

benchmarks/ekomark/benchmark/external/apfel_utils.py
```diff
@@ -0,0 +1,231 @@
...
+        apfel.SetExternalGrid(1, len(yad_xgrid) - 1, yad_deg, xgrid)
+
+    # set DIS params
+    apfel.SetPDFSet(pdf)
+    # set Target
+
+    # init evolution
+    apfel.InitializeAPFEL()
+
+    return apfel
```
AleCandido, line 154: This function it's more or less the same between `ekomark` and `yadmark`: should we lift at least part of it in banana? @felixhekhorn

benchmarks/ekomark/benchmark/external/lhapdf_utils.py
```diff
@@ -0,0 +1,61 @@
...
+    out_tabs = {}
+    for pid in br.flavor_basis_pids:
+
+        if pid in skip_pdfs:
+            continue
+
+        # collect lhapdf
+        me = []
+        for x in target_xgrid:
+            xf = pdf.xfxQ2(pid, x, operators["Q2grid"][0])
```
AleCandido, line 40: Here you are hardcoding `operators["Q2grid"][0]` as we were doing, but in principle the `[0]` is there because it is a list and may contain more than a single element.
If more than one element is present all of them should be benchmarked, and since `lhapdf` can do it this should be tested (eko it's somehow optimizing for computing multiple `Q2` at once, so the test it's meaningful).

benchmarks/runners/sandbox.py
giacomomagni: True, moved to the runner.

benchmarks/setup.py
giacomomagni: 👍

benchmarks/ekomark/plots.py
giacomomagni: Fixed

benchmarks/ekomark/benchmark/external/lhapdf_utils.py
giacomomagni: Yes, same for apfel, when everything is fine with one q2, I'll add a loop.

AleCandido: :ok_hand:

benchmarks/ekomark/benchmark/external/apfel_utils.py
giacomomagni: see https://github.com/N3PDF/banana/pull/13

AleCandido:
I merged #65 in here. Since I believe it's actually working I would like to merge also this in `develop`.

Please the two of you @felixhekhorn and @giacomomagni have a look to the content and experiment a bit if you find something not working (you should use the `develop` version of banana).

In particular we should wait @giacomomagni for the wiki page report on failing (or not so good) benchmarks.

giacomomagni:
Just tested, it's working. 
I see codefactor is finding some issues, but not sure if it's necessary to solve them now... 
I'm writing the summary of apfel bench [here](https://github.com/N3PDF/eko/wiki/Benchmark-against-Apfel)

AleCandido:
> Just tested, it's working.
> I see codefactor is finding some issues, but not sure if it's necessary to solve them now...

I noticed them as well (the only red cross over a green ticks field), and I would have liked to turn it to green tick as well, but they are mostly stupid, so I'm postponing. A small review:
- the TODOs have to stay there
- the unused argument are there for compatibility, if not required even by yadmark we can even remove them, otherwise just prepend an underscore
- the tuples assignment I don't know, for me it's fine to keep
- `skip_pdf` maybe it's the only one that deserve a review

> I'm writing the summary of apfel bench [here](https://github.com/N3PDF/eko/wiki/Benchmark-against-Apfel)

I had a look, that's perfect, thanks :)

benchmarks/runners/apfel_bench.py
```diff
@@ -122,12 +114,7 @@ def benchmark_ic(self):
 
         th = self.ffns_theory.copy()
         th.update(
-            {
-                "PTO": [1],
-                "IC": [1],
-                "mc": [1.4142135623730951, 2.0],
-                "Qmc": [1.4142135623730951, 2.0],
-            }
+            {"PTO": [1], "IC": [1], "mc": [2.0], "Qmc": [2.0],}
```
felixhekhorn, line 117: - ``Qmc`` should not be used
- if ``IC`` is working then it is because it doesn't have an effect, so maybe we should remove it as well

felixhekhorn (changes requested):

AleCandido:
`subtract_tables` removed, it was actually quite copy-pasted from `yadmark`, just supporting the `dict(q2: DFdict)` layout that is not used any longer.

AleCandido:
Now we have a green tick, but only solving minor stuffs. Still to be done:
- this TODO maybe it has to be implemented (or simply removed if it's already been done):
https://github.com/N3PDF/eko/blob/e1390a705245349e682ae19901ee0399367a224b/benchmarks/runners/CT14_bench.py#L23
- I still need to have a look to `self.skip_pdfs`

benchmarks/runners/CT14_bench.py
```diff
@@ -0,0 +1,82 @@
...
+class LHAPDFBenchmark(Runner):
+    """
+    Globally set the external program to LHAPDF
+    """
+
+    external = "LHAPDF"
+
+    # Rotate to evolution basis
+    # rotate_to_evolution_basis = True
+    # TODO: rotate also lhapdf to evbasis
```
AleCandido, line 23: Implement (or remove if it has already been done)

AleCandido (changes requested):

benchmarks/runners/apfel_bench.py
giacomomagni: Okay, yes it's used by apfel but only when HQ is MSBAR

benchmarks/ekomark/navigator/navigator.py
```diff
@@ -0,0 +1,229 @@
...
+        Collect important information of the log record.
+
+        Parameters
+        ----------
+            lg : dict
+                database record
+            obj : dict
+                to be updated pandas record
+        """
+        q2s = lg["log"].q2s
```
felixhekhorn, line 116: @AleCandido do you have the implementation of .q2s still at head? I guess you just didn't commit - or should I reimplement a new? I guess it should be a unique sum of all children; note that I changed the key in the dfdict to give a proper name, so other things might be broken as well

felixhekhorn (changes requested):

benchmarks/ekomark/navigator/navigator.py
AleCandido: It should be in banana/develop:
https://github.com/N3PDF/banana/blob/f134fcb394ca65a4875ac1dfe873e2b7f96226d0/src/banana/data/dfdict.py#L72-L77

Maybe you didn't pull banana, is it possible?

felixhekhorn: you are right

felixhekhorn:
@AleCandido sorry - https://github.com/N3PDF/eko/pull/64/commits/bffdbaf46b918856ac2c166db2c53b17fff6cb3c was a side effect of me not pulling

AleCandido:
> @AleCandido sorry - [bffdbaf](https://github.com/N3PDF/eko/commit/bffdbaf46b918856ac2c166db2c53b17fff6cb3c) was a side effect of me not pulling

I know, don't mind, we are working separately across multiple projects. It's fine to be complicate (people often use monorepos for this reason, but I don't like so much...).

AleCandido (approved):

PR #65, opened 2021-02-25
Introduce ORM backend for ekomark

AleCandido:
Update to the new version of banana (the would-be v0.3.x)


giacomomagni (approved):
Great it looks fine!!

AleCandido:
Did you tested it? I'm going to do it now :)

AleCandido:
Now I tested and it is working, but I had to fix banana in order to obtain the result: currently you are using in ekomark a layout for the logs different from the yadmark one
- **yadmark**: `DFDict(x, Q2, res, err, ext, rel_err)`
- **ekomark**: `dict(muF2, DFDict(z, res, err, ext, rel_err))`

The moment you rename `muF2=Q2` and `z=x` you see that a `Q2` has been lifted into a dictionary.

Before `banana` was more agnostic, and let you pass whatever as a log. It was simply dumping it as `bytes` and turning it back the moment you wish something else.
Now it's more strict: it requires that the log you're passing it's something with a `to_document` method, because it want (in principle) to dump a simple dict (e.g. a one you can dump with json), rather than a complicate python object, and this is what the `to_document` method it's supposed to do (we could even try to enforce, but we simply trust our users... ourselves...).

For the time being a dirty trick it's solving the compatibility (but it cannot stay in banana: it has to remain agnostic on the user!), but I see only two ways for permanently solving the issue:
1. we try to remain more agnostic on the log content, and we simply make a **new object** in ekomark that wraps your dict, with a `to_document` method that encapsulates the content of the dirty trick (that is dirty only inside banana, in ekomark may perfectly fit)
2. we construct a **unique** `DFDict(z, muF2, res, err, ext, rel_err)` even in ekomark, splitting it in the navigator if required (in this way we have the **further benefit** of being able to get a slice in muF2 at given z, simply by filtering the `pandas.Dataframe`)

AleCandido:
If `pandas.DataFrame` it's not the proper object to contain a 3d table (with double index on `(x, Q2)`) we can move to an [`xarray`](http://xarray.pydata.org/en/stable/), that is made actually for this goal.

giacomomagni:
> For the time being a dirty trick it's solving the compatibility (but it cannot stay in banana: it has to remain agnostic on the user!), but I see only two ways for permanently solving the issue:
> 
> 1. we try to remain more agnostic on the log content, and we simply make a **new object** in ekomark that wraps your dict, with a `to_document` method that encapsulates the content of the dirty trick (that is dirty only inside banana, in ekomark may perfectly fit)
> 2. we construct a **unique** `DFDict(z, muF2, res, err, ext, rel_err)` even in ekomark, splitting it in the navigator if required (in this way we have the **further benefit** of being able to get a slice in muF2 at given z, simply by filtering the `pandas.Dataframe`)

Hi Alessandro! I would go for the second option, I feel that this way will be easier and more consistent with yadmark. There might be also the possibility to lift some pieces of code to banana/navigator...

PR #66, opened 2021-03-09
NNLO anomalous dims, EXA EXP TRN and SV.

felixhekhorn:


tests/test_ad_nnlo.py
```diff
@@ -25,7 +25,7 @@ def test_gamma_2():
     sx = np.append(sx, harmonics.harmonic_S4(2))
     gS2 = ad_nnlo.gamma_singlet_2(2, NF, sx)
     # gluon momentum conservation
-    np.testing.assert_allclose(gS2[0, 1] + gS2[1, 1], 0, atol=9e-2)
+    np.testing.assert_allclose(gS2[0, 1] + gS2[1, 1], 0, atol=4e-3)
```
AleCandido, line 28: Still `1e-3` it's a bit too much... don't you think so?

giacomomagni: Hi Ale, I checked all the ad with mathematica and now I'm quite sure they all agree with the fortran code and the tests reach the same precision on the fortran code.

giacomomagni: The difference is proportional to NF or ( NF^2 ) so if you use NF=2 it's around 1e-4

felixhekhorn: - did you check this contribution specifically?
- otherwise I'm afraid the story from NLO continues https://github.com/N3PDF/eko/blob/cc248184e7555ec75bc073cf4a561d8569246899/tests/test_ad_nlo.py#L18

giacomomagni: What do you mean? Here CA is encoded inside the numbers...
From mathematica and referring to the fortran code I get:
```
GG1 + QG1 = 0.000676541  (* prop to nf^1 *)
GG2 + QG2 = 0.0000561758 (*prop to nf^2 *)
GG0 = -0.0008998329547011963 (*prop tp nf ^0*)
```

AleCandido: I believe that the problem it's not CA, but the term whose coefficient is CA...
What Felix said is that that term it's hard to integrate, the convergence it's numerically difficult, thus he lifted a little bit the constraint. But he lifted `2e-6 -> 4e-5`, instead `4e-3` it's completely out by 2000 times, hard to believe it's just integration error, even if tough to converge.

felixhekhorn: Not "integration error" but evaluation error - otherwise yes!

src/eko/kernels/singlet.py
```diff
@@ -540,18 +540,21 @@ def eko_perturbative(gamma_singlet, a1, a0, nf, ev_op_iterations, ev_op_max_orde
...
-            R vector
+        is_exact : boolean
+            fill up r-vector?
 
     Returns
     -------
         e_s^1,2 : numpy.ndarray
             singlet NLO or NNLO order perturbative EKO
     """
+    r = r = r_vec(gamma_singlet, nf, ev_op_max_order, pto, is_exact)
```
felixhekhorn, line 557: A typo I guess?

giacomomagni: yes, thanks!

src/eko/anomalous_dimensions/nnlo.py
```diff
@@ -0,0 +1,593 @@
...
+    E1 = S1 / n ** 2 + (S2 - zeta2) / n
+    E2 = 2.0 * (-S1 / n ** 3 + (zeta2 - S2) / n ** 2 - (S3 - zeta3) / n)
+    B11 = -(S1 + 1.0 / (n + 1.0)) / (n + 1.0)
+    B12 = -(S1 + 1.0 / (n + 1.0) + 1.0 / (n + 2.0)) / (n + 2.0)
+
+    # with special care for the first moment of x^-1 ln(1-x)
+    if abs(n.imag) < 1.0e-5 and abs(n - 1.0) < 1.0e-5:
+        B1M = -zeta2
+    else:
+        B1M = -(S1 - 1.0 / n) / (n - 1.0)  # pragma: no cover
```
felixhekhorn, line 209: To fix this: maybe can you upload your MMa script for the NNLO benchmark in the wiki (for lack of a better place at the moment) and we can add an reference value for N=2?

giacomomagni: Okay I can add such test!

felixhekhorn:
Maybe can you improve the documentation in `src/eko/kernels/evolution_integrals.py` ?
- make the notation consistent with the DGLAP section
- put `_{exp}` where it belongs
- make `_{exp} = ` instead of `\approx` (because they are what they are)
- maybe repeat the `_{exp}` definition on top of the module
- check the formula layout

felixhekhorn:
@giacomomagni good you noticed 6c32615 and you fixed it (although it is for sure not a "minor fix" ;-) ) - even better to write in addition a test against that -> 3b83a26

giacomomagni:
> 3b83a26

Great, thanks for the additional test!! That happened because I moved the repeated calculation of the NLO correction only in the first condition.

giacomomagni (approved):

benchmarks/ekomark/benchmark/external/LHA_utils.py
```diff
@@ -19,10 +19,16 @@
 toy_xgrid = np.array([1e-7, 1e-6, 1e-5, 1e-4, 1e-3, 1e-2, 0.1, 0.3, 0.5, 0.7, 0.9])
 
 # list
-raw_label_list = ["u_v", "d_v", "L_m", "L_p", "s_p", "c_p", "b_p", "g"]
+raw_label_list = {
+    "default": ["u_v", "d_v", "L_m", "L_p", "s_p", "c_p", "b_p", "g"],
+    "NNLO_FFNS": ["u_v", "d_v", "L_m", "L_p", "s_v", "s_p", "c_p", "g"],
```
felixhekhorn, line 24: - I don't like the repetition in the code ... 
- I'd rather either repeat some zeros in the data file (as I've been already doing with c_+ and b_+)
- or adjust the data in `compute_LHA_data` where all informations are present

benchmarks/runners/paper_lha_bench.py
```diff
@@ -125,24 +136,23 @@ class BenchmarkFFNS(LHABenchmark):
...
-    is_FFNS = True
 
-    def ffns_nnlo_pdfs(self):
-        """
-        Set different pdfs set for NNLO FFNS
-        """
-        self.skip_pdfs.extend([-5, 5, "T24"])
-        if "V8" in self.skip_pdfs:
-            self.skip_pdfs.remove("V8")
+    def get_skip_pdfs(self, theory_updates):
```
AleCandido, line 140: In general naming things `get_*` it's not really pythonic, since usually they are used in other languages (like Java) for accessing attributes, while Python uses `@property` to virtualize an attribute.

In this case you don't have a virtual attribute but really a function, nevertheless you could wish to avoid the `get_*` pattern, and the easiest way it's simply to drop `get_` and retain `*` (`skip_pdfs` in this case) since `get_` it's not adding anything.
It's really a minor aesthetic issue in this case.

felixhekhorn: yes, I know - Problem is, we can not use just `skip_pdf`, I believe, since the attribute is already called this way and I need to pass directly some parameters (i.e. theory) to it

src/eko/kernels/evolution_integrals.py
```diff
@@ -143,3 +146,192 @@ def j01_expanded(a1, a0, nf):
...
+    .. math::
+        j^{(2,2)}(a_s,a_s^0) &=
+            \int\limits_{a_s^0}^{a_s}\!da_s'\,\frac{a_s'^3}
+                        {\beta_0 a_s'^2 + \beta_1 a_s'^3 + \beta_2 a_s'^4}
+            = \frac{1}{\beta_2}\ln\left(
+                    \frac{1 + a_s ( b_1 + b_2 a_s ) }{ 1 + a_s^0 ( b_1 + b_2 a_s^0 )}\right)
+                - \frac{b_1 \delta}{ \beta_2 \Delta} \\
+            \delta &= \atan \left( \frac{b_1 + 2 a_s b_2 }{ \Delta} \right)
+                    - \atan \left( \frac{b_1 + 2 a_s^0 b_2 }{ \Delta} \right) \\
+            \Delta &= \sqrt{4 b_2 - b_1^2}
```
felixhekhorn, line 165: If I remember correctly you had a reference for this? Can you please add it (e.g. here and at the top of the module)

giacomomagni: The reference is cited in the docs, I haven't quoted them here because they don't report the result exactly this way, so I thought it was more confusing having it here.

benchmarks/ekomark/benchmark/external/LHA_utils.py
felixhekhorn: Good you addressed this in e764625  but can we not do this more consistently? i.e. also remove this https://github.com/N3PDF/eko/pull/66/commits/e764625564fc5e3c1ef5e1ca024f122c524510c5#diff-552b6e67715034c2e3566ba07603a7f1e15d554ab40879d61393b7b41eb9ab75R45 and this https://github.com/N3PDF/eko/pull/66/commits/e764625564fc5e3c1ef5e1ca024f122c524510c5#diff-552b6e67715034c2e3566ba07603a7f1e15d554ab40879d61393b7b41eb9ab75R80 (as Alessandro already said already, I think ...) - maybe even remove `LHA_rotate_to_evolution` as a whole?

giacomomagni: To remove this function we will need to add the LHA basis to ``eko.basis_rotation`` , would this way be better ?

felixhekhorn: No, not remove `rotate_data` - just the double `LHA_to_flavor` and `LHA_to_evol' since one of them, say `_to_flavor`, should be enough

felixhekhorn: And remove the special treatment of FFNS NNLO if possible ...

PR #67, opened 2021-03-18
Hunting for sv bug

felixhekhorn:


AleCandido:
Second thought: this is the proper place for our notes.

- which is the relevant log?
  - `log(kcThr)` or `log(kcThr/fact_to_ren)`
- the log should be different for each threshold considered
- is the inverse (going down/up) correctly computed
  - since it's taken by apfel code: is it consistent with apfel? (do the benchmark)
  - which is the default direction: down or up?

PR #69, opened 2021-04-08
pegasus benchmark

giacomomagni:


AleCandido:
Hi @giacomomagni unfortunately we have a problem even in starting: we are not able to compile Pegasus, so please provide suitable instructions on how to install it (so to compile the Fortran code, not the wrapper, for which the instructions are present).

Just a sample of the current errors:
```
inppmom2.f:152:30:

   67 |          AUV = NUV * ( EBETA (PUV(2)*E, PUV(3)+E) *
      |                              2
......
  152 |      2               + EBETA (0.5+PSS(2), PSS(3)+E) * (PSS(4) +
      |                              1
Error: Type mismatch between actual argument at (1) and actual argument at (2) (REAL(8)/COMPLEX(8)).
```
We are both trying to compile on Linux with gfortran 10.2

giacomomagni:
> We are both trying to compile on Linux with gfortran 10.2

I see, I was compiling with the "conda" gfortran: GNU Fortran (GCC) 9.3.0 ...

AleCandido:
Fair enough, but `f2py` it's not a fortran compiler on its own, so it make use of the `gfortran` available on your system -> I tried and as expected it's not working the very same way.

giacomomagni:
> Fair enough, but `f2py` it's not a fortran compiler on its own, so it make use of the `gfortran` available on your system -> I tried and as expected it's not working the very same way.

Good point. I tried as well with gfortran 10 and I do get the same issue...  we are probably missing a flag

AleCandido:
In N3PDF/external@659f8de1329ae3fd65abaf8e02f847c10a7c31ef I added the fix:
- essentially it amounts to the flag `-fallow-argument-mismatch`
- very simple `Makefile` added, for independent compilation
- `f2py.sh` fixed accordingly

Last remark `g77` it's just a rather old version of `gfortran`, not maintained any longer since more or less gcc 4 http://www.kilmnj.com/g77/why.html
Actually if the code was working with that one there is still a chance that it will work again, but making `g77` work might be hard.

PR #71, opened 2021-05-11
Backward evolution

giacomomagni:


src/eko/evolution_operator/grid.py
```diff
@@ -233,15 +233,20 @@ def generate(self, q2):
...
-        if len(path) > 1 and len(self.ome_members) == 0:
-            if is_backward is False:
+        if len(path) > 1 and self.ome_members is None:
+            if (
+                self.config["intrinsic_range"] is not None
+                or self.config["backward_inversion"] == "exact"
+            ):
+                ome = OperatorMatrixElement(self.config, self.managers)
+            elif is_backward is False:
                 self.config["backward_inversion"] = None
```
AleCandido, line 246: Changing `self.config` is weird and might be wrong:
- the configs should be the ones set by the user, so you can think about replacing that information by passing it directly as an argument to the function `self.compute_matching_coeffs()`
- in particular this will be an actual bug if your grid contains some forward and some backward Q2s: the first forward Q2 will replace the value for `self.config` forever, and the moment you try to compute a backward one you will not fall in any valid branch but in `ValueError`

src/eko/evolution_operator/grid.py
```diff
@@ -254,14 +259,9 @@ def generate(self, q2):
...
-            if self.config["backward_inversion"] == "exact" and is_backward:
-                if self.config["intrinsic_range"] is None:
-                    # compute inverse exact
-                    ome.compute(a_s)
-                else:
-                    # compute intrinsic inverse exact
-                    ome.compute(a_s, nf=op.nf)
+            # compute the matching if not done yet
+            if self.ome_members is None:
+                ome.compute(a_s)
```
AleCandido, line 264: Since the matching are done at threshold then it would be a good idea to group the computation together with the threshold operator, packing inside the function `get_threshold_operators()`, and in this way calling `ome.compute()` in a single place

src/eko/matching_conditions/__init__.py
```diff
@@ -97,11 +97,11 @@ def split_ad_to_evol_map(
                         {
                             f"{hq}+.S": ome_members["S_Hq"],
                             f"{hq}+.g": ome_members["S_Hg"],
```
AleCandido, line 99: Intrinsic is when the input is intrinsic, so the second one is an H. If the first one is an H it's fine, and it should be present even if there was no intrinsic in the beginning, because it is some light flavor / the gluon that generates the (was-)heavy flavor.

AleCandido: Furthermore `f"T{n}.S"` has to be removed, as we agreed to match in the (g, S, V, ..., c+, c-) -> (g, S, V, ..., c+, c-) basis. Then there will be a `f"{hq}+.S" = ome_members["S_hq"]` there, for example.

tests/test_matching.py
```diff
@@ -23,6 +23,12 @@ def mkOME(self):
         gg = mkOM(self.shape)
         return dict(NS_qq=ns, S_qq=qq, S_qg=qg, S_gq=gq, S_gg=gg)
 
+    def update_intrinsic_OME(self, ome):
+        for key in [ 'Hq', 'Hg', 'HH', 'qH', 'gH']:
```
AleCandido, line 27: Again `Hq` is not intrinsic.

src/eko/evolution_operator/grid.py
```diff
@@ -77,7 +79,8 @@ def __init__(
             interpol_dispatcher=interpol_dispatcher,
         )
         self._threshold_operators = {}
-        self.ome_members = {}
+        self.is_backward = False
```
felixhekhorn, line 82: `is_backward` can't be a attribute of `OperatorGrid` since this is a property of an individual `Operator` instead

felixhekhorn:
https://github.com/N3PDF/eko/blob/47e1f9c22357f4767ac8e4cfe5c99affee15256b/src/eko/matching_conditions/nnlo.py#L274
`A_ns_2` should be added to the qq bit right?

giacomomagni:
> https://github.com/N3PDF/eko/blob/47e1f9c22357f4767ac8e4cfe5c99affee15256b/src/eko/matching_conditions/nnlo.py#L274
> 
> 
> `A_ns_2` should be added to the qq bit right?

I believe it's better non to do it here, since S_Hq should contain only the pure singlet part.

src/eko/evolution_operator/grid.py
giacomomagni: Okay good, I'll fix it.

src/eko/evolution_operator/grid.py
giacomomagni: Okay, I agree.

src/eko/evolution_operator/grid.py
giacomomagni: This is also a good suggestion! Thanks

src/eko/matching_conditions/__init__.py
giacomomagni: sure this is fixed.

src/eko/evolution_operator/flavors.py
```diff
@@ -160,9 +160,13 @@ def rotate_matching(nf, inverse=False):
...
-        q = quark_names[k - 1]
-        for sgn in "+-":
-            l[f"{q}{sgn}.{q}{sgn}"] = 1.0
+        # q = quark_names[k - 1]
+        # for sgn in "+-":
+        #     l[f"{q}{sgn}.{q}{sgn}"] = 1.0
+        n = k ** 2 - 1
+        l[f"V{n}.V"] = 1.0
+        l[f"T{n}.S"] = 1.0
+        l[f"T{n}.g"] = 1.0
```
felixhekhorn, line 169: - this seems wrong: we agreed to do the matching in the basis which does include (e.g.) `c+,c-` but NOT `T15,V15`
- instead https://github.com/N3PDF/eko/blob/7716902548c60646ae6340b12bd27b5800a2e0d3/src/eko/matching_conditions/__init__.py#L60 again here only `c+,c-` should appear, both in the input and the output

src/eko/evolution_operator/grid.py
```diff
@@ -135,7 +134,7 @@ def from_dict(
             config, q2_grid, thresholds_config, strong_coupling, interpol_dispatcher
         )
 
-    def get_threshold_operators(self, path):
+    def get_threshold_operators(self, path, is_backward):
```
felixhekhorn, line 137: I believe this is too much information: `is_backward` could instead be attributed to `PathSegment` - this is the most atomic element (and perfectly aware)

felixhekhorn:
> I believe it's better non to do it here, since S_Hq should contain only the pure singlet part.

Actually why? because you would need to add it *always* - and the consistent way to do it is here

as we're doing here https://github.com/N3PDF/eko/blob/7716902548c60646ae6340b12bd27b5800a2e0d3/src/eko/anomalous_dimensions/nlo.py#L266

and the "singlet" here refers to the full distribution, that has a non-singlet part (actually NS_+) and a pure singlet part

src/eko/matching_conditions/nnlo.py
```diff
@@ -32,8 +34,10 @@ def A_qq_2_ns(n, sx):
     S1 = sx[0]
     S2 = sx[1]
     S3 = sx[2]
+    S1m = harmonic_S1(n - 1)
```
felixhekhorn, line 37: of course S1 and S1m are related ...

src/eko/matching_conditions/nnlo.py
```diff
@@ -50,12 +54,29 @@ def A_qq_2_ns(n, sx):
...
-    return constants.CF * constants.TR * a_qq_2
+    a_qq_2_l1 = (
+        2
+        * (
+            -12
+            - 28 * n
+            + 9 * n ** 2
+            + 34 * n ** 3
+            - 3 * n ** 4
+            + 40 * (n * (n + 1)) ** 2 * S1m
```
felixhekhorn, line 65: I would collect by S1: this simplifies the prefactor a lot and I think it is numerically good since S1 itself can grow rapidly (and hence shouldn't be yet more enhanced to avoid numbers with big difference in magnitude)

felixhekhorn: more generally collect by S (or any complicated function)

src/eko/matching_conditions/nnlo.py
```diff
@@ -127,10 +162,14 @@ def A_hg_2(n, sx):
     S1 = sx[0]
     S2 = sx[1]
     S3 = sx[2]
+    S1m = harmonic_S1(n - 1)
+    S2m = harmonic_S2(n - 1)
+    S2mhalf = harmonic_S2((n - 1) / 2)
```
felixhekhorn, line 167: - if you wish you can call this `Sp2m` https://github.com/N3PDF/eko/blob/feature/backward_evol/src/eko/anomalous_dimensions/nlo.py#L41 and `Sp2p`
- we should finally start caching the S-es also in `ad.nlo` and then reuse them here

src/eko/matching_conditions/nnlo.py
AleCandido: I perfectly agree if you are passing by Mathematica in any case, if instead you don't have Mathematica in your stack already think if it's really so expensive: the less we process the expressions the less we have to debug them.

tests/test_matching_nnlo.py
```diff
@@ -21,20 +21,24 @@ def get_sx(N):
 
 def test_A_2():
     N = 1
+    L = 100
```
felixhekhorn, line 24: I would suggest `for L in [0,100]:`

src/eko/matching_conditions/nnlo.py
giacomomagni: Yes, I agree, we can simplify this easily.

src/eko/matching_conditions/nnlo.py
felixhekhorn: >     we should finally start caching the S-es also in ad.nlo and then reuse them here

actually that's a bit stupid, because `ad` and `matching` are never computed in the same run and thus could share S-es ...

felixhekhorn:
@giacomomagni  I wrote a bit on the matching section - if you want you can continue there (as discussed: split the things and if necessary rearrange everything) or else I will try to do it some time ...

src/eko/matching_conditions/operator_matrix_element.py
```diff
@@ -207,7 +207,7 @@ def quad_ker(u, order, mode, is_log, logx, areas, a_s, L, backward_method):
 
     # select the need matrix element
     ker = ker[indeces[mode[-2]], indeces[mode[-1]]]
-    if ker == 0.0:
+    if np.real(ker) == 0.0:
```
felixhekhorn, line 210: that seems wrong ... why do you want to skip at all? `ker` can be purely imaginary and still contribute to the integral ...

src/eko/evolution_operator/grid.py
```diff
@@ -179,7 +179,11 @@ def get_threshold_operators(self, path):
 
             # Compute the matching conditions and store it
             if len(path) > 1:
```
felixhekhorn, line 181: I guess, we should add a `and seg.q2_to not in self._matching_operators` as we did above for caching

src/eko/matching_conditions/operator_matrix_element.py
giacomomagni: Indeed I forgot to remove this

tests/test_matching_nnlo.py
```diff
@@ -77,3 +77,24 @@ def test_pegasus_sign():
...
+
+    ref_val_l0 = [9.96091, 10.6616,9.27572, 8.25694, 7.49076]
+    ref_val_l10 = [289.097, 278.051, 223.261, 186.256, 160.027]
+    for n in range(1,6):
+        N = 2 * n
+        sx = get_sx(N)
+
+        L = 0.0
+        aS2 = A_singlet_2(N, sx, L)
+        np.testing.assert_allclose(aS2[2,0], ref_val_l0[n-1], rtol=6e-4)
```
felixhekhorn, line 96: Mmm is `6e-4` really sufficient? ...

giacomomagni: Indeed this seems odd, since the other ones  I am trying with are matching to ``5e-6`` accuracy....
Unfortunately this Bluemlein expressions can not be used in python directly because they depends on this HarmonicSums package.
It's not fully clear to me why the computed only even moments also at NNLO, I need to check the reference...

felixhekhorn: - Maybe you can bisect the expressions and check bit by bit? and hence verify your conjecture about `M[Li_3(1-x)](N)` of beeing potentially bugged?
- I think the reason  why they compute only even moments is related to the way the OPE works ... I faced the same problem for the NLO ad and one needs to be very careful about the correct analytically continuation into complex N

felixhekhorn:
@giacomomagni do you have meanwhile also the `exact` available? do you think this might improve the issue?

tests/test_matching_nnlo.py
felixhekhorn: Looking at Silvias expresions:
- the most complicated expression for `AQg123N` is `S[-2, 1, N]` - which we know via `g_3`: https://eko.readthedocs.io/en/develop/theory/Mellin.html#harmonic-sums
- so this should work at least on the same level as all the others (and as the `ad.NLO`)
- did you triple check NLO (since without this everything beyond is invisible)?
- if necessary you can translate this expressions using https://github.com/N3PDF/eko/wiki/Parse-NLO-expressions (or otherwise I can do it for you ...)

giacomomagni: Hi Felix, Okay I'll work on this. yes probably I can translate ``AQg123N`` it in python completely, but still it won't work because it has only the even moments, right? 
 no I haven't checked the NLO yet.

felixhekhorn: I still believe in my comment I stated in the docs:
> where formally η=(-1)^N but in all singlet-like quantities it has to be analytically continued with 1 and with -1 elsewise.

src/eko/matching_conditions/nnlo.py
```diff
@@ -159,25 +167,82 @@ def A_hg_2(n, sx, L):
...
+        / (3 * (1 + n) ** 2 * (2 + n))
+        + (1 + 2 / n + n)
+        / ((1 + n) * (2 + n))
+        * (
+            24 * Sm2 * S1
+            + 10 * S1 ** 3 / 9
+            + 46 * S1 * S2 / 3
+            + 176 * S3 / 9
+            - 24 * Sm21
+        )
```
AleCandido, line 245: Is it much slower to evaluate than the former one?

giacomomagni: A bit slower, The matching integration now takes around 50 second with the usual grid.

AleCandido: And before how much was it taking?

giacomomagni: 40 seconds, probably...

AleCandido: Ok, 25% we can probably afford (it is 25% on a part of the calculation, even not on the full one)

src/eko/strong_coupling.py
```diff
@@ -306,6 +318,7 @@ def a_s(self, scale_to, fact_scale=None, nf_to=None):
         is_downward_path = False
         if len(path) > 1:
             is_downward_path = path[1].nf < path[0].nf
+        shift = 3 if not is_downward_path else 4
```
giacomomagni, line 321: We learnt the lesson!

AleCandido: Indeed, but I'm not completely sure about it. And I don't know actually how to test it properly at the PDF level (APFEL is not doing backward, so we should evolve forward and see if eko is able to go back...), probably it deserves a dedicated section in `benchmark_strong_coupling`...

giacomomagni:
Benkmark with `apfel` it's working also for me with  Kt != 0

src/eko/strong_coupling.py
giacomomagni: It looks consistent with the rest of the backward implementation, whenever you match you use the `nf` in the lower patch.

felixhekhorn:
> Benchmark with apfel it's working also for me with Kt != 0

good! please make a dedicated section in the `apfel_bench.py` so we can reproduce it any time we want - keep in mind that at some point we want to return to run the benchmark more frequently (i.e. automatically)

PR #79, opened 2021-11-16
Init Ekobox

felixhekhorn:
- [x] strip `apply_pdf` from `Output`
- [x] dump LHAPDF files from `eko`
- [x] import `eko_identity` from [`pineko`](https://github.com/N3PDF/pineko/blob/f9c578735f0587c170f75a250832ff280edb7771/old.py#L50-L54)
- [x] import `eko`s product from @giacomomagni scripts


AleCandido:
Thank you, this we'll make us closer to usability, and then it'll be good for production (even if maybe not strictly required for publication).

benchmarks/ekomark/plots.py
```diff
@@ -193,6 +193,8 @@ def plot_operator(var_name, op, op_err, log_operator=True, abs_operator=True):
 
     ax = plt.subplot(1, 3, 3)
     plt.title("|error/value|")
+    # TODO: next line will cause a "invalid value encountered in true divide
+    np.seterr(divide="ignore", invalid="ignore")
```
felixhekhorn, line 197: you can do that, but you should revert afterwards: https://numpy.org/doc/stable/reference/generated/numpy.seterr.html?highlight=seterr#numpy.seterr

benchmarks/ekomark/plots.py
```diff
@@ -257,6 +261,7 @@ def save_operators_to_pdf(path, theory, ops, me, skip_pdfs):
 
 
                 for label_in in ops_names:
+                    import pdb; pdb.set_trace()
```
AleCandido: Commited `pdb`, please remove

src/ekobox/gen_theory.py
```diff
@@ -0,0 +1,165 @@
+import io
+import math
+import pathlib
+
+import yaml
+from banana.data import sql
+
+
+def gen_theory_card(fns, pto, initial_scale, update=None, dump=False, name=None):
```
AleCandido, line 9: This should exist, at least similar, in `banana`. So if we need to expose with this signature is fine to define it, but it should at least rely on the existing one (and existing default theory)

src/ekobox/gen_theory.py
```diff
@@ -0,0 +1,165 @@
...
+        for k in update.keys():
+            if k not in def_theory.keys():
+                raise ValueError("Provided key not in theory card")
+        def_theory.update(update)
+    if dump:
+        dump_theory_card(name, def_theory)
+    return def_theory
+
+
+def dump_theory_card(name, theory):
```
AleCandido, line 105: Same comment of `op_card` -> `import`/`export`

src/ekobox/gen_op.py
```diff
@@ -0,0 +1,74 @@
+import io
+
+import yaml
+from banana.data import sql
+from ekomark.data import operators
+
+
+def gen_op_card(Q2grid, update=None):
```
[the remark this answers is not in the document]
andreab1997: In operators.py (in ekomark) there is the default op.card I used. Anyway I thought that, to make the user life easier, it would be good to have a function to build an op. card from default (with just the Q2grid  mandatory input) and to modify it in a simple way. So what do you think?

andreab1997:
I was wondering if also ekobox needs a CLI. What do you think? @AleCandido @felixhekhorn @giacomomagni

src/ekobox/gen_theory.py
andreab1997: For the same reason of op.cards I implemented this function. Now it also rely on the default card you mentioned.

AleCandido:
> I was wondering if also ekobox needs a CLI. What do you think? @AleCandido @felixhekhorn @giacomomagni

We already thought about an `eko` CLI, and maybe this will actually be called `eko` and implemented inside `ekobox`, but consider the next three items:

1. The purpose of `ekobox` is to support mainly notebook programming: for users plots are more meaningful, and whatever data is better to see in pandas dataframes, so notebook are the perfect match (we are usually using navigator and scroll through numbers, but this is mostly interesting for an eko developer I guess)
2. The CLI is fine, but should support only 2 operations:
   - given theory and operator cards, generate the corresponding operator (in the main output format, i.e. tar archive)
   - given theory and operator cards, and an initial PDF, generate and dump the corresponding LHAPDF grid
3. No NNPDF project will directly use CLI, at least for the time being, since FkTable generation will use evolution through `pineko` (thus making use of python API), and most likely even fit evolution will be managed by `vp` (that is another python code and will make use of python API) -> so it's not urgent

andreab1997:
> > I was wondering if also ekobox needs a CLI. What do you think? @AleCandido @felixhekhorn @giacomomagni
> 
> We already thought about an `eko` CLI, and maybe this will actually be called `eko` and implemented inside `ekobox`, but consider the next three items:
> 
> 1. The purpose of `ekobox` is to support mainly notebook programming: for users plots are more meaningful, and whatever data is better to see in pandas dataframes, so notebook are the perfect match (we are usually using navigator and scroll through numbers, but this is mostly interesting for an eko developer I guess)
> 2. The CLI is fine, but should support only 2 operations:
>    
>    * given theory and operator cards, generate the corresponding operator (in the main output format, i.e. tar archive)
>    * given theory and operator cards, and an initial PDF, generate and dump the corresponding LHAPDF grid
> 3. No NNPDF project will directly use CLI, at least for the time being, since FkTable generation will use evolution through `pineko` (thus making use of python API), and most likely even fit evolution will be managed by `vp` (that is another python code and will make use of python API) -> so it's not urgent

OK so eventually the CLI will be for ekobox and eko together -> it should wait

AleCandido:
> OK so eventually the CLI will be for ekobox and eko together -> it should wait

If you feel like you can even start, simply at the moment is not a priority (but for as long as you are able to, please keep choosing things you prefer to do :) ).

AleCandido:
@andreab1997 : did you commit the `cached_out` by chance or intentionally?

andreab1997:
> @andreab1997 : did you commit the `cached_out` by chance or intentionally?

Intentionally. I provided the possibility to evolve using a cached output and I was testing it (this helps also to make the test of CT14 faster).

AleCandido:
Keep an eye at pylint, you can see the report diff directly on [codefactor](https://www.codefactor.io/repository/github/n3pdf/eko/pull/79)

AleCandido:
I imported `eko_identity`, such that I can remove it from `pineko` (where it is not used any longer).

But this PR is getting old, and so annoying to update with `develop`. Let's get this last point within next week and merge what we have (or even this week if @andreab1997 is available, but no pressure and priority to MHOU).

src/ekomark/benchmark/runner.py
```diff
@@ -14,6 +14,7 @@
 
 import eko
 from eko import basis_rotation as br
+from ekobox import apply
```
AleCandido, line 17: Here we have a problem: we're using `ekobox` in `ekomark`, but the dependency is supposed to go the other way round.

Consider that the database and `navigator` are defined inside `ekomark`, but we want to expose them in the UI.
On the other hand, what it's done here is intuitive, so we really need a more involved solution...

(I wonder if it's worth to split the PDF dumping in a further separate package...)

AleCandido:
Moreover, for a yet unknown reason, I've broken the tests (or they were broken even before...).

```
FAILED tests/test_ekobox/test_evol_pdf.py::test_evolve_single_member - RuntimeError: Metadata for key: Interpolator not found.
FAILED tests/test_ekobox/test_evol_pdf.py::test_evolve_more_members - RuntimeError: Metadata for key: Interpolator not found.
```

tests/test_ekobox/fakepdf/myCT14llo_NF3/myCT14llo_NF3.info
```diff
@@ -0,0 +1,29 @@
+SetDesc: 'fake CT14llo, 1-loop PDF evolution with HOPPET and HOPPET alphas_s running solution. This set has 1 member PDFs. mem=0 --> max NF=3'
```
giacomomagni, line 1: Just for curiosity, there is no way to reuse the ones that we have already in the `banana` test ?

AleCandido: It's going to be complicated: every time you want to test you need to download even the `banana` repository, since they are not packaged in `banana`, but only contained in the repo.

In general, I'd really like to avoid duplication, but in this specific case, I'd like tests to be fully self contained, in both `eko` and `banana`, so if we need these PDF files, better to have them in both places.

Actually, given the duplication and the issue with `lhapdf`, I'd like to avoid having PDF files and `lhapdf` at all, if possible. I know that `ekobox` is supposed to produce files that are going to be read by `lhapdf`, but I'm suggesting a replacement:
- let's make unit tests actually *unit*, and move everything that requires `lhapdf` in benchmarks instead (they will run in a container anyhow)

AleCandido: If they are in benchmarks, we can even think about packaging the PDF files within the benchmarks container, and reuse that one for both `banana` and `ekomark`, solving the duplication.

felixhekhorn: Avoiding `lhapdf` here (and in ``banana`) is not possible - we actually want to test we can produce LHAPDF files and the only way to find out we can, is by using `lhapdf` - and indeed that is non-trivial, as their parser is picky

AleCandido: I'm simply proposing to change section, and call it a benchmark. I'm telling you that running it as a test is not possible, since the python package is not reliable, and it's going to make the whole test suite failing forever, making all the tests useless.

src/ekomark/benchmark/runner.py
felixhekhorn: or we just move the apply stuff inside `ekomark`? if this is the first place we need it ... (and it is a benchmark - with `lhapdf`)

src/ekobox/member.py
```diff
@@ -0,0 +1,266 @@
+# -*- coding: utf-8 -*-
```
felixhekhorn, line 1: Wait: this looks a lot like https://github.com/N3PDF/eko/blob/29f0cae94d3f67e60d29321e01bc4c0f22c675f8/src/eko/member.py - why do we need a copy here? - Can someone remind me please what we had in mind with `import ekos product from @giacomomagni scripts`

AleCandido: Consistent `eko`s can be composed, simply by matrix multiplication, and I guess @giacomomagni did it for the backward study. We wanted to keep the work done, and `ekobox` should be a suitable place.

giacomomagni: So the product between ekos is done for example [here](https://github.com/N3PDF/eko/blob/497773932f048a8e0c5873be87df082dd4dd0b25/backward_paper/runner.py#L92). 
There should be something similar also in the eko paper repo. 
But why do you need to repeat `meber.py`?

AleCandido: This I don't know, no idea.

src/ekomark/benchmark/runner.py
andreab1997: Do we want to move apply in ekomark then?

AleCandido: Even fine, let's do this

andreab1997: Ok, should be fine now.

src/ekomark/benchmark/runner.py
```diff
@@ -14,7 +14,7 @@
 
 import eko
 from eko import basis_rotation as br
-from ekobox import apply
+from ekomark import apply
```
felixhekhorn, line 17: Inside `ekomark` you should use `from . import apply`

tests/test_output.py
```diff
@@ -11,7 +11,7 @@
 
 from eko import basis_rotation as br
 from eko import output
-from ekobox import apply
+from ekomark import apply
```
felixhekhorn, line 14: I know this is a bit tricky, but maybe can you find a way to split the apply tests (`test_apply` + `test_apply_flavor`) into a seperate file: `tests/ekomark/test_apply.py` ?

felixhekhorn:
@AleCandido we're failing the tests with

```
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
lhapdf.pyx:706: in lhapdf.mkPDF
    ???
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

>   ???
E   RuntimeError: Metadata for key: Interpolator not found.

lhapdf.pyx:654: RuntimeError
----------------------------- Captured stdout call -----------------------------
LHAPDF 6.4.0 loading /home/runner/work/eko/eko/tests/test_ekobox/fakepdf/myCT14llo_NF3/myCT14llo_NF3_0000.dat
myCT14llo_NF3 PDF set, member #0, version 1
```
which most likely means we're not loading `lhapdf.conf` (for me located next to `pdfsets.index`) - any idea on how to fix?

tests/test_output.py
andreab1997: Should be ok now

tests/test_apply.py
```diff
@@ -0,0 +1,105 @@
...
+import pytest
+
+from eko import basis_rotation as br
+from eko import output
+from ekomark import apply
+
+from .test_output import FakePDF
+
+
+class TestApply:
```
felixhekhorn, line 19: Ok - you can copy ;-) could you please find out if there is a way to inherit also in tests (which I assume there is)

andreab1997:
I noticed an old folder in tests (test_ekobox) which is basically empty since I believed it has been replaced by the folder ekobox in tests. It is ok to delete it? @felixhekhorn @AleCandido @giacomomagni

felixhekhorn:
> I noticed an old folder in tests (test_ekobox) which is basically empty since I believed it has been replaced by the folder ekobox in tests. It is ok to delete it? @felixhekhorn @AleCandido @giacomomagni

@andreab1997 I think this is just a local problem for you and the folder may contain some temporary files - in the repo there is no such thing: https://github.com/N3PDF/eko/tree/feature/init-ekobox/tests

AleCandido:
For sure there will be the `__pycache__` folder inside: since it's gitignored, it can't be deleted by `git`.

In general, `git` does not manage your folders, but your files. If you have an empty folder, there is no way to add it to the repo (that's why sometimes there are `.gitkeep` files commited, to retain otherwise empty folders).

andreab1997:
> > I noticed an old folder in tests (test_ekobox) which is basically empty since I believed it has been replaced by the folder ekobox in tests. It is ok to delete it? @felixhekhorn @AleCandido @giacomomagni
> 
> @andreab1997 I think this is just a local problem for you and the folder may contain some temporary files - in the repo there is no such thing: https://github.com/N3PDF/eko/tree/feature/init-ekobox/tests

Ok then is not a problem at all

andreab1997:
Regarding commit 039806a, I have basically fixed the only test that was not working. However I noticed a possible issue. When I first wrote ekobox, the ekos were dumped with a name resulting from the combination of operator_card hash and theory_card hash. Therefore, depending on that,  evol_pdf would look for the right eko. Now this is no more the case I believe @felixhekhorn @AleCandido

pyproject.toml
```diff
@@ -52,11 +52,13 @@ pdbpp = "^0.10.3"
 pytest = "^6.2.5"
 pytest-cov = "^3.0.0"
 pytest-env = "^0.6.2"
+a3b2bbc3ced97675ac3a71df45f55ba = "^6.4.0"
```
giacomomagni, line 55: What is this?

AleCandido: `lhapdf` in disguise

AleCandido: https://pypi.org/project/a3b2bbc3ced97675ac3a71df45f55ba/

AleCandido: Unfortunately, it is not working on Mac M1 (not yet), and Andy Buckley did not accept yet my [Merge Request](https://gitlab.com/hepcedar/lhapdf/-/merge_requests/16), so it's still unofficial...

src/ekobox/evol_pdf.py
```diff
@@ -0,0 +1,177 @@
...
+        ope2_dict[Q2] = eko_fin["Q2grid"][Q2]["operators"]
+        ope2_error_dict[Q2] = eko_fin["Q2grid"][Q2]["operator_errors"]
+    final_op_dict = {}
+    final_op_error_dict = {}
+    final_alphas_dict = {}
+    final_dict = {}
+    for Q2 in ope2_dict.keys():
+        final_op_dict[Q2] = np.einsum("ajbk,bkcl -> ajcl", ope1, ope2_dict[Q2])
+        final_op_error_dict[Q2] = np.einsum(
+            "ajbk,bkcl -> ajcl", ope1_error, ope2_error_dict[Q2]
```
giacomomagni, line 163: Is this correct? Or was just to have and approximation?

src/ekobox/gen_theory.py
```diff
@@ -0,0 +1,83 @@
...
+    """
+    # Constructing the dictionary with some default values
+    theory = copy.deepcopy(theories.default_card)
+    # delete unuseful member
+    del theory["FNS"]
+    # Adding the mandatory inputs
+    theory["PTO"] = pto
+    theory["Q0"] = initial_scale
+    # Update user choice
+    if isinstance(update, dict):
```
giacomomagni, line 38: `if update is not None:`

src/ekobox/evol_pdf.py
AleCandido: I'd say it's wrong, but this integration errors are almost useless, and they are becoming a plague.
I would make them optional (is a factor 2 in memory and disk storage), and do not propagate them in non-trivial places anyhow...

pyproject.toml
giacomomagni: it looks he forgot about it..

src/ekobox/gen_info.py
```diff
@@ -0,0 +1,50 @@
...
+    template_info["NumFlavors"] = 14
+    template_info["Flavors"] = [-6, -5, -4, -3, -2, -1, 1, 2, 3, 4, 5, 6, 21, 22]
+    template_info["XMin"] = operators_card["interpolation_xgrid"][0]
+    template_info["XMax"] = operators_card["interpolation_xgrid"][-1]
+    template_info["NumMembers"] = num_members
+    template_info["OrderQCD"] = theory_card["PTO"]
+    template_info["QMin"] = round(math.sqrt(operators_card["Q2grid"][0]), 4)
+    template_info["QMax"] = round(math.sqrt(operators_card["Q2grid"][-1]), 4)
+    template_info["MZ"] = theory_card["MZ"]
+    template_info["MUp"] = 0.0
```
giacomomagni, line 41: Some of these info are already in the template. Is this to prevent a possible change?

src/ekobox/evol_pdf.py
andreab1997: In fact it is only temporary since I did not know what we wanted to do with them. So, should I simply remove them and keep the original errors from the initial ekobox operator?

src/ekobox/gen_info.py
andreab1997: Yes exactly

src/ekobox/genpdf/flavors.py
```diff
@@ -0,0 +1,136 @@
...
+        labels : list()
+            list of labels
+
+    Returns
+    -------
+        bool :
+            is evolution basis
+    """
+    for label in labels:
+        if not isinstance(label, str):
```
giacomomagni, line 106: why do you need this check?

src/ekobox/gen_info.py
giacomomagni: Maybe you can raise a warning and ignore the change

src/ekobox/evol_pdf.py
giacomomagni: I'd set them to zeros, better than something random

AleCandido: Nope, better no errors than wrong errors. You have two operators, the errors on the product are not any of: the errors of the first, the errors on the second, the product of the errors.
If you wish they should be the sum of the products of the error on one with the value of the other:

```
(a + Da)(b + Db) = ab + bDa + aDb + DaDb ~ ab + bDa + aDb
```

AleCandido: However, it's a bit weird that the product of two EKOs (that does not involve any pdf) is inside the `evol_pdf` module.

Put in a separate one. In case of doubt (if there is not any other suitable module, nor it's worth a module on its own) put in a `utils.py` module (that is the generic name we use for extra stuffs).

src/ekobox/gen_info.py
felixhekhorn: but surely just overwriting (1 line) is easier then checking and warning (2 lines) - and we want a consistent output ...

src/ekobox/evol_pdf.py
giacomomagni: They should be added in quadrature no? 
```
np.sqrt((Da*b)**2 + (Db*a)**2) 
```

AleCandido: It depends if you consider them statistical or not, and actually they are not (unless you have a statistical interpretation of quadrature integration, that is not Monte Carlo).

src/ekobox/gen_info.py
giacomomagni: okay fine

src/ekobox/evol_pdf.py
andreab1997: Should be fixed now

src/ekobox/utils.py
```diff
@@ -35,9 +35,10 @@ def ekos_product(eko_ini, eko_fin, in_place=True, force=False):
     final_dict = {}
     for Q2 in ope2_dict.keys():
         final_op_dict[Q2] = np.einsum("ajbk,bkcl -> ajcl", ope1, ope2_dict[Q2])
-        final_op_error_dict[Q2] = np.einsum(
-            "ajbk,bkcl -> ajcl", ope1_error, ope2_error_dict[Q2]
-        )
+        final_op_error_dict[Q2] = np.add(np.einsum(
```
AleCandido: If you wish, `np.add` is equivalent to `+` for `np.ndarray`s

src/ekobox/genpdf/flavors.py
andreab1997: I honestly not remember. Probably we can just remove it.

pyproject.toml
```diff
@@ -97,6 +99,7 @@ python_classes = ['Test*', 'Benchmark*']
 python_functions = ['test_*', 'benchmark_*']
 addopts = [
   '--cov=eko',
+  '--cov=ekobox',
```
felixhekhorn, line 102: We should also add `ekomark`

src/ekobox/genpdf/flavors.py
felixhekhorn: I think to distinguish from pid = int

src/ekobox/utils.py
andreab1997: Yes I know (it is in the docs) but I had some problems with that. That's why I prefer doing like this. Anyway I can change if you prefer.

AleCandido:
Current status is that:

- I removed all "benchmarks" from unit tests, this makes the CI working again
- however, we lowered the coverage: there is an 87% only for `eko/output.py`, corresponding to `apply`
  - [ ] is apply still needed in `eko`? is it not superseded by the one in `ekomark`, re-exported by `ekobox`?
  - [ ] make sure `ekobox` is actually re-exporting `ekomark.apply`
  - [ ] remove code from `eko/output.py` if not needed any longer
- at this point, a campaign for `ekomark` and `ekobox` **unit** tests (without `lhapdf` or other external ingredients) coverage might become crucial, but I would postpone to another PR (and we don't need to target 100%, we have also *unit benchmarks*)
- [ ] I want to decorate *unit benchmarks* (that are not unit at all, but they belonged to *unit tests* up to one moment ago), to split them from the runner based ones
- [ ] I have to write a workflow for them alone
  - it will be extremely simple
  - it will run in a dedicated container with `lhapdf` pre-installed https://github.com/N3PDF/external/pkgs/container/bench-evol
  - it will provide coverage (this coverage will be summed up with the unit tests one in codecov, but with a different label, and it should be considered somewhat independent of the unit tests one)

@andreab1997 can you take care about `eko/output.py` items now?
Moreover, remember to:
- install `pre-commit` in the repo (you need not only to install the CLI, but also to run `pre-commit install` inside the repo)
- have a look every now and then to codefactor (or just run `pylint` yourself): there were quite a few unused imports

pyproject.toml
AleCandido: Scheduled for later #102

src/ekobox/utils.py
AleCandido: I did it, since I guessed the problems were only syntactical. In any case, can you please test it that is still fine for you?

AleCandido:
> * have a look every now and then to codefactor (or just run `pylint` yourself): there were quite a few unused imports

It was a general advice, since the former status: now I moved the tests and `pylint` went crazy...
No need to solve all the *No name 'x' in module 'eko'*, ignore them.

src/ekomark/apply.py
```diff
@@ -0,0 +1,100 @@
...
+    Returns
+    -------
+        out_grid : dict
+            output PDFs and their associated errors for the computed Q2grid
+    """
+    if rotate_to_evolution_basis:
+        return apply_pdf_flavor(
+            output, lhapdf_like, targetgrid, br.rotate_flavor_to_evolution
+        )
+    return apply_pdf_flavor(output, lhapdf_like, targetgrid)
```
AleCandido, line 33: Purge corresponding `eko.output.Output.apply_pdf`

src/ekomark/apply.py
```diff
@@ -0,0 +1,100 @@
...
+    # rotate/interpolate to target grid
+    if targetgrid is not None:
+        b = interpolation.InterpolatorDispatcher.from_dict(output, False)
+        rot = b.get_interpolation(targetgrid)
+        for evpdf in out_grid.values():
+            for pdf_label in evpdf["pdfs"]:
+                evpdf["pdfs"][pdf_label] = np.matmul(rot, evpdf["pdfs"][pdf_label])
+                evpdf["errors"][pdf_label] = np.matmul(rot, evpdf["errors"][pdf_label])
+
+    return out_grid
```
AleCandido, line 100: Same of above

AleCandido (changes requested):

AleCandido:
Update:
- the *Isolated benchmarks* workflow is finally running!
- unfortunately, they are still broken, especially all and only those that involve `lhapdf`: now the problem is not any longer `lhapdf` availability (it is available), but we have truly 2 kinds of errors (that of course were happening also locally):
  1. `RuntimeError: Metadata for key: Interpolator not found.`
  2. `RuntimeError: PDF grid data error on line 7: 2 flavor entries seen but 14 expected`

A further couple of weird points:
- the first kind of error (1.) is not happening on my local environment, so I still believe is an `lhapdf` installation problem (i.e. it is related to running in a virtual environment, with `lhapdf` out of the environment, so most likely it's using the one provided by the package), I guess this has to be solved from the workflow side :disappointed: 
- the isolated benchmark: `benchmarks/eko/benchmark_msbar_evolution.py` it is broken on my local environment, but it is **not** in the workflow
  I guess it's still due to the fact that `apfel` is not available because of the environment

```
E                   AssertionError:
E                   Not equal to tolerance rtol=1e-07, atol=0
E                   order=0
E                   Mismatched elements: 1 / 9 (11.1%)
E                   Max absolute difference: 184.00223939
E                   Max relative difference: 0.36568735
E                    x: array([[  1.604613,   5.824621, 319.165946],
E                          [  0.918422,   3.3338  , 182.6789  ],
E                          [  0.889274,   3.227995, 176.881194]])
E                    y: array([[  1.604613,   5.824621, 503.168186],
E                          [  0.918422,   3.3338  , 182.6789  ],
E                          [  0.889274,   3.227995, 176.881194]])
```

AleCandido:
I gave up with fancy privileges restrictions inside the container, since it looks like GitHub workflows, when the job is fully inside a container, require to run it as root:
https://docs.github.com/en/actions/creating-actions/dockerfile-support-for-github-actions#user

Since [19ee4c5](https://github.com/N3PDF/eko/pull/79/commits/19ee4c58b12919ab4d743ef014f93489c13655ef) the new *Isolated benchmarks* are actually fully working, and they reproduce local results (the trick has been, unsurprisingly, `poetry config virtualenvs.create false`).

Now, the situation is the following:
- the error of first kind (i., see previous comment) is gone, as expected it was only related to a missing full installation of LHAPDF (since it was used through the package), and thus (most likely) missing `lhapdf.conf` (to solve locally, install LHAPDF by hand in poetry environment `poetry env info -p`, e.g. from https://github.com/N3PDF/external)
- [ ] the error of the second kind (ii.) is still there, to be solved
- [ ] the error on `msbar_masses` is back, since now even APFEL is available; maybe we just need to upgrade APFEL version, what do you think @giacomomagni and @felixhekhorn?
- [ ] of course, I forgot to mention an error already known to @andreab1997, i.e. `FileNotFoundError: [Errno 2] No such file or directory: '/__w/eko/eko/benchmarks/ekobox/cached_out/o7f0768_te3dbf6.tar'`, so, most likely, it needs to be regenerated (or we can drop the idea of caching it, and just generate online, since it's no longer a unit test)

Solved this 3 errors, plus completing the unit tests coverage on `eko` (and only `eko`), it's enough to get back the green tick on workflows and codecov, and enough for me to merge.
Final caveats:
- for the coverage see [above](https://github.com/N3PDF/eko/pull/79#issuecomment-1059814575), it might be enough to delete some old code
- for the full green tick we would also need to satisfy Pylint (this is the last thing I'll try to take care myself); I don't consider this needed to merge

P.S.: I start to believe this branch to be cursed, let's close it as soon as possible

andreab1997:
> * [ ]  of course, I forgot to mention an error already known to @andreab1997, i.e. `FileNotFoundError: [Errno 2] No such file or directory: '/__w/eko/eko/benchmarks/ekobox/cached_out/o7f0768_te3dbf6.tar'`, so, most likely, it needs to be regenerated (or we can drop the idea of caching it, and just generate online, since it's no longer a unit test)

Actually, in [039806a](https://github.com/N3PDF/eko/pull/79/commits/039806a1ecb4a758661ca94518c14add89810f76) I should have fixed that test (I tried and for me it is working). Is failing for you? @AleCandido

AleCandido:
Now it's not a matter of local environments: it's failing in the CI, you can have a look at the logs by yourself.

In any case, I would give up on pre-conputed output: everything that is stored has to be kept up to date, and we know that Eko is going to change (including the output format itself).

Online computation will ensure consistency.

andreab1997:
> Now it's not a matter of local environments: it's failing in the CI, you can have a look at the logs by yourself.
> 
> In any case, I would give up on pre-conputed output: everything that is stored has to be kept up to date, and we know that Eko is going to change (including the output format itself).
> 
> Online computation will ensure consistency.

ok, perfect

giacomomagni:
>   I guess it's still due to the fact that `apfel` is not available because of the environment (so failing when `apfel` available); here @giacomomagni might help, my hypothesis is that these numbers have been obtained with a patched version of `apfel`, and not v3.0.5
> 
Yes indeed, it looks the failure is due to top evolution (`Q=1`) which was the bugged one.

AleCandido (approved):

PR #82, opened 2022-01-03
Introduce poetry

AleCandido:
Switch build tool and dependency management to `poetry`, and we other things:

- [x] update testing workflow
- [x] update deploy workflow
- [x] add task runner
- [x] add dynamic version
- [x] update docs building


AleCandido:
Deploy workflow is upgraded, but not yet tested (we need a pre-release).

The update is sufficiently small not to justify a dedicated setup with `test.pypi.org`, if it'll break, I'll fix it in a point release.

AleCandido:
It's relatively ready to merge, but if someone can review this and try to figure out if there was anything else depending on setuptools, or any more dead code, it would be appreciated :)

AleCandido:
I'm updating the docs section, but as per `yadism` there is the usual issue with `rtd`, documented in the following issues:

- https://github.com/python-poetry/poetry/issues/754
- https://github.com/readthedocs/readthedocs.org/issues/4912

The workaround we apply for the time being is that one coming from this comment https://github.com/readthedocs/readthedocs.org/issues/4912#issuecomment-664002569 and above.

Not elegant (as the comment itself and preceding says), but it's currently a limitation of `rtd`, i.e.:
- we would like to have `rtd` take into account also *dev-dependencies* (those dependencies that are not needed by package users)
- since *dev-deps* are not a `pip` feature, and `poetry` is supported by `rtd` only through `pip` and PEP 517, they are completely ignored
- instead `extras` are a pip feature, installed like `pip install package[extra]` (but they are a different feature, because are dependencies needed by the end user, but not for all use cases), and so we avoid making them main dependencies by making them extra, and tell `rtd` to install extras of `docs` group (and duplicating them as *dev-deps*, since they are actually needed in development)

pyproject.toml
```diff
@@ -1,10 +1,84 @@
...
+
+[tool.poetry]
+name = "eko"
+version = "0.0.0"
+description = "Evolution Kernel Operator"
+authors = [
+  "A. Candido <alessandro.candido@mi.infn.it>",
+  "S. Carrazza <stefano.carrazza@cern.ch>",
+  "J. Cruz-Martinez",
+  "F. Hekhorn",
```
felixhekhorn, line 13: Could you please add the other mails? Do we want to drop SC and JCM?

.github/workflows/pypi.yml
```diff
@@ -3,35 +3,49 @@ name: deploy
 on:
   push:
     tags:
-      - '*'
+      - "*"
 
 jobs:
-  build-n-publish:
+  publish:
```
felixhekhorn, line 9: Why is this different from https://github.com/N3PDF/yadism/blob/master/.github/workflows/pypi.yml ? will this be the new style?

pyproject.toml
AleCandido: I've been lazy, and since I didn't want to make decisions on which email to put I left it for you guys :)
Add your own, Juan and Giacomo can do the same.

.github/workflows/pypi.yml
AleCandido: I added a cache layer, to skip dependency installation when not needed. I updated even in `yadism`, and I'm going to do it in `banana`.

Actually, this is annoying, since it is the exact same workflow, so I scheduled even to [factorize workflows](https://docs.github.com/en/actions/learn-github-actions/reusing-workflows).
But since this is possible only inside a single organization, if I want to share across repos I have to pick one to put them all, maybe a new one, and similar complications... I postponed and copy-pasted, for the time being.

pyproject.toml
felixhekhorn: @giacomomagni I added your Nikhef mail (as in yadism) - is that alright?

pyproject.toml
```diff
@@ -1,10 +1,81 @@
...
+[tool.poetry-dynamic-versioning.substitution]
+files = ["src/eko/version.py"]
+
+[tool.poe.tasks]
+# until https://github.com/numba/numba/pull/5660 is confirmed we need to deactivate numba prior running
+test = { shell = "NUMBA_DISABLE_JIT=1 pytest tests" }
+lint = "pylint src/*/*.py -E"
+lint-warnings = "pylint src/*/*.py --exit-zero"
+sandbox = "python benchmarks/runners/sandbox.py"
+# navigator = "navigator"
```
felixhekhorn, line 68: Why did you comment? (apart from the fact that in `eko` the thing is called `ekonavigator` - for name clash reason, if I remember correctly @giacomomagni )

pyproject.toml
```diff
@@ -1,10 +1,81 @@
...
+black = "black ."
+isort = "isort . --skip env"
+format = ["black", "isort"]
+docs = { "shell" = "cd doc; make html" }
```
felixhekhorn, line 72: VS code is comlaining 
```
[{
	"resource": "/home/felix/Physik/N3PDF/EKO/eko/pyproject.toml",
	"owner": "_generated_diagnostic_collection_name_#2",
	"severity": 8,
	"message": "Expected \"}\", [ \\t] or [A-Za-z0-9_\\-] but \"\\\"\" found.",
	"source": "Toml Parser",
	"startLineNumber": 72,
	"startColumn": 11,
	"endLineNumber": 72,
	"endColumn": 12
}]```

are you sure this is working?

File: pyproject.toml (hunk 1 above)
Comment (giacomomagni) [reply pyproject.toml]: perfect, thanks @felixhekhorn :)

File: pyproject.toml (hunk 4 above)
Comment (AleCandido) [reply pyproject.toml:72]: Have you tried to run `poe docs`, or even simply [`poetry check`](https://python-poetry.org/docs/cli/#check), or a `toml` loader?

I autoformat everything, and that should be valid.
For reference, my current `toml` formatter is: [`taplo`](https://taplo.tamasfe.dev/).

File: .github/workflows/pypi.yml (hunk 2 above)
Comment (felixhekhorn) [reply .github/workflows/pypi.yml:9]: on the same grounds: should we also duplicate https://github.com/N3PDF/yadism/blob/593a012dc7f4666c2c993dbdd37a6a2d55db0455/install.py ?

File: pyproject.toml (hunk 4 above)
Comment (felixhekhorn) [reply pyproject.toml:72]: well, seems to work ... maybe this is a bug in https://github.com/bungcip/better-toml - I switched to the taplo implementation as well (as the other one doesn't seem to be maintained any longer ...)

File: pyproject.toml (hunk 3 above)
Comment (AleCandido) [reply pyproject.toml:68]: Before there was not, since I'm introducing `poetry` in this PR.

The base have been copied from `yadism`, and I remember that was named `ekonavigator`, but I didn't test the whole part of `ekomark`, so the comment here.
If you can test, it would be perfect :)

File: pyproject.toml (hunk 4 above)
Comment (AleCandido) [reply pyproject.toml:72]: Yeah, unfortunately `toml` is being adopted, but the adoption is rather recent, and TOML 1.0 is even more recent, thus we're lacking tooling.
(Even though `taplo` is quite good, and rather a complete toolbox, but for some reason it doesn't rank high in searches).

Actually, major uses case of `toml` are `Cargo.toml` and `pyproject.toml` themselves (within a few other majors, but I'd say a bit less relevant). In this [official list](https://github.com/toml-lang/toml/wiki#projects-using-toml) all tools related to python are listed because of `pyproject.toml` (black, flit, poetry, pip, tox, pipenv, isort, ...).

Comment (AleCandido) [reply pyproject.toml:72]: All in all, I'd say that TOML is pretty good, but young. It's the younger and improved version of INI (with the clear advantage of having a standard, excluding which INI was already rather good for configs).

Nevertheless, keep in mind that even TOML has pitfalls, and [not everyone likes it](https://hitchdev.com/strictyaml/why-not/toml/).

File: pyproject.toml (hunk 3 above)
Comment (felixhekhorn) [reply pyproject.toml]: at least navigator is working - so I added it back

**Comment by felixhekhorn** [comment]
- do you understand the codecov complaint https://www.codefactor.io/repository/github/n3pdf/eko/pull/82 ? we do not import eko in there, which I guess would be the equivalent
- except for the  `install.py` business, I guess, this is good for merging


Unfortunately there are still the known issues with `poetry`
- a fresh install (even with all packages cached) still takes around 5 mins
- we're waiting eagerly for `v1.2` to get editable ...

Nevertheless, I guess we want to keep going, since we also gain some significant features ...

**Comment by felixhekhorn** [comment]
Actually, there is another issue and I don't know how serious this actually is:

https://github.com/N3PDF/eko/blob/d374b52c45865300e9c776c769f2517e715e03e0/benchmarks/setup.py#L18

this does pull `eko` from PyPI and does *NOT* use the local version

**Comment by AleCandido** [comment]
> * do you understand the codecov complaint https://www.codefactor.io/repository/github/n3pdf/eko/pull/82 ? we do not import eko in there, which I guess would be the equivalent 

I guess this `pylint` doesn't like the concept of *exposing*, or it's simply a bug in the `pylint` implementation itself (since `from .` means import from myself, but in the package sense, and not as a module, so maybe it's simply confused).

> (PS: and a local `pylint` run does not bring the issue)

This has to be related to the pylint version, I guess...
 
> Unfortunately there are still the known issues with `poetry`

It's even worse, I need tens of minutes to upgrade dependencies, when not more than an hour.
Nevertheless, `poetry` is the most actively maintained dependency manager in the python ecosystem, at the moment, and it offers significant advantages over bare `virtualenv` use, so it's still worth for me, and we hope issues will be solved in the near future.

If you wish, you're free to test `Pipenv` and compare the two ;) (it's the only other relevant alternative)

**Comment by AleCandido** [comment]
> Actually, there is another issue and I don't know how serious this actually is:
> 
> https://github.com/N3PDF/eko/blob/d374b52c45865300e9c776c769f2517e715e03e0/benchmarks/setup.py#L18
> 
> this does pull `eko` from PyPI and does _NOT_ use the local version

Here I simply didn't touch anything, yet...

We're in a complicated situation, since `ekomark` depends on `eko`, but at the same time is a development tool for `eko`.

For `yadmark` I already sketched a layout, that to me seemed the cleanest: the benchmark suite is a devdep of the package, since it's actually a part of it, and it's not shipped to the end user.

Things will change with `ekobox`, because at that point `ekomark` will be shipped.
When the time will come, I would simply move `ekomark` in `src/ekomark`, as another package we're developing, and keep everything else in `benchmark` (like runners). So they are simply living side by side.
Unfortunately, we'll have some issues in packaging as well https://github.com/python-poetry/poetry/issues/936, so we might consider shipping them in a unique bundle...

**Comment by felixhekhorn** [comment]
- Okay, but then let's move `ekomark` also to Poetry (and do the same fix, i.e., skip the dependency) - could you please take care of this @AleCandido ?
- and that we will move `ekomark` at some point, I guess we already agreed upon ...
- at this point, `benchmarks` will indeed contain nothing but actual implementations of `ekomark` (=runners) and their associated data

**Comment by AleCandido** [comment]
> * Okay, but then let's move `ekomark` also to Poetry (and do the same fix, i.e., skip the dependency) - could you please take care of this @AleCandido ?

Okay, I'll copy `yadmark` (but it will have same issues).

>  * and that we will move `ekomark` at some point, I guess we already agreed upon ...

Yes, but as said before `poetry` has no support for monorepo, so we might be forced to move to "monopackage"...

File: .github/workflows/pypi.yml (hunk 2 above)
Comment (felixhekhorn) [reply .github/workflows/pypi.yml:9]: the answer is no - should go to README

**Comment by felixhekhorn** [comment]
let's
- [ ] adjust the README
- merge
- immediately after setup monorepo structure

**Comment by AleCandido** [comment]
Wait, it's being tricky to release `banana`, gimme a while to fix it...

**Comment by AleCandido** [comment]
Please @felixhekhorn and @giacomomagni review this, and I'll merge.

File: doc/source/conf.py [hunk 5]
```diff
@@ -17,29 +17,25 @@
...
 
 # The short X.Y version
-version = eko.version.short_version
-if not eko.version.is_released:
-    version = "develop"
+version = eko.__version__
 
 # The full version, including alpha/beta/rc tags
-release = eko.version.full_version
+release = eko.__version__
```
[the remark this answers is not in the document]
giacomomagni: It should be `eko.version.__version__`

.github/CONTRIBUTING.md
```diff
@@ -2,24 +2,54 @@
...
+    Conduct](https://github.com/N3PDF/eko/blob/master/.github/CODE_OF_CONDUCT.md)
+    and use the provided [Issue
+    Templates](https://github.com/N3PDF/eko/issues/new/choose)
+
+## Tools
+
+-   [`poetry`](https://github.com/python-poetry/poetry) is the
+    dependency manager and packaging back-end of choice for this
+    project, refers to official [installation
+    guide](https://python-poetry.org/docs/#installation)
```
giacomomagni, line 20: I would link their git repo https://github.com/python-poetry/poetry, since some instructions on their docs are not up to date.

doc/source/conf.py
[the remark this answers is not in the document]
AleCandido: Yes, it changed in the meanwhile, you're right

.github/CONTRIBUTING.md
AleCandido: The repo is actually linked (line 17). But referring to repo for installation instructions is risky:
- repo by default is `master`, and so even instructions might change (they are unstable) and refer to the latest release (in the best case)
- website by default is `stable`, so it's the advised version to be used

In general, you expect to find official instructions on official website, and developer instructions on repo (I'd say).

giacomomagni (approved):

PR #74, opened 2021-07-01
Backward Evolution Study

giacomomagni:
This PR is for the backward evolution study. We seek for the charm pdf below its mass threshold.

PR #83, opened 2022-01-17
N3LO matching

giacomomagni:
Integrate the eko relevant features coming from #74 inside develop


src/eko/matching_conditions/n3lo/__init__.py
```diff
@@ -0,0 +1,126 @@
+# -*- coding: utf-8 -*-
+r"""
+This module defines the matching conditions for the N3LO |VFNS| evolution.
+
+The expressions are based on:
```
felixhekhorn, line 5: Repeat this on all N3LO matching files

giacomomagni: I added it on purpose only in the `__init__`, I thought the spinx API would have been ugly

felixhekhorn: Well, this is what we have to do ... the only other option would be not to repeat it in the docstring, but just as a comment

AleCandido: I guess in `__init__` should be fine, if really really it's not enough, then simply add a comment to look to `__init__` for the full set of references.
I don't really want to blow up our code with copy-pasting of docstrings, if he will complain it's not my concern (everyone will consider it unreasonable, apart from him no one else care about citations in the code itself).

felixhekhorn: No - we have to do it ... and indeed it will be *my* problem ;-) the files are big in any case, so let's just do it - as said also comments

AleCandido: Nope, put a pointer, not the full copy-paste, it's unreasonable and we don't have to do anything anyone is going to ask us to do.

Even in papers you put references, and not the full bibliography every time you cite. Here we cite the block, and that's it.

felixhekhorn: But this we had already - and it wasn't sufficient; the request was explicitly to put it everywhere

AleCandido: It was in the code documentation, and nowhere in the code (so e.g. it was not distributed in the python package, since it only contains the code, and not the sphinx source). Now it's also in the code.

For everything else a pointer is sufficient, copy-paste does not add any value, people are able to open the file aside, and if they want to ignore, they would ignore anyhow.

AleCandido: And in any case, don't be proactive in complaining in his place. If he'll complain, I'll reply as I'm replying to you, but don't fight for someone else, unless you genuinely agree with him.

felixhekhorn: I do not agree, but I'm scared

giacomomagni: Okay, I'll add in each `OME`  docstring this sentence:
```
When using the code, please cite the complete list of reference, is available in :mod:`eko.matching_conditions.n3lo`
```
This sounds to me unambiguous and neat.
If you don't agree, I would go for repeat it as comment.

AleCandido:
@giacomomagni if you believe [2539cb4](https://github.com/N3PDF/eko/pull/83/commits/2539cb4435d05e6e6d7eab2cfdb8ad408a3ff5ec) to address completely the issue, you should use as commit message `Closes issue #103`, this way GitHub would link the issue to the PR.

However, this is just sugar: if you believe to have solved that issue in this PR, go in the column on the right of this page, and manually link the issue yourself :)

benchmarks/eko/benchmark_evol_to_unity.py
```diff
@@ -104,7 +104,7 @@ def test_operator_grid(
             o_back.op_members[(21, 21)].value @ o.op_members[(21, 21)].value
             + o_back.op_members[(21, 100)].value @ o.op_members[(100, 21)].value,
             np.eye(dim[0]),
-            atol=9e-2,
+            atol=11.2e-2,
```
AleCandido, line 107: Is this not too bad? You're failing of `0.112` on something that has to be `[0, 1]`...

src/eko/evolution_operator/__init__.py
```diff
@@ -245,7 +246,12 @@ class Operator:
             cut to the upper limit in the mellin inversion
     """
 
-    def __init__(self, config, managers, nf, q2_from, q2_to, mellin_cut=5e-2):
+    operator_type = "Evolution"
+    n_pools = int(os.cpu_count() / 2)
```
AleCandido, line 250: I'm still convinced you can avoid to manually define, but it's even like this.

src/eko/evolution_operator/__init__.py
```diff
@@ -424,15 +449,28 @@ def compute(self):
                 self.fact_to_ren,
                 "exponentiated" if self.sv_mode == 1 else "expanded",
```
AleCandido, line 450: Here we should do a better job in general, and start defining `enum`s for this kind of objects.

AleCandido (commented):
Not a very accurate review: most of the stuff consists of N3LO anomalous dimensions, that I have no way to quick check, so I gave up completely.

I provide just a few comments, and delegate to @felixhekhorn a more proper job :)

src/eko/matching_conditions/operator_matrix_element.py
```diff
@@ -300,6 +300,23 @@ class OperatorMatrixElement(Operator):
...
+    # complete list of possible matching operators labels
+    full_labels = [
+        *br.singlet_labels,
+        (br.matching_hplus_pid, 21),
+        (br.matching_hplus_pid, 100),
+        (21, br.matching_hplus_pid),
+        (100, br.matching_hplus_pid),
+        (br.matching_hplus_pid, br.matching_hplus_pid),
+    ]
+    full_labels.extend(
```
felixhekhorn, line 312: you can just continue the previous list

src/eko/evolution_operator/__init__.py
```diff
@@ -245,7 +246,12 @@ class Operator:
             cut to the upper limit in the mellin inversion
     """
 
-    def __init__(self, config, managers, nf, q2_from, q2_to, mellin_cut=5e-2):
+    operator_type = "Evolution"
```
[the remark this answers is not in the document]
felixhekhorn: I see you use it for logging - maybe `log_label` ? in principle this would not be needed since the logger is scoped by `__name__` and hence knows who is logging ... maybe we should change our default log format in `ekomark` and use the more verbose `logging.Formatter("%(asctime)s %(name)s/%(levelname)s: %(message)s")` as [here](https://github.com/N3PDF/pineko/blob/fkutils-move/src/pineko/theory.py#L248) - of course at the price of spam

src/eko/evolution_operator/__init__.py
giacomomagni: okay fine, can remove the division by 2

benchmarks/eko/benchmark_evol_to_unity.py
giacomomagni: I checked that this large discrepancy occurs in the second-last row of the matrix so basically when you have only one off diagonal element which should be 0 but it's not...

src/eko/matching_conditions/operator_matrix_element.py
```diff
@@ -4,24 +4,67 @@
...
 import numpy as np
 from scipy import integrate
 
 from .. import basis_rotation as br
-from .. import interpolation, mellin
 from ..anomalous_dimensions import harmonics
+from ..evolution_operator import Operator, QuadKerBase
 from ..member import OpMember
-from . import nlo, nnlo
+from . import n3lo, nlo, nnlo
```
felixhekhorn, line 18: `import eko` was 19 min for me just now - maybe we should delay this line? Now, I'm quite convinced that we're abusing numba in some way - it should be JIT as promised ... Reading [here](https://numba.pydata.org/numba-doc/dev/reference/jit-compilation.html) and if I remember correctly it seems related to the signature ... Someone remembers why we have signatures everywhere? (I think otherwise cache doesn't work?) and why can't we run in `nopython` mode and would that help us?

src/eko/matching_conditions/operator_matrix_element.py
```diff
@@ -170,65 +224,58 @@ def quad_ker(
...
-        indices = {200: 0, 91: 1}
-    n = mellin.Talbot_path(u, r, o)
-    jac = mellin.Talbot_jac(u, r, o)
+    ker_base = QuadKerBase(u, is_log, logx, mode0)
+    integrand = ker_base.integrand(areas)
+    if integrand == 0.0:
+        return 0.0
 
     # compute the harmonics
     sx = np.zeros(3, np.complex_)
```
felixhekhorn, line 246: the original idea here was to have `sx` with a *fixed* size and to fill the elements when needed - now, already in the old code this was broken as PTO >= 1 already used a reassignment (`sx = np.array(...)`); however, I'd say let's make it consistent now (since with N3LO you're adding more and more elements)

src/eko/matching_conditions/operator_matrix_element.py
```diff
@@ -4,24 +4,67 @@
 |VFNS| evolution.
 """
 
+import functools
 import logging
-import time
 
 import numba as nb
 import numpy as np
 from scipy import integrate
```
felixhekhorn, line 12: This was already a bug in the old code - let's fix it now: `src/eko/matching_conditions/operator_matrix_element.py:12:0: W0611: Unused integrate imported from scipy (unused-import)`

General advice: take a look at `poe lint-warnings` - there are some easy fixes, maybe you can tackle the ones related to your new files?

src/eko/matching_conditions/n3lo/s_functions.py
```diff
@@ -0,0 +1,801 @@
...
+            Hamrmonic sum :math:`S_{2}(N)`
+
+    Returns
+    -------
+        S21 : complex
+            Harmonic sum :math:`S_{2,1}(N)`
+
+    See Also
+    --------
+        eko.mathcing_conditions.n3lo.g_functions.mellin_g18 : :math:`g_18(N)`
```
felixhekhorn, line 180: `mathcing_conditions`

src/eko/matching_conditions/n3lo/s_functions.py
```diff
@@ -0,0 +1,801 @@
...
+zeta3 = harmonics.zeta3
+zeta4 = harmonics.zeta4
+zeta5 = harmonics.zeta5
+
+li4half = 0.517479
+log2 = np.log(2)
+
+
+@nb.njit("c16(c16)", cache=True)
+def harmonic_Sm1(N):
```
felixhekhorn, line 24: I'd suggest to move the simple S_{-n} into the  `ad.harmonics` package since we need to reuse some of them for the ad - maybe we should even open a dedicated subpackage `eko/harmonics` with something like `single_letter`, containing `s_1,s_m1,...`, `tow_letter` with `s_21`, etc.

src/eko/matching_conditions/operator_matrix_element.py
AleCandido: Indeed, I still believe we have signatures because of cache, but in general specifying signatures is like declaring types in C, it should not be harmful.

The actual problem is: why caching yields AOT?
I'd like to compile JIT, and once compiled caching it, but not before.

src/eko/matching_conditions/n3lo/agg.py
```diff
@@ -0,0 +1,1556 @@
...
+    -------
+        A_gg_3 : complex
+            :math:`A_{gg}^{S,(3)}(N)`
+
+    See Also
+    --------
+        A_ggTF2_3: eko.matching_conditions.n3lo.aggTF2.A_ggTF2_3
+            Incomplete part proportional to :math:`T_{F}^2`.
+    """
+    S1, S2, S3, S4 = sx[0], sx[1], sx[2], sx[3]
```
felixhekhorn, line 45: now, I guess it is too late (though I'm not sure since this affects only the function head), but rather then this cryptic lines with random ordering, I rather would have made 3 arguments:
- `sx` a list with 11 elements: `s_{-5},...,s_{-1},nan,s_1,...s_5`
- `sxx` a 11x11 array: `s_{-5,-5},...` where most of the elements never get computed, but are simply nan or 0
- `sxxx` same in 3D

AleCandido: If possible [`np.nan`](https://numpy.org/doc/stable/reference/constants.html?highlight=nan#numpy.nan), since it has a more explicit meaning.

src/eko/matching_conditions/n3lo/f_functions.py
```diff
@@ -0,0 +1,6980 @@
...
+
+Mellin transform is defined with the convention x^(N).
+F19, F20, F21 are not present explicitly in the paper
+"""  # pylint: disable=line-too-long
+import numba as nb
+import numpy as np
+
+
+@nb.njit("c16(c16,c16,c16,c16)", cache=True)
+def F17(n, S1, S2, S3):
```
felixhekhorn, line 17: I'd suggest to split each function in a separate module

felixhekhorn:
Since we decided in #95 to drop the `_pto` prefix, let's extend this also to OMEs: https://github.com/N3PDF/eko/blob/f93c78c602b9c19399bbaff205b5ae133f59217e/src/eko/matching_conditions/nlo.py#L17

(do we want to prepare for new matching conditions and also rename the modules á là `as1` ?)

src/eko/matching_conditions/n3lo/agg.py
felixhekhorn: again, if you wish you can also separate by weight, but then please define a unique ordering somewhere, because by weight the sorting is not straight (you also can define an algorithm á là "biggest positive first + 1 number (31), next biggest + 1 number (22), then +2 number (211)" - but as said is a bit more trickier .. )

src/eko/matching_conditions/n3lo/f_functions.py
felixhekhorn: Mainly, yes - and to get rid of the 7k lines file, and better maintainability (imagine someone discover Fxy is wrong by tomorrow)

src/eko/matching_conditions/n3lo/agg.py
felixhekhorn: And *if* you go for weights, please adjust the names: `s3x -> s_w3` (or similar)

src/eko/matching_conditions/n3lo/f_functions.py
AleCandido: Then you can make a subpackage:
```
f_functions/
  __init__.py
  f17.py
  ...
```

giacomomagni: okay this is fine

AleCandido:
> (do we want to prepare for new matching conditions and also rename the modules á là `as1` ?)

On one side is fine, it will be consistent with the already existing modules. On the other, it's not strictly needed to be in this PR, so do you as you wish :)

AleCandido:
> (it seems github is overloaded by our PR - he doesn't show your latest commit so I have to just comment (hoping I can because he even refused that before))

GitHub has [its own troubles](https://www.githubstatus.com/incidents/sksd097hm0y5)

src/eko/matching_conditions/n3lo/agg.py
giacomomagni: you are right, the order I had in mind was something like by weights but I too creative... So the more logical would be the one with arrays matrices and tensors, the more physical one the one with weights...

src/eko/matching_conditions/n3lo/s_functions.py
giacomomagni: yes indeed the separation by weights is a bit more convenient, since the `w+1` can be sometimes contain some `w` sums ...

giacomomagni:
> > (it seems github is overloaded by our PR - he doesn't show your latest commit so I have to just comment (hoping I can because he even refused that before))
> 
> GitHub has [its own troubles](https://www.githubstatus.com/incidents/sksd097hm0y5)

src/eko/matching_conditions/operator_matrix_element.py
giacomomagni: okay, we need a better trick than the simple import statement delay... 
I think we are running in `nopython ` when we use `njit`, no?

AleCandido: Yes, `nb.njit` should be just an alias for `nb.jit(nopython=True)`

AleCandido:
@giacomomagni I'd suggest to go all-in on `enum`s: instead of comparing with values (`Modes.whatever.value`), that are essentially random
- assign them randomly, by using `enum.auto()`, e.g.:
  ```py
  class Ciao(enum.Enum):
      come = enum.auto()
      va = enum.auto()
  ```
  such that there is no unique translation to `int` (they are meaningless)
- convert read values to the `enum` as soon as possible:
  ```py
  mode = geattr(Modes, serialized)
  ```
- serialize by name, to be consistent with the former one:
  ```py
  serialized = mode.name
  ```

giacomomagni:
> @giacomomagni I'd suggest to go all-in on `enum`s: instead of comparing with values (`Modes.whatever.value`), that are essentially random

Okay, i'll try to implement your suggestion, the only constrain is that we have to pass to `quad_ker` an integer/float because of `numba`

AleCandido:
> Okay, i'll try to implement your suggestion, the only constrain is that we have to pass to `quad_ker` an integer/float because of `numba`

Yes, this is perfectly fine. Maybe a proper solution would be to use an [`IntEnum`](https://docs.python.org/3/library/enum.html#enum.IntEnum), maybe this way `numba` can use them as `int`.
I'm not 100% sure it's going to work, if not just keep calling `.value` whenever you have to pass through `numba`.

src/eko/evolution_operator/__init__.py
```diff
@@ -185,22 +184,22 @@ def quad_ker(
     if ker_base.is_singlet:
         gamma_singlet = ad.gamma_singlet(order, ker_base.n, nf)
         # scale var A is directly applied on gamma
-        if sv_mode == sv.Modes.exponentiated.value:
+        if sv_mode == sv.Modes.exponentiated:
```
AleCandido, line 187: Another suggestion on `enum`s (I'm really sorry to be annoying 😞), but the comparison should be done with `is`:
```suggestion
        if sv_mode is sv.Modes.exponentiated:
```
since the variants are singletons, so there is always a unique copy (like `None`).

giacomomagni: No, problem at all, your comments are always welcome :)
Unfortunately this isn't accepted because `numba` automatically cast `IntEnum`  to `int`. Maybe it will work if the signatures will be removed.

AleCandido: Ok, it makes sense: it works for `numba` because it can recast them not as objects.

Actually, I'd have expected it to work even for `int`s, since even integers are singleton in Python (there should exist only one `1`):
```py
i = 1
j = 2 - 1
assert i is j
```
But maybe it's just `numba`'s fault once more: the moment they become `numba` integers, they are not any longer Python objects, so it's not safe to use `is` any longer.

Better to keep always `==` inside `numba`.

AleCandido: If all occurrences of `==` are inside `numba` functions, mark this suggestion as solved.

src/eko/matching_conditions/operator_matrix_element.py
giacomomagni: Will be fixed in #110

src/eko/mellin.py
```diff
@@ -222,8 +222,8 @@ class Path:
 
     def __init__(self, t, logx, axis_offset):
         self.t = t
-        # TODO: shall we use: 0.4 * 16.0 / (1.0 - logx) ?
-        self.r = 0.4 * 16.0 / (-logx)
+        # TODO: shall we use: 0.4 * 16.0 / ( - logx) ?
+        self.r = 0.4 * 16.0 / (1.0 - logx)
```
AleCandido, line 226: I guess this has been tested, are the benefits documented somewhere? It might be useful for the future.

Consider writing something either here in the docstring, or in the Mellin page of the docs.

felixhekhorn:
What are we missing to close this PR?

tests/eko/test_g_functions.py
```diff
@@ -0,0 +1,113 @@
...
+log2 = h.constants.log2
+
+
+testN = [1, 10, 100]
+
+# compare the exact values of some harmonics with Muselli parametrization
+def test_g3():
+    ns = [1.0, 2.0, 1 + 1j]
+    # NIntegrate[x^({1, 2, 1 + I} - 1) PolyLog[2, x]/(1 + x), {x, 0, 1}]
+    mma_ref_values = [0.3888958462, 0.2560382207, 0.3049381491 - 0.1589060625j]
```
AleCandido, line 18: Stupid consistency remark: what here is called `mma_ref_values` everywhere else is `refvals`

AleCandido (commented):
From now on, we should refrain to address multiple big tasks in a single PR:

- the `harmonics` branch should have not been merged in here, because it has been reviewed twice in this way
- the `numba` cache enhancement was really needed, and even related to this PR, but it has been scattered all over, blowing up the number of files, so we should have done in a separate one (and merged immediately on develop, updating afterwards this, or otherwise merge this one before, and next apply the `numba` chore from a second PR, we have `develop` branch exactly for this reason...)

P.S.: of course, I didn't check details and numbers of every single function, I checked tests and relying on them, but MMa numbers I had no way to check, just relying on you

AleCandido:
The only relevant point is documenting Mellin path choice https://github.com/N3PDF/eko/pull/83#discussion_r869079192, but we can even postpone and just merge, at this point we have to do it as soon as possible.

felixhekhorn:
I suggest to close and instead postpone the documentation to #116

AleCandido:
I also updated dependencies with Poetry, so everything is alright.

I leave up to @giacomomagni the task to resolve conflicts with `develop`, and the pleasure to hit the green button.

Issue #84, opened 2022-01-17
Call anomalous mass dimension \gamma_m


felixhekhorn:
To avoid confusion, let's call he anomalous mass dimension `\gamma_m` (in contrast to the anomalous PDF dimension that we keep at `\gamma`). This has to be changed:
- in the [docs](https://eko.readthedocs.io/en/develop/theory/pQCD.html)
- in the [paper](https://github.com/N3PDF/papers/blob/e20c79f90724d66251efe416c7c32a80b01668e4/eko/sections/phenomenology/msbar-masses.tex)
- in the code (if used anywhere ... I think, gamma function and anomalous dimension are fine as names ...)

giacomomagni:
Okay I agree. Will take care of fixing this.

giacomomagni:
I noticed that we do have all the ingredients for easily adding also the N3LO msbar evolution. Shall I open a PR?

felixhekhorn:
I guess later we want to go full N3LO - so I'd say go ahead ...

PR #85, opened 2022-01-17
N3LO msbar and alpha_s

giacomomagni:
- [x] Allow for N3LO msbar running and fix issue #84. 

- [x] Allow for N3LO `alpha_s` evolution:


doc/source/theory/pQCD.rst
```diff
@@ -127,24 +127,28 @@ In doing so EKO takes advantage of the monotony of the |RGE| solution
...
 solving the previous integral and finally compute the evolved
-:math:`m_{\overline{MS},h}`. In fact, the function :math:`\gamma(a_s)` is the
+:math:`m_{\overline{MS},h}`. In fact, the function :math:`\gamma_m(a_s)` is the
 anomalous QCD mass dimension and, as the :math:`\beta` function, it can be evaluated
-perturbatively in :math:`a_s` up to :math:`\mathcal{O}(a_s^3)`:
+perturbatively in :math:`a_s` up to :math:`\mathcal{O}(a_s^4)`:
 
 .. math ::
-    \gamma(a_s) &= - \sum\limits_{n=0} \gamma_n a_s^{n+1} \\
+    \gamma_m(a_s) &= \sum\limits_{n=0} \gamma_{m_n} a_s^{n} \\
```
AleCandido, line 135: Maybe just `\gamma_{m,n}`?

doc/source/theory/pQCD.rst
```diff
@@ -127,24 +127,28 @@ In doing so EKO takes advantage of the monotony of the |RGE| solution
...
 
-Even here it is useful to define :math:`c_k = \gamma_k/\beta_0, k>0`.
+Even here it is useful to define :math:`c_k = \gamma_{m_k}/\beta_0, k \ge 0`.
```
AleCandido, line 137: And so on...

felixhekhorn:
> Tests will fail until N3LO alpha_s is not allowed.

I'd say maybe let's merge on top of #83 instead of develop and maybe even rebase on that

giacomomagni:
> > Tests will fail until N3LO alpha_s is not allowed.
> 
> I'd say maybe let's merge on top of #83 instead of develop and maybe even rebase on that

I thought #83 was already big enough and better have small PR...
What if we add also N3LO `alpha_s` here? I have to add `beta_3` in any case to complete this and what we only missing are these expression: [matching_coeffs_up](https://github.com/N3PDF/eko/blob/feature/N3LO_msbar/src/eko/strong_coupling.py#:~:text=def%20compute_matching_coeffs_up(mass_scheme)%3A) (if available..., I checked the refs quickly but was not able to find them straight )

felixhekhorn:
> I thought #83 was already big enough and better have small PR... What if we add also N3LO `alpha_s` here? I have to add `beta_3` in any case to complete this and we are missing only these expression:
> 
>     * [matching_coeffs_up](https://github.com/N3PDF/eko/blob/feature/N3LO_msbar/src/eko/strong_coupling.py#:~:text=def%20compute_matching_coeffs_up(mass_scheme)%3A) (if available..., I checked the refs quickly but was not able to find them straight )
> 
>     * the expanded solution at N3LO

maybe that's a good idea (and it even makes the other PR smaller ...)

giacomomagni:
Hi @felixhekhorn, 
I found the values in this two references: [Chetyrkin:1997sg]( https://arxiv.org/pdf/hep-ph/9706430.pdf) (eq 6 for MSbar and eq 10 for pole, there is a factor of 4 (and 4^2 ...) and a minus sign of difference in the log) and in [Vogt:2004ns](https://arxiv.org/pdf/hep-ph/0408244.pdf) (eq 2.43). 
Furthermore this values were also reported in [apfel](https://github.com/scarrazza/apfel/blob/532b77ca9f99863134179705e66305150429ad46/src/Evolution/a_QCD.f#L122).  
However while I was double checking the values I found that there might be some inconsistencies: what do you think? 

I attach here my understanding of the situation: 

```
matching_coeffs_up = np.zeros((4, 4))
    if mass_scheme == "MSBAR":
        matching_coeffs_up[2, 0] = -22.0 / 9.0
        
        # TODO: here apfel and Chetyrkin:1997sg (eq 6) do not agree...
        # Apfel has: matching_coeffs_up[2, 1] = 22.0 / 3.0 while the other report same coeff for pole and |MSbar|
        
        matching_coeffs_up[3, 0] = (
            (82043.0 / 27648.0) * zeta3 - 564731.0 / 124416.0 + 2633.0 / 31104.0 * nf
        ) * 4 ** 3
        matching_coeffs_up[3, 1] = 6793.0 / 27.0 - 281.0 / 27.0 * nf

    elif mass_scheme == "POLE":
        matching_coeffs_up[2, 0] = 14.0 / 3.0

        # TODO: I haven't checked this value yet.
        matching_coeffs_up[3, 0] = 340.729 - 16.7981 * nf
        # TODO: this value is not in agreement between Chetyrkin:1997sg and  Vogt:2004ns
        # Chetyrkin:1997sg: matching_coeffs_up[3, 1] = 8512.0 / 27.0 - 409.0 / 27.0 * nf
        matching_coeffs_up[3, 1] = 8941.0 / 27.0 - 409.0 / 27.0 * nf

    matching_coeffs_up[1, 1] = 4.0 / 3.0 * constants.TR
    matching_coeffs_up[2, 1] = 38.0 / 3.0
    matching_coeffs_up[2, 2] = 4.0 / 9.0
    matching_coeffs_up[3, 3] = 8.0 / 27.0

    # TODO: this value is not in agreement between Chetyrkin:1997sg and  Vogt:2004ns
    # Chetyrkin:1997sg has matching_coeffs_up[3, 2] = 131.0 / 9.0
    matching_coeffs_up[3, 2] = 511.0 / 9.0
```

src/eko/strong_coupling.py
```diff
@@ -445,8 +478,8 @@ def compute_matching_coeffs_down(mass_scheme, nf):
     .. code-block:: Mathematica
 
         Module[{f, g, l, sol},
-            f[a_] := a + Sum[d[n, k]*L^k*a^(1 + n), {n, 3}, {k, 0, n}];
-            g[a_] := a + Sum[c[n, k]*L^k*a^(1 + n), {n, 3}, {k, 0, n}] /. {c[1, 0] -> 0};
+            f[a_] := a + Sum[d[n, k)*L^k*a^(1 + n), {n, 3}, {k, 0, n}];
+            g[a_] := a + Sum[c[n, k)*L^k*a^(1 + n), {n, 3}, {k, 0, n}] /. {c[1, 0] -> 0};
```
felixhekhorn, line 482: I guess that change was not intended, as I think this is not valid MMa code ...

doc/source/theory/pQCD.rst
```diff
@@ -23,14 +23,29 @@ We implement two different strategies to solve the |RGE|:
...
+    L_{\mu} &= \ln(\mu_R^2/\mu_0^2) \\
+    L_{\text{LO}} &= \ln(a^{\text{LO}}_s(\mu_R^2)) \\
+    L_{\alpha_s} &= \ln \left( \frac{a^{\text{LO}}_s(\mu_R^2)}{(1 + b_0 L_{\mu} a^{\text{LO}}_s(\mu_R^2)} \right) \\
 
 When the renormalization scale crosses a flavor threshold matching conditions
 have to be applied :cite:`Schroder:2005hy,Chetyrkin:2005ia`.
-
+In particular the matching when passing form :math:`n_f` to :math:`n_f-1` schemes 
+is presented in equation 3.1 of :cite:`Schroder:2005hy` for |MSbar| masses, while the
+same expression for POLE masses is reported in Appendix A.
```
felixhekhorn, line 48: Most likely you need a new line here to separate the heading below

felixhekhorn:
- Please add a docstring here https://github.com/N3PDF/eko/blob/6139861c08e71e990f1aef0d849bab8f804701fe/src/eko/gamma.py#L13
similar to this one
https://github.com/N3PDF/eko/blob/6139861c08e71e990f1aef0d849bab8f804701fe/src/eko/beta.py#L15
- you can add numba to gamma

src/eko/msbar_masses.py
```diff
@@ -276,18 +297,8 @@ def compute_msbar_mass(theory_card):
     masses = np.concatenate((np.zeros(nfa_ref - 3), np.full(6 - nfa_ref, np.inf)))
     fact_to_ren = theory_card["fact_to_ren_scale_ratio"] ** 2
 
-    # TODO: why is this different than calling
```
giacomomagni, line 251: For the records: there were two typos:
 * `hqm_scheme="MSBAR"`  was somehow missing
 * `masses`were not divided by `fac_to_ren` which could have led to possible mistakes in the future... 
 Thanks for spotting this @felixhekhorn

src/eko/strong_coupling.py
```diff
@@ -465,6 +465,24 @@ def compute_matching_coeffs_down(mass_scheme, nf):
...
+            Heavy quark mass scheme: "POLE" or "MSBAR"
+        nf:
+            number of active flavors in the lower patch
+
+    Returns
+    -------
+        matching_coeffs_down:
+            downward matching coefficient matrix
+    """
+    _c = compute_matching_coeffs_up(mass_scheme, nf)
```
felixhekhorn, line 480: the prepended underscore is not needed

src/eko/msbar_masses.py
```diff
@@ -262,17 +263,83 @@ def evolve_msbar_mass(
...
+    ----------
+        nf:
+            number of active flavors in the lower patch
+
+    Returns
+    -------
+        matching_coeffs_down:
+            downward matching coefficient matrix
+    """
+    _c = compute_matching_coeffs_up(nf)
```
felixhekhorn: the prepended underscore is not needed

tests/benchmark_msbar_evolution.py
```diff
@@ -125,3 +146,102 @@ def benchmark_APFEL_msbar_evolution(self):
...
+                apfel.SetVFNS()
+                apfel.SetMSbarMasses(
+                    theory_dict["mc"], theory_dict["mb"], theory_dict["mt"]
+                )
+                apfel.SetMassScaleReference(
+                    theory_dict["Qmc"], theory_dict["Qmb"], theory_dict["Qmt"]
+                )
+                apfel.SetRenFacRatio(theory_dict["fact_to_ren_scale_ratio"])
+                apfel.InitializeAPFEL()
+                apfel.EnableWelcomeMessage(1)
```
felixhekhorn, line 185: of course you need to recompute `apfek_vals` here - as we're doing here: https://github.com/N3PDF/eko/blob/d02dbe839f02d43915f54650fe3dbad83a583d0b/tests/benchmark_strong_coupling.py#L130-L134

giacomomagni: So I was not able to extract the values of the computed quark masses with an apfel function... so I just copied them from the log message manually. I know it's an hack, the proper way would be to have https://github.com/N3PDF/external/blob/afd30d08f95ed96e4b03df21e0ecc1d284a7292e/apfel/src/Evolution/MSbarmass.f#L478 available in the python wrapper.

benchmarks/apfel_bench.py
```diff
@@ -94,7 +94,8 @@ def benchmark_msbar(self, pto):
         when  passing kthr != 1 both apfel and eko use ``kThr * msbar``,
         as thr scale, where ``msbar`` is the usual ms_bar solution.
         However apfel and eko mange the alpha_s thr differently
-        (apfel uses the given mass parameters as thr), so the
+        (apfel uses the given mass parameters as thr multipied by the kthr),
```
AleCandido, line 97: multipied -> multiplied

benchmarks/apfel_bench.py
```diff
@@ -94,7 +94,8 @@ def benchmark_msbar(self, pto):
...
+        while in eko only the already computed thr matters, so the
```
AleCandido, line 98: *while* sounds weird, since the sentence is opposed to is in parentheses: I'd suggest to remove the parentheses above

doc/source/theory/pQCD.rst
```diff
@@ -23,13 +23,29 @@ We implement two different strategies to solve the |RGE|:
...
+being: 
+
+.. math ::
+    L_{\mu} &= \ln(\mu_R^2/\mu_0^2) \\
+    L_{0} &= \ln(a_s(\mu_0^2)) \\
+    L_{\text{LO}} &= \ln(a^{\text{LO}}_s(\mu_R^2)) \\
 
 When the renormalization scale crosses a flavor threshold matching conditions
 have to be applied :cite:`Schroder:2005hy,Chetyrkin:2005ia`.
+In particular the matching when passing form :math:`n_f` to :math:`n_f-1` schemes
```
AleCandido, line 46: form -> from

AleCandido: In particular -> In particular,

AleCandido: when passing -> involved in the change (or something like this)

tests/benchmark_msbar_evolution.py
felixhekhorn: okay, I see ... then I guess we can not do it

tests/benchmark_msbar_evolution.py
```diff
@@ -2,7 +2,7 @@
 """This module benchmarks MSbar mass evolution against APFEL."""
 import numpy as np
 
-from eko.msbar_masses import evolve_msbar_mass
+from eko.msbar_masses import compute_msbar_mass, evolve_msbar_mass, compute_msbar_mass
```
AleCandido, line 5: You're importing twice `compute_msbar_mass`.

Moreover, even here I'd simplify names and use the scope instead
```py
from eko import msbar_masses

msbar_masses.compute
msbar_masses.evolve
```
When possible I'd always use syntax to convey semantics, it's more meaningful and explicit (and from my personal point of view more beautiful, since you repeat less).

felixhekhorn:
@AleCandido unfortunately tests are currently failing due to banana ...

tests/test_msbar_masses.py
```diff
@@ -88,22 +88,22 @@ def test_compute_msbar_mass_VFNS(self):
...
-                dict(
-                    mc=1.0,
-                    mb=1.00001,
-                    Qmc=1.0,
-                    Qmb=1.00001,
-                )
-            )
-            compute_msbar_mass(theory_dict)
-        with pytest.raises(ValueError, match="masses need to be sorted"):
+        # with pytest.raises(ValueError, match="do not preserve the correct ordering"):
```
AleCandido, line 96: Why you're not testing this any longer?

src/eko/msbar_masses.py
```diff
@@ -31,9 +32,9 @@ def msbar_ker_exact(a0, a1, order, nf):
         ker: float
             Exact |MSbar| kernel:
 
-            ..math:
-                k_{exact} = e^{\int_{a_s(\mu_{h,0}^2)}^{a_s(\mu^2)} \gamma(a_s) / \beta(a_s) da_s}
-    """
+            .. math::
+                k_{exact} = e^{- \int_{a_s(\mu_{h,0}^2)}^{a_s(\mu^2)} \gamma_m(a_s) / \beta(a_s) da_s}
+    """  # pylint:disable=line-too-long
```
AleCandido, line 37: If possible break it in two lines instead.

src/eko/msbar_masses.py
```diff
@@ -143,33 +172,123 @@ def msbar_ker_dispatcher(q2_to, q2m_ref, strong_coupling, fact_to_ren, nf):
...
+    ----------
+        nf:
+            number of active flavors in the lower patch
+
+    Returns
+    -------
+        matching_coeffs_down:
+            downward matching coefficient matrix
+    """
+    _c = compute_matching_coeffs_up(nf)
```
AleCandido, line 221: Why `_c`?

`_` have a conventional meaning in python, and in any case this variable has a cryptic name: you're giving up one chance to explain better what you're doing.
If they are coefficients to go up, call them `coeffs_up`.

src/eko/msbar_masses.py
```diff
@@ -2,13 +2,14 @@
...
 from scipy import integrate, optimize
 
 from .beta import b, beta
 from .evolution_operator.flavors import quark_names
 from .gamma import gamma
-from .strong_coupling import StrongCoupling, strong_coupling_mod_ev
+from .strong_coupling import StrongCoupling, invert_matching_coeffs
 
 
 def msbar_ker_exact(a0, a1, order, nf):
```
AleCandido, line 15: Here and below: as already mentioned, you're in the `msbar_masses` module, so everything that is `msbar` and `masses` you can omit, is the default.
If you compute/evolve/do something on something else, that instead you should specify.

AleCandido (changes requested):

src/eko/gamma.py
```diff
@@ -40,7 +97,8 @@ def gamma_3(nf):
     )
 
 
-def gamma(order, nf):
+@nb.njit("f8(u1,u1)", cache=True)
+def gamma(order, nf: int):
```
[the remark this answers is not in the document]
giacomomagni: Mmm you are right, I just copied it from `beta.py`, then I din't changed to meke it even more inconsistent...

tests/benchmark_msbar_evolution.py
giacomomagni: True, this is a good point, I'll change it.

src/eko/gamma.py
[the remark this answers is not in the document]
AleCandido: Okay, so it's the correct time to revert them all :D
`sed` for `: int` (or simply use vscode).

AleCandido:
> @AleCandido unfortunately tests are currently failing due to banana ...

Why `banana` is related to the tests? It should only be relevant in benchmarks, isn't it?

src/eko/gamma.py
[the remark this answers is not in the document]
giacomomagni: okay, fine let's fix them all!

tests/test_msbar_masses.py
giacomomagni: I'd propose to drop this addition check, see https://github.com/N3PDF/eko/pull/85#discussion_r803488102

src/eko/msbar_masses.py
```diff
@@ -376,21 +443,25 @@ def sc(thr_masses):
             fact_to_ren,
         )
 
+    # TODO: this test seems to be quite hard to contradict
+    # what if we check just the mass ordering?
```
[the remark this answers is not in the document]
AleCandido: Okay, I see.
I'm not user because it's a bit involved, but for sure you thought more than me, so if you're convinced enough keep going.

giacomomagni:
> > @AleCandido unfortunately tests are currently failing due to banana ...
> 
> Why `banana` is related to the tests? It should only be relevant in benchmarks, isn't it?
 
It's probably due to `ekomark` not able to import `banana.cfg`

src/eko/strong_coupling.py
```diff
@@ -13,7 +13,7 @@
 import scipy
 
 from . import constants, thresholds
-from .beta import beta
+from .beta import b, beta
```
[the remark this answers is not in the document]
felixhekhorn: I guess that's the best solution

src/eko/output.py
```diff
@@ -9,9 +9,9 @@
 import tempfile
 import warnings
 
-import lz4.frame
 import numpy as np
 import yaml
+from lz4 import frame
```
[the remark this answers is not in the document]
felixhekhorn: if he still complains we can do a similar trick as below and import as `lz4_frame` ...

felixhekhorn:
> Why `banana` is related to the tests? It should only be relevant in benchmarks, isn't it?

> It's probably due to `ekomark` not able to import `banana.cfg`

yes, to be more correct it is not the tests but the linter failing on ekomark

felixhekhorn:
Actually, is it a version problem, because we need banana@develop?

AleCandido:
For sure we need `develop` at the moment, but I'll make soon another release (as soon as I'll have the time to implement a proper  solution, instead of a temporary hack).

If we're not in a hurry to merge this, let's wait a second :)

src/eko/output.py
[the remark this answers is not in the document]
giacomomagni: Okay I agree with `from lz4 import frame as lz4_frame `

giacomomagni:
> For sure we need `develop` at the moment, but I'll make soon another release (as soon as I'll have the time to implement a proper solution, instead of a temporary hack).
> 
> If we're not in a hurry to merge this, let's wait a second :)

Sure, no problem. Thanks for your review and inputs, @AleCandido , @felixhekhorn

AleCandido (approved):

src/eko/strong_coupling.py
```diff
@@ -67,20 +68,55 @@ def as_expanded(order, as_ref, nf, scale_from, scale_to):
...
+        if order >= 2:
+            b2 = beta_b(2, nf)
             res = as_LO * (
                 1.0
                 + as_LO * (as_LO - as_ref) * (b2 - b1**2)
                 + as_NLO * b1 * np.log(as_NLO / as_ref)
             )
-
+            # N3LO expansion is taken from Luca Rottoli and it's simpler
+            # than the APFEL expansion which is more accurate
```
AleCandido, line 83: Can you quantify the level of discrepancy with APFEL?  And what's the difference that makes you call it "more accurate"?
(I expect something like "the expansion order", that if it's ridiculously large I agree we can keep this one)

tests/benchmark_msbar_evolution.py
```diff
@@ -23,79 +45,201 @@ def benchmark_APFEL_msbar_evolution(self):
...
-                apfel_vals, np.sqrt(np.array(my_vals)), rtol=2e-3
+                apfel_vals, np.sqrt(np.array(my_masses)), rtol=5e-5
             )
+
+    def benchmark_msbar_solution_kthr(self):
+        """
+        With this test you can see that in EKO
+        the solution value of mb is not affected by "kbThr",
+        since mb is searched with an Nf=5 larger range.
+        While in Apfel this doesn't happend
```
AleCandido, line 196: happend -> happen

tests/test_interpolation.py
```diff
@@ -240,20 +240,20 @@ def test_log_eval_N(self):
             assert_almost_equal(act_c, res_c)
         # Full -> \tilde p_0(N) = exp(-N)(exp(N)-1-N)/N^2
         # MMa: Integrate[x^(n-1) (-Log[x]),{x,1/E,1}]
-        p0Nref_full = lambda N, lnx: ((np.exp(N) - 1 - N) / N**2) * np.exp(
+        p0Nref_full = lambda N, lnx: ((np.exp(N) - 1 - N) / N ** 2) * np.exp(
```
AleCandido: Here you are using an outdated version of `black`, since 22.1.0 will collapse spaces around `**`, update it in your VS code.

Then I also wonder they have not been fixed by `pre-commit`, might be possible that you forgot to install it?

src/eko/strong_coupling.py
giacomomagni: This was actually a @felixhekhorn guess, but I think the comment can be dropped...

felixhekhorn: The comment is a bit misleading ...
-  the N3LO is coming from Luca and I don't believe there is such a thing in Apfel
- the "more accurate" should refer to NNLO - since we take that from Apfel and not from Luca
- than "more accurate" is what has been discussed maybe around ~June 20 (searching in the minutes): APFEL does not implement the ordinary expansion (as done by Luca), but an advanced expression that a) we don't have an explanation for and b) solves the RGE better, i.e. the subleading terms are numerically smaller (this is also discussed in one of my off-line MMa noteboooks)

AleCandido: Ok, fine. So, no discussion on N3LO, and we don't need to discuss NNLO here (that in any case is something we've already done).

AleCandido: @giacomomagni or @felixhekhorn, remove or improve the comment, please

giacomomagni: I'll drop it!

tests/test_interpolation.py
giacomomagni: I must have messed up something while merging with develop, thanks for fixing

giacomomagni:
Shall we merge ?

AleCandido:
As of N3PDF/banana@1e1b576 `banana` has been merged, we don't need `develop` any longer.

felixhekhorn:
Can we then merge?

AleCandido:
I would revert `src/eko/output.py` (personally, I prefer to preserve original names, where possible).

I checked everything else, and I'm fine with it.

giacomomagni:
> I would revert `src/eko/output.py` (personally, I prefer to preserve original names, where possible).
> 
> I checked everything else, and I'm fine with it.

then we keep ´pylint´ complaining?

AleCandido:
I don't see it complaining on CodeFactor: if it were, it shoud be listed here https://www.codefactor.io/repository/github/n3pdf/eko/pull/fixed/85

Which version of `pylint` do you have? What is complaining about?

giacomomagni:
> I don't see it complaining on CodeFactor: if it were, it shoud be listed here https://www.codefactor.io/repository/github/n3pdf/eko/pull/fixed/85
> 
> Which version of `pylint` do you have? What is complaining about?

Indeed codeFactor is complaining about `import. yaml`:
 `third party import "import yaml" should be placed before "import lz4.frame`
 that's why I originally modified this. [see here]( https://www.codefactor.io/repository/github/n3pdf/eko/issues/develop?category=Maintainability&groupId=499)

AleCandido:
Okay, then the solution is not to change the name: we need to reconcile `isort` with `pylint` (otherwise this problem might come back over and over).

There are two possible ways of doing it:
- we select/define an `isort` profile that is compliant with `pylint` (but we already selected one that is compatible with `black`, and we change too much we have to maintain ourselves), i.e. *configure `isort`*
- we prevent `pylint` to complain, whatever `isort` does is fine: we're giving up on the check, but on the other hand we are all using `isort`, so we can also *configure `pylint`* and give up on the double check

AleCandido:
https://github.com/PyCQA/pylint/issues/3817#issuecomment-687892090

giacomomagni:
> At the end of the day, they come from the same place [PyCQA/pylint#3817 (comment)](https://github.com/PyCQA/pylint/issues/3817#issuecomment-687892090)

Yes I agree this is the origin of this issue, so may solution was just a workaround. I can revert it and we solve this on a later stage properly.

felixhekhorn:
> Yes I agree this is the origin of this issue, so may solution was just a workaround. I can revert it and we solve this on a later stage properly.

agreed - after that you can merge :+1:

PR #91, opened 2022-02-07
Scale Variations Expanded

giacomomagni:
- [x] Move scale variation to a subpackage
- [x] implementation of scheme B (`ModSV=expanded`)
- [x] Generalise quad ker
- [x] Generalise Operator class (try parallel integration) 
- [x] Benchmark: Pegasus/Lha `ModSV=exponentiated`
- [x] Benchmark: Apfel `ModSV=exponentiated`. Bug found and fixed.


benchmarks/pegasus_bench.py
```diff
@@ -79,7 +79,8 @@ def benchmark_sv(self, pto, scheme):
         th.update(
             {
                 "PTO": [pto],
-                "fact_to_ren_scale_ratio": [np.sqrt(0.5), np.sqrt(2.0)],
+                # TODO: why do we need this epsilon to have a good match ??
+                "fact_to_ren_scale_ratio": [np.sqrt(0.5),np.sqrt(2.0-1e-8)],
```
AleCandido, line 83: Good point.

If you find by yourself that's perfect, but maybe we should have a look together.

AleCandido:
Hi @giacomomagni, I like a lot the way you're structuring it :D

About modules naming: since it's not ambiguous inside `scale_variations` subpackage, I'd call `scheme_A.py` just `a.py`, keeping the docstring inside the module telling that is *scheme A* implementation. It would look like 
```py
import scale_variations.a
```
that is not that bad.

On the other hand, it is always true that explicit is better than implicit, so if you don't like you can also keep `scheme_a`, but at least make it lowercase to respect conventions :)

giacomomagni:
> Hi @giacomomagni, I like a lot the way you're structuring it :D
> 
> About modules naming: since it's not ambiguous inside `scale_variations` subpackage, I'd call `scheme_A.py` just `a.py`, keeping the docstring inside the module telling that is _scheme A_ implementation. It would look like
> 
> ```python
> import scale_variations.a
> ```
> 
> that is not that bad.
> 
> On the other hand, it is always true that explicit is better than implicit, so if you don't like you can also keep `scheme_a`, but at least make it lowercase to respect conventions :)
> 
> P.S.: same thing goes for B, of course

I agree, I like the first option as well!

benchmarks/pegasus_bench.py
giacomomagni: Indeed it was the same trick needed for the LHA tables. Now bench with `scheme a` (the only one available in pegasus) it's perfect!

AleCandido: Very good, thank you :D

benchmarks/apfel_bench.py
```diff
@@ -5,7 +5,7 @@
 import numpy as np
 from banana.data import cartesian_product
 
-from ekomark.banana_cfg import register
+from ekomark import register
```
felixhekhorn, line 8: I think  `register` moved into banana: https://github.com/N3PDF/banana/commit/b51feaa2e2d31f41b73ec30f421fc36e08d79f9b

src/eko/evolution_operator/grid.py
```diff
@@ -276,6 +278,19 @@ def generate(self, q2):
...
+                self.config, self.managers, path[-1].nf, final_op.q2_final
+            )
+            scale_var.compute()
+            scale_var_op = physical.PhysicalOperator.ad_to_evol_map(
+                scale_var.op_members,
+                scale_var.nf,
+                scale_var.q2_to,
+                intrinsic_range,
+             )
+            final_op = scale_var_op @ final_op
```
felixhekhorn, line 293: I wonder whether we need to do this path by path - instead of only at the end ... see scheme A: here we're affected patch by patch and I think we have to continue to do so; should we ask for a confirmation?

benchmarks/apfel_bench.py
giacomomagni: I see, but then we need to remove this: 
https://github.com/N3PDF/eko/blob/6accf8b20333d3b56a0b12c40a5e181191be58cb/src/ekomark/__init__.py#L13

doc/source/theory/pQCD.rst
```diff
@@ -46,23 +46,55 @@ Note the additional minus in the definition of :math:`\gamma`.
...
+        1 - a_s L \gamma^{(0)} + a_s^2 \left [ L \gamma^{(1)} + \frac{1}{2} L^2 \gamma^{(0)} (\gamma^{(0)}+\beta_0) \right ]
 
-while keeping the evaluation of the strong coupling always at :math:`\mu_R^2`.
-Estimating the theoretical uncertainties imposed on |PDF| determination due to
-missing higher order corrections using scale variation in the evolution
-corresponds to schemes A and B in :cite:`AbdulKhalek:2019ihb`.
+    In this way the dependence of the |EKO| on :math:`L` is factorized outside the unvaried evolution kernel.
 
+Notice that the two methods should be equivalent, especially for fully linearized solutions
+(`ModEv=truncaded,ev_op_max_order=1`)
```
felixhekhorn, line 97: `truncated` doesn't use `ev_op_mak_order`, does it? it is linear by construction ...

benchmarks/apfel_bench.py
AleCandido: Yes, I initially fixed for `ekomark` only, the moment I had to do it for `yadmark` I did the proper thing in `banana`, but `ekomark` has still to be updated.

src/eko/evolution_operator/grid.py
AleCandido: Uhm... I'd say that what we split from the exponential is only in the last patch, because the patches are in the argument of the exponential itself.
But we might discuss.

doc/source/theory/pQCD.rst
AleCandido: Nope, `TRN` does use `ev_op_max_order`, and indeed we found that it was a major flaw in the comparison. Furthermore, at the time we found that the correct value was `ev_op_max_order=2`, and because of that we corrected a bug in the counting, now the correct value is actually `1`.

felixhekhorn: Nope again ;-)
- the bug was on `ev_op_max_iterations`: we needed to do the whole evolution in a singlet step per patch
- instead: `ev_op_max_order` is used in the `perturbative-*` strategies of the singlet to determine the maximum coefficient of `U` https://eko.readthedocs.io/en/latest/theory/DGLAP.html#nlo-singlet-evolution https://github.com/N3PDF/eko/blob/d02dbe839f02d43915f54650fe3dbad83a583d0b/src/eko/kernels/singlet.py#L314

giacomomagni: Okay, I confused them as well, so I meant `ev_op_max_iterations`

src/eko/evolution_operator/grid.py
giacomomagni: mmm but the actual SV kernel depends on `nf`, so it can make differences

src/eko/mellin.py
```diff
@@ -233,12 +233,15 @@ def __init__(self, t, logx, axis_offset):
...
         return Talbot_path(self.t, self.r, self.o)
 
     @property
     def jac(self):
+        """Returns the Jacobian of the Mellin path"""
         return Talbot_jac(self.t, self.r, self.o)
 
     @property
     def prefactor(self):
+        r"""Returns the mellin inversion prefactor :math:`-\frac{1}{\pi}`"""
```
felixhekhorn, line 246: it is `1/(i*pi) = - i/pi` (the text book value beeing `1/(2pi i)`, but we're even folding the integral https://eko.readthedocs.io/en/latest/theory/Mellin.html#mellin-space-and-transformations )

giacomomagni: yes! I forgot the `i`

src/eko/evolution_operator/__init__.py
```diff
@@ -45,6 +46,72 @@ def select_singlet_element(ker, mode):
...
 
+spec = [
+    ("is_singlet", nb.boolean),
+    ("is_log", nb.boolean),
+    ("logx", nb.float64),
+    ("u", nb.float64),
+]
+
+
+@nb.experimental.jitclass(spec)
```
AleCandido, line 56: Experimental sounds a bit worrying, but I guess it might be fine (if you maintain for a while...).

src/eko/evolution_operator/__init__.py
```diff
@@ -45,6 +46,72 @@ def select_singlet_element(ker, mode):
...
+class QuadKerBase:
+    """
+    Class to manage the common part of Mellin inversion integral
```
AleCandido, line 59: "Class to" is implicit, you can even skip (just write "Manage the...").

src/eko/evolution_operator/__init__.py
```diff
@@ -144,42 +160,32 @@ def quad_ker(
...
+    integrand = ker_base.integrand(areas)
+    if integrand == 0.0:
         return 0.0
+
     # compute the actual evolution kernel
-    if is_singlet:
-        gamma_singlet = gamma_singlet_fact(order, n, nf, L)
+    if ker_base.is_singlet:
+        gamma_singlet = ad.gamma_singlet(order, ker_base.n, nf)
+        if is_sv_scheme_a:
```
felixhekhorn, line 179: Actually, do we need this guard? or can't we rely on L=0? if ever, add an early exit in `gamma_fact`?

src/eko/evolution_operator/__init__.py
```diff
@@ -230,6 +231,27 @@ def __init__(self, config, managers, nf, q2_from, q2_to=None, mellin_cut=5e-2):
         self._mellin_cut = mellin_cut
         self.op_members = {}
 
+    @cached_property
+    def fact_to_ren(self):
+        """Returns the interpolation dispatcher"""
```
AleCandido, line 236: Copy pasted docstring

src/eko/evolution_operator/__init__.py
AleCandido: I thought that as well.

I'm not sure that's the case, but we might want to convert the string in a couple of integers (they have to be integers and not boolean, because in the QED case the coupled sector will be larger than 2x2, namely 4x4).

AleCandido: Even a `u1` should be fine, if available.

giacomomagni:
Hi, 
I tried to run the current implementation of `scheme B` for `VFNS` and unfortunately it's really slow and has quite large discrepancies with `scheme A`.  
I believe this is because the integrand of the variation it's really close to the identity ( or for the off diagonal singlet entries is close to 0) and this can affect both the speed and the numerical accuracy. 
So I propose that we apply the additional step directly on the evolution kernel (similarly to what we do on the `ad` for `scheme A`). 
Some of the implemented things regarding `quad_ker` can eventually be used for the matching conditions, so I would not revert them. 
What do you think @felixhekhorn, @AleCandido ?

AleCandido:
Fine for me, the only concern is if you can avoid to blow up the complexity of `quad_ker` or not.

If in a non-scale varied computation the overhead is negligible, and the complexity does not increase too much (undermining significantly maintainability), I see no further problem.

felixhekhorn:
> I believe this is because the integrand of the variation it's really close to the identity ( or for the off diagonal singlet entries is close to 0) and this can affect both the speed and the numerical accuracy.

did you by chance check this explicitly?

> So I propose that we apply the additional step directly on the evolution kernel (similarly to what we do on the `ad` for `scheme A`). 

so, if I understand correctly, what you suggest is to add after these lines: https://github.com/N3PDF/eko/blob/312fb36d7d17398872beeba694ee2fe4a415c033/src/eko/evolution_operator/__init__.py#L189-L197

```py
if is_scheme_b:
  ker = ker_fact(ker, order, nf, L)
```

as you said in analogy to A ...

I think that's a good idea - as in general, I think we should put as much under the Mellin integral as we can (at some point I was thinking to change EKO around to put *everything* under 1 integral - but it would maybe be a too heavy change ...)

> Some of the implemented things regarding `quad_ker` can eventually be used for the matching conditions, so I would not revert them. 

that we wanted to do in any case, so yes, keep them

giacomomagni:
> > I believe this is because the integrand of the variation it's really close to the identity ( or for the off diagonal singlet entries is close to 0) and this can affect both the speed and the numerical accuracy.
> 
> did you by chance check this explicitly?
> 
No, it was an educated guess, since NNLO was faster than NLO. I checked that the result was not consistent with `scheme A`

> > So I propose that we apply the additional step directly on the evolution kernel (similarly to what we do on the `ad` for `scheme A`).
> 
> so, if I understand correctly, what you suggest is to add after these lines:
> 
> https://github.com/N3PDF/eko/blob/312fb36d7d17398872beeba694ee2fe4a415c033/src/eko/evolution_operator/__init__.py#L189-L197
> 
> ```python
> if is_scheme_b:
>   ker = ker_fact(ker, order, nf, L)
> ```
> 
> as you said in analogy to A ...
> 
> I think that's a good idea - as in general, I think we should put as much under the Mellin integral as we can (at some point I was thinking to change EKO around to put _everything_ under 1 integral - but it would maybe be a too heavy change ...)
> 

yes something similar, it will be also faster.

> > Some of the implemented things regarding `quad_ker` can eventually be used for the matching conditions, so I would not revert them.
> 
> that we wanted to do in any case, so yes, keep them

good!

src/eko/evolution_operator/__init__.py
```diff
@@ -40,8 +42,8 @@ def select_singlet_element(ker, mode):
         ker : complex
             singlet integration kernel element
     """
-    k = 0 if mode[2] == "q" else 1
-    l = 0 if mode[3] == "q" else 1
```
AleCandido: Write explicitly this encoding in the docstring, and do it everywhere, otherwise it will be impossible to debug in a few weeks/months.

tests/test_ev_operator.py
```diff
@@ -15,6 +20,22 @@
 from eko.thresholds import ThresholdsAtlas
 
 
+def test_full_labels_dict():
+    for mode in ["S_qq", "S_qg", "S_gq", "S_gg"]:
+        assert full_labels_dict[mode] < 4
```
AleCandido, line 25: Test explicitly the encoding, one by one. At least we get a reference that is enforcing the thing.

src/eko/evolution_operator/__init__.py
```diff
@@ -195,7 +195,9 @@ def quad_ker(
         )
         # scale var B is applied on the kernel
         if sv_scheme == sv.scheme_B:
-            ker = b.singlet_variation(ker, gamma_singlet, a1, order, nf, L)
+            ker = np.ascontiguousarray(ker) @ np.ascontiguousarray(
```
AleCandido, line 198: If I understood correctly `np.ascontiguousarray` is just memory optimization: why do you need it here?
(I'd believe it to be more useful when you apply the function once, and you reuse the return value over and over)

doc/source/theory/pQCD.rst
```diff
@@ -92,6 +92,7 @@ evolution, always evaluating the strong coupling at :math:`\mu_R^2`.
         1 - a_s L \gamma^{(0)} + a_s^2 \left [ L \gamma^{(1)} + \frac{1}{2} L^2 \gamma^{(0)} (\gamma^{(0)}+\beta_0) \right ]
 
     In this way the dependence of the |EKO| on :math:`L` is factorized outside the unvaried evolution kernel.
+    This procedure is repeated for each different |FNS| present in the evolution path.
```
AleCandido, line 95: In some sense even a VFNS is a FNS. Maybe replacing with `|FFNS|` might be better?
I'm not actually sure, do what you prefer.

felixhekhorn: I'd say "scheme" is an overall treatment (meaning a global thing) - why don't we stick to our nomenclature "patch"?

src/eko/evolution_operator/__init__.py
felixhekhorn: I don't remember the exact circumstances, but I do remember, that I definitely needed that here: https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/eko/kernels/singlet.py#L338
(and similar places) otherwise numba wouldn't work correctly - and I also seem to remember that it was non-trivial to find ...

src/eko/evolution_operator/__init__.py
```diff
@@ -226,6 +251,32 @@ def __init__(self, config, managers, nf, q2_from, q2_to, mellin_cut=5e-2):
         self._mellin_cut = mellin_cut
         self.op_members = {}
 
+    @property
+    def fact_to_ren(self):
+        """Returns the factor (µ_F/µ_R)^2"""
```
felixhekhorn, line 256: change to `:math:`?

src/eko/scale_variations/b.py
```diff
@@ -0,0 +1,87 @@
...
+        sv_ker : numpy.ndarray
+            scale varion kernel
+    """
+    sv_ker = np.eye(2, dtype=np.complex_)
+    if order >= 1:
+        sv_ker -= a_s * L * gamma[0]
+    if order >= 2:
+        sv_ker += a_s**2 * (
+            -gamma[1] * L
+            + 1
```
felixhekhorn, line 79: Stupid thing: I believe with `+ (1/2) *` I believe you can trick black and make the thing a bit more readable ...

doc/source/theory/pQCD.rst
giacomomagni: Okay patch is fine!

src/eko/evolution_operator/__init__.py
giacomomagni: yeah, so numba complains when you do the product of two matrices... Now I added a minimal fix.

AleCandido:
We discussed with Stefano about schemes A and B, and in some sense, their definition is broader than `eko` alone.

Thus, I'd propose:
- [ ] to rename them in `eko`, with some more telling names, like `exponentiated` and `expanded`
- [ ] to add in the docs a discussion about (in the theory section) in which we explain the difference, and we tell that the consistent way would be to use the current `a` with `exact` solutions, and the current `b` with `truncated` solutions (but actually all the options are available, so do what you prefer)

Then, the scale variation entry in the theory will also be something like `ModEv` (we can call it `ModSV`, following the same cryptic scheme for names, but at least we're consistent) and it will be either `exponentiated` or `expanded`.

giacomomagni:
> We discussed with Stefano about schemes A and B, and in some sense, their definition is broader than `eko` alone.
> 
> Thus, I'd propose:
> 
> * [ ]  to rename them in `eko`, with some more telling names, like `exponentiated` and `expanded`
> * [ ]  to add in the docs a discussion about (in the theory section) in which we explain the difference, and we tell that the consistent way would be to use the current `a` with `exact` solutions, and the current `b` with `truncated` solutions (but actually all the options are available, so do what you prefer)
> 
> Then, the scale variation entry in the theory will also be something like `ModEv` (we can call it `ModSV`, following the same cryptic scheme for names, but at least we're consistent) and it will be either `exponentiated` or `expanded`.

Okay I can agree with this renaming convention. Do we want to add some warning in case of "wrong" usage?

AleCandido:
> Okay I can agree with this renaming convention. Do we want to add some warning in case of "wrong" usage?

Maybe?

I'm not sure which is my position about warnings, if we should be proactive and warn, or simply since we're able to do it just say nothing, and keep this for the docs.
In any case, I'm planning to add an `input` layer for `eko`, like we have for `yadism` (and improving the current one for `yadism` at the same time). I would keep this kind of warnings for there.

Once input layer is passed, you can do everything you wish, and it's part of the `eko` "application". If you use `eko` as a library, you can do whatever you want, and the library would have no responsibility to prevent it (it would be overly complicated, it would couple much more the code, and in any way it's not needed).

src/eko/evolution_operator/__init__.py
```diff
@@ -360,57 +439,30 @@ def compute(self):
...
 
-            logger.info(
-                "Evolution: computing operators - %d/%d took: %f s",
-                k + 1,
-                self.grid_size,
-                time.perf_counter() - start_time,
+        # TODO: which is the optimal numbers of pool to open?
+        # make it an external parameter?
+        # run integration in parallel for each grid point
+        with Pool(8) as pool:
```
AleCandido, line 452: I'm not sure that parallelization at this level is a good idea, we already have a couple of different ones concerning `eko` performance improvements, maybe at some point we can meet and discuss (however, just to mention: top level Q2 splitting, maybe you saw the issue, and bottom level `quad_vec` integration`).

In any case, do not hard-code 8: my laptop has 4 cores, and my desktop 12, simply leave it empty:
> processes is the number of worker processes to use. If processes is None then the number returned by [os.cpu_count()](https://docs.python.org/3/library/os.html#os.cpu_count) is used.

from [python docs](https://docs.python.org/3/library/multiprocessing.html?highlight=multiprocessing#module-multiprocessing.pool)

AleCandido (changes requested):

src/eko/evolution_operator/__init__.py
giacomomagni: I tried this way in order to generalise even more this class, and eventually make it compatible with the one performing the matching condition computation. Just with 8 pools it speeds up the code quite well. 
I think this will not contradict/prevent any of the other two improvements, but maybe I am mistaken...

> In any case, do not hard-code 8: my laptop has 4 cores, and my desktop 12, simply leave it empty:

Are you sure the numbers of pool is related to the number of cores you have?

AleCandido: The short answer is yes: if you have much more processes in your pools than processors, most of them are just waiting, and you only pay for maintenance overhead, if they are too few, you don't exploit parallelization as much as you can (think about having a single 1).

The best is not necessarily the number of course, because it's a bit more complicated, but for sure it's close to it. And that's why is also the default, otherwise it wouldn't make sens, don't you think so? ;)

AleCandido: In any case, don't put too much effort in this: python is known to be very bad in parallelization, because of the infamous GIL (Global Interpreter Lock, that is otherwise very useful). You can read a lot online about it, and there are several attempts to [GILectomy](https://pythoncapi.readthedocs.io/gilectomy.html), no one successful until now.

In general, middle level parallelization is not a clever idea: if you parallelize bottom level, you speed your primitives, and there is no risk of locking resources, if you do top level, you can split completely separate tasks. If you attempt to parallelize middle-level, most times the result is just suboptimal, and in some cases might even break (I believe this not to be the case).

src/eko/evolution_operator/__init__.py
```diff
@@ -446,10 +433,8 @@ def compute(self):
             self.config["method"],
         )
 
-        # TODO: which is the optimal numbers of pool to open?
-        # make it an external parameter?
         # run integration in parallel for each grid point
-        with Pool(8) as pool:
+        with Pool(int(os.cpu_count() / 2)) as pool:
```
AleCandido, line 437: Keep the default, it's good enough and much simpler

giacomomagni: Have you tried it?
I tried with the full number and thought was too much..

AleCandido: Fine, I was just saying to keep it simple, if it's slightly suboptimal we don't need.

I steel hope this to be needed only temporary.

giacomomagni:
Regarding this [comment](https://github.com/N3PDF/eko/blob/637e56c4fc82b3d53990f78696084056fb0c3dd8/src/eko/msbar_masses.py#L333): I I think `-n.log(fact_to_ren)`  should not be added. 

I'm not really sure why in `apfel` `kren` is used during the determination of the MS bar masses: [see here](https://github.com/scarrazza/apfel/blob/532b77ca9f99863134179705e66305150429ad46/src/Evolution/MSbarmass.f#L426),
 but not in the evolution routines: [see here for example](https://github.com/scarrazza/apfel/blob/532b77ca9f99863134179705e66305150429ad46/src/Evolution/MSbarmass.f#L65)

benchmarks/eko_bench.py
```diff
@@ -95,10 +95,21 @@ def benchmark_sv(self, pto):
...
-        high["ModSV"] = "A"
+        high["ModSV"] = "exponentiated"
         # needed for apfel
         high["XIR"] = np.sqrt(0.5)
         high["EScaleVar"] = 0
+
+        # here we need to adjust also the
+        # apfel initial nf, which we can't
+        # be accessed in other ways
+        # Be careful: this is and hack not a proper solution
```
felixhekhorn, line 106: 1) "a hack and" :see_no_evil: 
2) so we're only testing the bottom threshold?

AleCandido: Great that you finally succeeded! So bad APFEL..

I would rename the file to `scale_variations.py`. It is useful and I'd keep it, but `eko_bench.py` is a bit too generic...

felixhekhorn: but I believe @giacomomagni is using this in a generic way - which happens to work also for SV ... correct me if I'm wrong @giacomomagni

giacomomagni: Yes, I'm using it in a generic way, but actually I'm not satisfied by this trick. Because as you can see the problem is just shift on T24 and happens to be less visible. But this is **not** the solution, I believe.

AleCandido: Actually, I tried to run it, having reinstalled dependencies as they are in `poetry.lock`, but I get the following error:
```
ValueError: 48 columns passed, passed data had 49 columns
```
Clearly coming from db...

By chance, are you using your `banana` branch or something like this?

giacomomagni: Nope, did you pull? Because I fixed a typo `ModSv`, which was causing this issue,  in one of the lasts commit.

AleCandido: I pulled `eko`, but I was relying on the released `banana` version.

If you wish, this time let me just finish as soon as possible the current batch, and we definitely release a new version. However, in general, we should try to stick as much as possible to released versions.

I've the first one to use `develop` branches together, but we're getting more and more stable (and we are also more and more developers with more and more machines), so better to use as much as possible releases, otherwise it's a mess to reproduce the environment (while with releases `poetry` is really helping us a lot).

giacomomagni: Okay, got it! In this case it was not possible due to the additional label `ModSv`. But I agree with you.

felixhekhorn: yes, I'd say so - let's write the issue and forget

benchmarks/eko_bench.py
```diff
@@ -0,0 +1,126 @@
+# -*- coding: utf-8 -*-
```
felixhekhorn, line 1: Ok, I thought you were using this really in a "broader sense" - but since this is strictly related to the LHA setting, please merge this file file `lha_...` e.g. you can extend `BenchmarkVFNS` to allow other externals ... otherwise there is too much code repeated

giacomomagni: okay I can do it!

src/eko/evolution_operator/__init__.py
```diff
@@ -21,7 +21,6 @@
 from ..kernels import non_singlet as ns
 from ..kernels import singlet as s
 from ..member import OpMember
-from ..scale_variations import expanded, exponentiated
```
giacomomagni: @felixhekhorn this fix doesn't work, since ``scale_variations.__init__`` doesn't contain ```import expanded, exponentiated```.  Would you prefer to revert or add them to the `__init__`?

felixhekhorn: added imports in [88f1d6f](https://github.com/N3PDF/eko/pull/91/commits/88f1d6fd6ba5875a6d6085356eee98953d4aa5e9) - good you realized, still I got the green tick ...

felixhekhorn:
To run the script I wait for a new banana release, hopefully to appear soon (@AleCandido :upside_down_face: ), with the fix you mentioned on the other PR

AleCandido:
> To run the script I wait for a new banana release, hopefully to appear soon (@AleCandido upside_down_face ), with the fix you mentioned on the other PR

The release is available since yesterday evening, you can keep going now :)

felixhekhorn:
> The release is available since yesterday evening, you can keep going now :)

Actually, I had to release again: https://github.com/N3PDF/banana/releases/tag/v0.6.3 - since you put the fix on https://github.com/N3PDF/banana/pull/18

felixhekhorn:
@AleCandido while this https://github.com/N3PDF/eko/pull/91/commits/2efccf36fa657a2ee95f6ee9104b5e718f8315d5#diff-ea5511facb0ac5f8e956356edebc8d3f991a50722e4c304b4a1bfdf3606ba771 is working and makes navigator run again, it is clearly not the solution ...

AleCandido:
> @AleCandido while this [2efccf3#diff-ea5511facb0ac5f8e956356edebc8d3f991a50722e4c304b4a1bfdf3606ba771](https://github.com/N3PDF/eko/commit/2efccf36fa657a2ee95f6ee9104b5e718f8315d5#diff-ea5511facb0ac5f8e956356edebc8d3f991a50722e4c304b4a1bfdf3606ba771) is working and makes navigator run again, it is clearly not the solution ...

I would suggest providing `banana.yaml` in a temporary folder, completely unrelated to benchmark one, and to delete it and results right after testing.
You can provide the full thing as a [`pytest.fixture`](https://docs.pytest.org/en/6.2.x/fixture.html), and just make a dependency of the relevant tests.

For an example, you can look at `banana` [`tests/conftest.py`](https://github.com/N3PDF/banana/blob/5524a3d721bafba11093f330dc920ddf88339b56/tests/conftest.py#L11-L35) and related [usage](https://github.com/N3PDF/banana/blob/5524a3d721bafba11093f330dc920ddf88339b56/tests/navigator/test_init.py#L10-L14).

AleCandido:
Ops, sorry it was not a test.

The solution for that is to pass the command from the CLI, and for the sake of the repo itself you can put it in a `poe` script, such that you don't have to repeat over and over the same path.

AleCandido:
However, I'm doing it.

AleCandido:
In [59795b2](https://github.com/N3PDF/eko/pull/91/commits/59795b2e78268b1a64c0225c10bfce12f23038a0) I introduce the current solution for `eko`'s `navigator`. You get it as usual just running `poe nav` or `poe navigator` (or copying what is written there, inside `pyproject.toml`).

I don't like so much this solution, because I'm actually loading the configurations myself in `ekomark`, while I would have liked to do it in `banana`.
But in order to move it on the other side I need to provide a hook to be called by `banana` on initialization, and this is already quite redundant with the `import` list (that is already more or less a hook).

So, it is working and not extremely hacky [1], and since this is all we want from `navigator`, and it's obtained without using config registration in `banana`, I'm considering dropping the latter, since it does not take part in the working solution (instead of reshuffling everything attempting to find another solution).

---

[1]: a bit it is hacky:
https://github.com/N3PDF/eko/blob/59795b2e78268b1a64c0225c10bfce12f23038a0/src/ekomark/navigator/glob.py#L2
so when I do:
```py
from . import glob

glob.glob["ciao"] = "come"
```
then `glob.ciao` becomes `"come"`. But that's it, and we've been doing something like this all the time:
https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/ekomark/navigator/__init__.py#L43

src/eko/scale_variations/exponentiated.py
```diff
@@ -0,0 +1,52 @@
...
+            perturbation order
+        nf : int
+            number of active flavors
+        L : float
+            logarithmic ratio of factorization and renormalization scale
+
+    Returns
+    -------
+        gamma : numpy.ndarray
+            adjusted singlet anomalous dimensions
```
AleCandido, line 29: Why here only the singlet is varied, and in the `expanded` both are?

Moreover, if it's only singlet, maybe the name of the function should reflect it.

doc/source/theory/pQCD.rst
```diff
@@ -80,23 +80,83 @@ The expression of the pure |QED| and of the mixed |QED| :math:`\otimes` |QCD| sp
...
+      leading to:
+
+        .. math ::
+            \tilde{\mathbf{K}}(a_s) \approx & 1 - a_s k \gamma^{(0)} + a_s^2 \left [ - k \gamma^{(1)} + \frac{1}{2} k^2 \gamma^{(0)} (\beta_0 + \gamma^{(0)}) \right ] \\
+            & + a_s^3 \left [ -k \gamma^{(2)} + \frac{1}{2} k^2 \left(\beta_1 \gamma^{(0)} + 2 \gamma^{(1)} (\beta_0 + \gamma^{(0)} ) \right) - \frac{1}{6} k^3 \gamma^{(0)} \left(2 \beta_0^2 + 3 \beta_0 \gamma^{(0)}+\gamma^{(0),2} \right) \right] + \mathcal{O}(a^4)
+
+
+      In this way the dependence of the |EKO| on :math:`k` is factorized outside the unvaried evolution kernel.
+      This procedure is repeated for each different flavor patch present in the evolution path.
+      It corresponds to scheme B of :cite:`AbdulKhalek:2019ihb`, and we recommend to use it along with
```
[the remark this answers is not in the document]
giacomomagni: Yes, that's true but I thought was nice to give an explicit source for our implementation, someone in the future might wonder where does it come from...

src/eko/evolution_operator/__init__.py
```diff
@@ -216,7 +240,7 @@ class Operator:
             cut to the upper limit in the mellin inversion
     """
 
-    def __init__(self, config, managers, nf, q2_from, q2_to, mellin_cut=5e-2):
+    def __init__(self, config, managers, nf, q2_from, q2_to=None, mellin_cut=5e-2):
```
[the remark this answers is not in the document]
giacomomagni: No at the moment it's not, it will if we will use it for the matching. But it's remnant of the old implementation, will drop it!

src/eko/scale_variations/exponentiated.py
giacomomagni: No it's a typo, it should be: `adjusted anomalous dimensions`. Thanks for spotting.

doc/source/theory/pQCD.rst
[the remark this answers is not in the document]
felixhekhorn: for sure we need a reference - maybe we should just point out (or point to the paper) the difference between the two things

src/eko/scale_variations/exponentiated.py
AleCandido: Ok, perfect :)

src/eko/evolution_operator/__init__.py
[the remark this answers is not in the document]
AleCandido: Ok

doc/source/theory/pQCD.rst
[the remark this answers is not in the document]
giacomomagni: it's cited also at the beginning. I can point out the equation of the expansion instead the methods names

AleCandido (approved):

felixhekhorn:
> To my understanding the matching conditions are function of `a_s(nf)` not `a_s(nf+1)`.

What was the outcome from the PC? (Sorry I missed that part)

To add more references: look at Eq. 2.37 and following of [Buza](https://inspirehep.net/literature/427603)
also they say in the paragraph right before

> Another comment we want to make is that as appear-
ing in the heavy-quark coefficient functions is renormal-
ized at nf flavours. This means that the renormalization
of the coupling constant is carried out in such a way that
all quarks equal to H and heavier than H decouple in the
quark loops contributing to as. In the right-hand-side of
(2.31)-(2.35) the decoupling of the heavy quark H has
been undone so that here as, depends on nf + 1 flavours.
This decoupling gives rise to additional logarithms of the
type In^i(µ2/m2), which are cancelled by the OME's of the
type A_{kl,H}. However explicit indication of this procedure
further complicates our mass factorization relations above,
which we want to avoid, so we have to bear in mind that
this decoupfing is implicitly understood.

giacomomagni:
> pegasus manual

Okay, so I have to pass `nf+1`. I'll fix it now.

giacomomagni:
in case of L=2/1 on develop we had  nfref=3, here nfrer=4. I thought we were not consistent, look at the ThresholdAtlas output.

felixhekhorn:
> in case of L=2/1 on develop we had nfref=3, here nfrer=4. I thought we were not consistent, look at the ThresholdAtlas output.

okay, but apparently that's what we needed to get the match - instead are you suggesting, that in order to get the match with apfel we need this adjustment? then maybe we should only do it there

AleCandido:
If this is enough to justify the discrepancy, then change `nfref` only in `apfel` runs, as @felixhekhorn suggested.

If we get general agreement, we have solved all our problems: `eko` is so good that can match different settings, and shows inconsistency between them.

giacomomagni:
> okay, but apparently that's what we needed to get the match - instead are you suggesting, that in order to get the match with apfel we need this adjustment? then maybe we should only do it there

I see the point, sorry to insist... but I didn't understand the meaning of this syntax: 
```
ThresholdsAtlas [0.00e+00 - 2.00e+00 - 2.02e+01 - 3.06e+04 - inf], ref=2.0000000000000004 @ 3
ThresholdsAtlas [0.00e+00 - 1.00e+00 - 1.01e+01 - 1.53e+04 - inf], ref=2.0000000000000004 @ 3
Strong Coupling: a_s(µ_R^2=2.000000)^(nf=3)=0.027852=0.350000/(4π)
Prepare threshold operator
Evolution: skipping unity operator at 2.000000e+00
-> a_s = self.sc.a_s(q2 / self.config["fact_to_ren"], q2, nf + 1)
(Pdb++) self.sc.thresholds.path(q2 / self.config["fact_to_ren"], nf + 1)
[PathSegment(2.0000000000000004 -> 1.0, nf=3), PathSegment(1.0 -> 1.0, nf=4)]
```
So are we saying that `nfref` should always be specified according to factorisation scales? 
As you can see this implies having a path which backward in `Q2` but forward in `nf`.
It really seems puzzling to me...

AleCandido:
Wait, the business about `nfref` it's complicate, I know, but `nfref` is the number of flavors active at the scale you define the `alpha_s`, so by definition is renormalization scale.

Then, `nfref` has nothing to do with the definition of the `ThresholdAtlas`:
- you define your thresholds in fact scale
- you apply scale variations (`fact_to_ren`) and you get your thresholds in renormalization scale (this is the `StrongCoupling`'s `ThresholdAtlas`)
- at this point you put one special point in ren scale, that is the reference scale
- then, if no `nfref` is specified, you determine with how many light flavors start running by the former `StrongCoupling`'s `ThresholdAtlas`

Now you can wonder: what happens if I define `nfref` different from the one determined from the suitable `ThresholdAtlas`? And the answer is the easiest you can get, preserving consistency.
`nfref` tells you how many flavor you have to run in at reference scale, then you get to the nearest thresholds by running with that number of flavor, in whatever direction (forward/backward) disregarding `ThresholdAtlas`. Once you get one value at a matching scale, you're back in the standard flow.

Examples:
1. the one you got
    ```
    ThresholdsAtlas [0.00e+00 - 1.00e+00 - 1.01e+01 - 1.53e+04 - inf], ref=2.0000000000000004 @ 3
    [PathSegment(2.0000000000000004 -> 1.0, nf=3), PathSegment(1.0 -> 1.0, nf=4)]
    ```
    meaning:
    - you got your reference at `2.0000000000000004`, with `nfref=3`, but `2.0000000000000004` according to the atlas happens to be in the `nf=4` patch
    - then, you evolve with `nf=3` to the closest threshold (the charm matching scale) `PathSegment(2.0000000000000004 -> 1.0`
    - you match `PathSegment(1.0 -> 1.0, nf=4)`
    - and then you're back in the standard flow
2. another made up
    ```
    ThresholdsAtlas [0.00e+00 - 4.00e+00 - 5.00e+00 - 6.00e+00 - inf], ref=7.00 @ 3
    [PathSegment(7.0 -> 4.0, nf=3), PathSegment(4.0 -> 5.0, nf=4), PathSegment(5.0 -> 6.0, nf=5), PathSegment(6.0 -> 8.0, nf=6)]
    ```
    meaning:
    - you got reference at `7.00`, with `nfref=3`, and you want to evolve up to `8.00`
    - your starting point in the flavor path is always determined (if specified) by `nfref`, and the final point by your target scale and the `StrongCoupling`'s `ThresholdAtlas`, so for `8.00` the final `nf=6`
    - but then, to step from `nf=3` to `nf=6` it is clear what to do: match subsequently charm, bottom, and top, so: first thing, you go back to the closest matching (closest in `nf`), meaning charm one, since you started in `nfref=3`, and you have to run in `nf=3`, because that is what has been specified by the user
    - then you're back in the normal flow, and you cross all the other thresholds going back where you started from

In principle, `nfref` is born to solve complicate matching specification: if reference scale is at matching, on which side of the matching it is?
But even if that was the original question, we made a fully consistent choice, that is working for whatever atlas, whatever reference scale, and whatever `nfref` chosen.

You might wonder what's the meaning, especially of the second example, and the meaning it's the following:
- consider that someone determined the full running of `a_s` in 3 flavors, up to Z mass, because they hard-coded 3 and do nothing else
- then, by doing the first weird step of tunneling backward in 3 flavors, you undo their evolution, determining a border condition in 3 flavors, that exactly coincides to the 3 flavors value of `a_s` at that scale for those who generated the weird value in the first place
- then, once you've your sensible border condition, you can apply the prescription given by the `ThresholdAtlas`

It is weird, and I would not use this way, unless it's strictly needed. And, as I said, `nfref` was born for another reason.
But this is consistent.

AleCandido:
Now that I took the time to write down the full explanation, please:
- [ ] describe in the documentation the `nfref` role (since I was unable to find it)
  - I would put in a subsubpage of theory, linked to https://eko.readthedocs.io/en/latest/theory/pQCD.html#strong-coupling
- [ ] describe the initial role (i.e. disambiguate value at threshold) briefly in https://eko.readthedocs.io/en/latest/theory/pQCD.html#strong-coupling
- [ ] describe deeper the inital role in the subsubpage with a relevant example
- [ ] describe the consistency in the weird scenario
  - put the examples above
  - describe the scenario about someone running in 3 flavor up to Z mass

It should be mainly a matter of copy-pasting (and a bit of re-wording).

giacomomagni:
Hi @AleCandido, I see thanks for your detailed explanation, will add it to the docs for sure.
Then do we have a problem when using this [property](https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/eko/thresholds.py#L32), to determine if a path is backward ? [see here](https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/eko/evolution_operator/grid.py#L168)
For instance:
```
ThresholdsAtlas [0.00e+00 - 4.00e+00 - 5.00e+00 - 6.00e+00 - inf], ref=7.00 @ 3
[PathSegment(7.0 -> 4.0, nf=3), PathSegment(4.0 -> 5.0, nf=4), PathSegment(5.0 -> 6.0, nf=5), PathSegment(6.0 -> 8.0, nf=6)]
```
this will be considered a `backward_path` in the first step, but this is not true according to the `nf` ordering, which is used for instance [here](https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/eko/strong_coupling.py#L342)

AleCandido:
> ```
> 
> ```
> 
> 
>     
>   
> 
> this will be considered a `backward_path` in the first step, but this is not true according to the `nf` ordering, which is used for instance [here](https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/eko/strong_coupling.py#L342)

I'm reading and trying to understand what's the problem and where it is coming from. But before solving a non-problem, consider that I made up that output myself (I wrote it by hand), it's not coming from `eko`.

AleCandido:
Actual output would be:

```
[...]
ThresholdsAtlas [0.00e+00 - 4.00e+00 - 5.00e+00 - 6.00e+00 - inf], ref=2.7224999999999997 @ None
ThresholdsAtlas [0.00e+00 - 4.00e+00 - 5.00e+00 - 6.00e+00 - inf], ref=7.000000000000001 @ 3
[...]
[PathSegment(7.000000000000001 -> 4.0, nf=3), PathSegment(4.0 -> 4.0, nf=4)]
[...]
[PathSegment(7.000000000000001 -> 4.0, nf=3), PathSegment(4.0 -> 5.000000000000001, nf=4), PathSegment(5.000000000000001 -> 5.000000000000001, nf=5)]
[...]
[PathSegment(7.000000000000001 -> 4.0, nf=3), PathSegment(4.0 -> 5.000000000000001, nf=4), PathSegment(5.000000000000001 -> 5.999999999999999, nf=5), PathSegment(5.999999999999999 -> 5.999999999999999, nf=6)]
[...]
```
with:
```py

  325         a_s = self.sc.a_s(q2 / self.config["fact_to_ren"], q2, nf + 1)
+   1         print(self.sc.thresholds.path(q2 / self.config["fact_to_ren"], nf + 1))
```
in `src/eko/matching_conditions/operator_matrix_element.py`.

AleCandido:
> Then do we have a problem when using this [property](https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/eko/thresholds.py#L32), to determine if a path is backward ? [see here](https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/eko/evolution_operator/grid.py#L168)

Then maybe I understood your concern: you're worried that since the patch is backward in `Q2`, the matching is done backward, but it should be forward since you're stepping from 3 to 4.

I would say **you're right**: to determine if a matching is backward or forward, you should look at how many active flavors you come from, and how many active flavors you're stepping into. And not the just `Q2` range of the patch you come from.

Indeed, matching is part of the **flavor flow** (because it's exactly the way you step from one flavor set to another), and while in most common cases this is consistent with the **`Q2` flow**, they might be two distinct things (in involved cases, like the one above).
So yes, amend please:
https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/eko/evolution_operator/grid.py#L168
replacing `seg.is_backward` with an expression that takes into account the former and the latter segments (one segment alone is not enough to determine the *flavor flow*).

AleCandido:
Actually, since the `path` should be sorted in *flavor flow*, you should not need to determine at each iteration if it's backward or forward, but you should be able to determine at the beginning, with something like:
```py
is_backward_matching = path[0].nf > path[1].nf
```

felixhekhorn:
I agree with the discussion above and just want to add a more pragmatical point of view: we need to match the official benchmark and if we need a weird setup, we don't care - benchmarks do not necessarily be physical;

nevertheless, the determination of `is_backward` seems inconsistent and the solution in flavor space is, as said, to check on the full path - and then I wonder whether to drop `is_backward` on the single segment

AleCandido:
> nevertheless, the determination of `is_backward` seems inconsistent and the solution in flavor space is, as said, to check on the full path - and then I wonder whether to drop `is_backward` on the single segment

This might well be an option: since there is not a single concept of backward, then dropping might be more consistent. 

In any case, the flavor concept of backward, so, if we found out that that `is_backward` method is used somewhere else than matching, then I'd be practical and keep it, renaming it to `is_backward_q2`. If it's only used for matching, then it's just confusing, and I'd drop it.

giacomomagni:
I agree we can drop `is_backward` property! I'll push the fix soon

AleCandido:
@giacomomagni have a look to CodeFactor, there are a few easy issues to fix.

doc/source/code/IO-tabs/ThresholdConfig.rst
```diff
@@ -1,6 +1,10 @@
 ThresholdConfig parameters:
```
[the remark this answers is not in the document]
felixhekhorn: let's drop it

doc/source/code/IO-tabs/ThresholdConfig.rst
```diff
@@ -1,6 +1,10 @@
...
 
-- :py:obj:`float` ``Q0`` reference scale from which to start; **Required**
-- :py:obj:`float` ``mc`` charm mass in GeV; **Required**
-- :py:obj:`float` ``mb`` bottom mass in GeV; **Required**
-- :py:obj:`float` ``mt`` top mass in GeV; **Required**
+- :obj:`list(float)`: **Required**
+    - :obj:`float` ``mc`` charm  mass squared in GeV;
```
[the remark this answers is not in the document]
felixhekhorn: this ought to match the runcard - so the description would be wrong; but I suggest let's drop rather then fix

doc/source/code/IO-tabs/ThresholdConfig.rst
```diff
@@ -1,6 +1,10 @@
...
+    - :obj:`float` ``mb`` bottom mass squared in GeV;
+    - :obj:`float` ``mt`` top mass squared in GeV;
+- :obj:`float` ``q2_ref`` reference scale squared in GeV;
+- :obj:`int` ``nf_ref`` number of active flavors at the reference scale
+- :obj`list(float)`: list of ratios between masses and matching thresholds squared
```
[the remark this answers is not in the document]
felixhekhorn: Shall we move this into a separate issue/PR?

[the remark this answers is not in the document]
AleCandido: Even fine, create the issue.

pyproject.toml
```diff
@@ -76,8 +76,10 @@ ekonav = "ekomark.navigator:launch_navigator"
 [tool.poe.tasks]
 test = "pytest tests"
 bench = "pytest benchmarks"
-bench-iso = "pytest benchmarks -m isolated"
-bench-run = "pytest benchmarks -m 'not isolated'"
+bench-iso.cmd = "pytest benchmarks -m isolated"
+bench-iso.env.NUMBA_DISABLE_JIT.default = "0"
```
AleCandido, line 80: Great, you reduced by half the runtime of the workflow :D

Thank you :)

doc/source/code/IO-tabs/ThresholdConfig.rst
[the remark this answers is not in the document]
giacomomagni: Okay I'll remove it

benchmarks/apfel_bench.py
```diff
@@ -65,11 +68,20 @@ def benchmark_plain(self, pto):
             cartesian_product(th), operators.build(operators.apfel_config), ["ToyLH"]
         )
 
-    def benchmark_sv(self, pto):
+    def benchmark_sv(self, pto, scheme):
```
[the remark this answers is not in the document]
giacomomagni: okay fine

benchmarks/eko/benchmark_msbar_evolution.py
```diff
@@ -187,7 +204,7 @@ def benchmark_APFEL_msbar_solution(self):
                 apfel.EnableWelcomeMessage(1)
```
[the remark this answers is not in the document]
giacomomagni: it probably can be removed

src/eko/thresholds.py
```diff
@@ -260,3 +260,29 @@ def nf(self, q2):
...
+        is_downward: bool
+            True is number of flavor is decreasing
+        shift: 3, 4
+            shift to number of light flavors
+    """
+    if len(path) == 1:
+        is_downward = path[0].is_downward_q2
+    else:
+        is_downward = path[1].nf < path[0].nf
+    shift = 4 if is_downward else 3
```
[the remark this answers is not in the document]
giacomomagni: I see it was just convenient to return them together...

doc/source/code/IO-tabs/InterpolatorDispatcher.rst
```diff
@@ -1,5 +1,6 @@
-InterpolatorDispatcher parameter:
+InterpolatorDispatcher parameters:
 
-- :py:obj:`list(float)` ``interpolation_xgrid`` the interpolation grid; **Required**
-- :py:obj:`int` ``interpolation_polynomial_degree`` polynomial degree of the interpolating function; **Required**
-- :py:obj:`bool` ``interpolation_is_log`` use logarithmic interpolation? **Required**
+- :obj:`list(float)` ``interpolation_xgrid`` the interpolation grid; **Required**
+- :obj:`int` ``interpolation_polynomial_degree`` polynomial degree of the interpolating function; **Required**
+- :obj:`bool` ``log`` use logarithmic interpolation?
```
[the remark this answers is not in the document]
felixhekhorn: I think we should also drop this part - if someone needs it's in the API https://eko.readthedocs.io/en/latest/modules/eko/eko.html#eko.interpolation.InterpolatorDispatcher.from_dict and improved here https://github.com/N3PDF/eko/pull/91/commits/42fb1b904825a4d9726c5aa58c8acc7ec41fdb62

doc/source/code/Utilities.rst
```diff
@@ -8,10 +8,56 @@ Apart from the :doc:`operator <Operators>` classes, `eko` also provides some uti
...
+    - A path to ``Q=7`` enforcing ``nf=4``:
+
+      .. code-block::
+
+        [PathSegment(64 -> 4, nf=3), PathSegment(4 -> 49, nf=4)]
+
+  As first step, you always go back to the closest matching scale (closest in ``nf``),
+  running in ``nf=3``, since that is what has been specified in the boundary conditions.
+  Then you are back in the normal flow, and you cross all the other thresholds going according to the prescription given
+  by the :class:``eko.thresholds.ThresholdAtlas``.
```
AleCandido, line 51: Two ``` ` ```, this is syntactically wrong (and indeed does not compile the proper link).

doc/source/code/Utilities.rst
```diff
@@ -8,10 +8,55 @@ Apart from the :doc:`operator <Operators>` classes, `eko` also provides some uti
...
      the :class:`eko.strong_coupling.StrongCoupling` and the :doc:`operators <../theory/Matching>`
 
+  .. include:: IO-tabs/ThresholdConfig.rst
 
-.. include:: IO-tabs/ThresholdConfig.rst
+  When running in |VFNS| it is important to specify the number of flavors active at each given scale, since the evolution path
+  can be different depending of the chosen setting. This path is determined by :meth:`eko.thresholds.ThresholdsAtlas.path`.
 
+  Let us consider an example to better illustrate how it works.
+  Imagine to have a boundary condition ``q2_ref=64, nf_ref=3`` and would like to evolve your object (|PDF| or :math:`a_s`) to a lower
```
felixhekhorn, line 17: seems like 64 is not your reference value - the code says Q2=49 - but your target value

doc/source/code/Utilities.rst
```diff
@@ -8,10 +8,55 @@ Apart from the :doc:`operator <Operators>` classes, `eko` also provides some uti
   -  Implementation of the flavor number scheme and the quark thresholds both for
      the :class:`eko.strong_coupling.StrongCoupling` and the :doc:`operators <../theory/Matching>`
 
+  .. include:: IO-tabs/ThresholdConfig.rst
 
-.. include:: IO-tabs/ThresholdConfig.rst
+  When running in |VFNS| it is important to specify the number of flavors active at each given scale, since the evolution path
+  can be different depending of the chosen setting. This path is determined by :meth:`eko.thresholds.ThresholdsAtlas.path`.
 
+  Let us consider an example to better illustrate how it works.
```
felixhekhorn, line 16: while an example is fine (and maybe even needed), let's not start with the most complicated and exotic one, but with a regular

doc/source/code/Utilities.rst
```diff
@@ -8,10 +8,56 @@ Apart from the :doc:`operator <Operators>` classes, `eko` also provides some uti
...
+      .. code-block::
+
+        [PathSegment(64 -> 4, nf=3), PathSegment(4 -> 49, nf=4)]
+
+  As first step, you always go back to the closest matching scale (closest in ``nf``),
+  running in ``nf=3``, since that is what has been specified in the boundary conditions.
+  Then you are back in the normal flow, and you cross all the other thresholds going according to the prescription given
+  by the :class:``eko.thresholds.ThresholdAtlas``.
+  You can noticed that :meth:`eko.thresholds.ThresholdsAtlas.path` are always ordered according to `nf` and not `Q2` scales.
+  This property is used in |VFNS| to determine if a path is backward or not modifying the way the different patches are matched.
```
AleCandido, line 53: Even the *backward-icity* of a path is unclear, mention the matching instead (i.e. something like "if the matchings are done for an increasing or decreasing number of light flavors").

doc/source/code/Utilities.rst
```diff
@@ -8,10 +8,55 @@ Apart from the :doc:`operator <Operators>` classes, `eko` also provides some uti
...
 
+  .. include:: IO-tabs/ThresholdConfig.rst
 
-.. include:: IO-tabs/ThresholdConfig.rst
+  When running in |VFNS| it is important to specify the number of flavors active at each given scale, since the evolution path
+  can be different depending of the chosen setting. This path is determined by :meth:`eko.thresholds.ThresholdsAtlas.path`.
 
+  Let us consider an example to better illustrate how it works.
+  Imagine to have a boundary condition ``q2_ref=64, nf_ref=3`` and would like to evolve your object (|PDF| or :math:`a_s`) to a lower
+  scale (say ``Q=7``) but giving different number of active flavors to the final state and having set the heavy quarks mass thresholds
```
felixhekhorn, line 18: let's stick to our default choice, that is we speak in Q2 and *not* Q (you can do `Q2=7**2` ...) - and actually why don't you go for `math:` - we speak about math here ...

AleCandido (changes requested):

AleCandido:
I completed my review, there are only minor fixes requested, mainly in the docs.

doc/source/theory/pQCD.rst
```diff
@@ -46,6 +46,11 @@ have to be applied :cite:`Schroder:2005hy,Chetyrkin:2005ia`.
 In particular, the matching involved in the change from :math:`n_f` to :math:`n_f-1` schemes
 is presented in equation 3.1 of :cite:`Schroder:2005hy` for |MSbar| masses, while the
 same expression for POLE masses is reported in Appendix A.
+For this reason the boundary conditions of :class:`eko.strong_coupling.StrongCoupling`
+can be specified at `scale_ref` along with ``nf_ref`` and, the computed result can
```
felixhekhorn, line 50: when you speak of code put double backticks (for scale_ref)

src/eko/thresholds.py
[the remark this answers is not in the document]
AleCandido: Not even always, since sometimes you trash the second one:

https://github.com/N3PDF/eko/blob/a0b5017aacdf7ba0b23f420a1e50e25fe8b30549/src/eko/evolution_operator/grid.py#L246

So just return the first, the second one you obtain by a trivial function applied on the first returned value:
```py
def flavor_shift(is_downward):
    return 3 if not is_downward else 4
```

doc/source/theory/pQCD.rst
```diff
@@ -80,23 +85,83 @@ The expression of the pure |QED| and of the mixed |QED| :math:`\otimes` |QCD| sp
...
 
-.. math ::
-    \gamma^{(1)}(N) &\to \gamma^{(1)}(N) - \beta_0 \ln(\mu_F^2/\mu_R^2) \gamma^{(0)} \\
-    \gamma^{(2)}(N) &\to \gamma^{(2)}(N) - 2 \beta_0 \ln(\mu_F^2/\mu_R^2) \gamma^{(1)} - ( \beta_1 \ln(\mu_F^2/\mu_R^2) - \beta_0^2 \ln^2(\mu_F^2/\mu_R^2) )  \gamma^{(0)}
+    * it naturally incorporates renormalization group invariance,
+      as the perturbative order increases, estimates of |MHOU| decrease;
+    * the same procedure can be used for any perturbative process,
+      since the scale dependence of the strong coupling :math:`a_s(\mu^2)` and of PDFs is universal;
+
+However there is no unique principle to determine the specific range of the scale variation;
```
felixhekhorn, line 105: - "However," (comma)
- I'd suggest: "principle" -> "prescription" ...
- also maybe avoid the semi-colon (if you have an easy fix)

doc/source/code/Utilities.rst
AleCandido: Actually I guess the error is not in here, but in the code, because otherwise it would be inconsistent with the paths and the rest of the discussion.

doc/source/theory/pQCD.rst
```diff
@@ -80,23 +85,83 @@ The expression of the pure |QED| and of the mixed |QED| :math:`\otimes` |QCD| sp
...
+      ``ModEv='truncated'`` in order to keep consistency with the evolution integral expansion.
+
+
+By construction, the corrections of the order :math:`\mathcal{O}(k^n)` will appear
+at the order :math:`n` in the expansion :math:`a_s`.
+This happens because :math:`\beta \approx \mathcal{O}(a_s^2)`, :math:`\gamma \approx \mathcal{O}(a_s)`
+and the contribution proportional to :math:`\mathcal{O}(k^n)` is originated
+by the `n-th` derivative in :math:`\gamma` :cite:`AbdulKhalek:2019ihb`.
 
+Furthermore the distance between the varied |EKO| and the unvaried one will decrease increasing
```
felixhekhorn, line 152: Can you find an easy rewording such that the sentence becomes more clean? (I find "decrease increasing" confusing :upside_down_face: )

benchmarks/eko/benchmark_strong_coupling.py
```diff
@@ -787,3 +787,79 @@ def benchmark_lhapdf_zmvfns_lo(self):
...
+                apfel.SetRenFacRatio(1.0 / theory_dict["fact_to_ren_scale_ratio"])
+                apfel.InitializeAPFEL()
+                # collect a_s
+                apfel_val = apfel.AlphaQCD(
+                    np.sqrt(Q2) / theory_dict["fact_to_ren_scale_ratio"]
+                ) / (4.0 * np.pi)
+            else:
+                apfel_val = 0.03478112968976964
+            # check myself to APFEL
+            np.testing.assert_allclose(apfel_val, my_val, rtol=0.03)
```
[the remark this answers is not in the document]
giacomomagni: It's due to the matching I believe, as you see, `apfel_val` matches `my_val_4`

benchmarks/eko/benchmark_msbar_evolution.py
```diff
@@ -187,7 +204,7 @@ def benchmark_APFEL_msbar_solution(self):
...
             # check myself to APFEL
             np.testing.assert_allclose(
-                apfel_vals, np.sqrt(np.array(my_masses)), rtol=5e-5
+                apfel_vals, np.sqrt(np.array(my_masses)), rtol=4e-4
```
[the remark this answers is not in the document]
giacomomagni: I honestly don't know

benchmarks/eko/benchmark_strong_coupling.py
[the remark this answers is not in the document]
giacomomagni: > How about other Q2s? Also please put the usual layout: define the cached value first and in `if use_apfel` assert that the cached and the actual value match

This test was just to check `a_s` at the thr scale, okay for the style.

doc/source/code/Utilities.rst
giacomomagni: yes it was wrong copy and paste

doc/source/code/Utilities.rst
giacomomagni: I thought it was more clear to follow the same syntax we use in the code since we are referring directly to our implementation. I have fixed the squared, which were not consistent.

felixhekhorn (approved):

PR #89, opened 2022-02-02
QED - flavor basis

niclaurenti:
This will hold the QED implementation on flavor basis

PR #95, opened 2022-02-17
Prepare extension of DGLAP to QED

felixhekhorn:


felixhekhorn:
@niclaurenti please push our changes! we forgot ...

AleCandido:
I discussed with @felixhekhorn and I believe we need a different and less random naming scheme for our new basis, and then even for the anomalous dimension.

#### Basis of flavor space

I propose something like the following:

- gluon is **g** (`g` in code), and I guess this is just fine
- **Sigma** and **V** are untouched, that's fine (as usual in code they are `S` and `V`)
- photon might be *\gamma* or *photon* (and respectively `ga` or `ph` in code)
  - I personally prefer **\gamma** and `ph`, but it's incosistent and we can discuss
- *\Delta Sigma* and *\Delta V* of Stefano are nice names for me, but if we want to start with the correct letter, I'd say let's use **Sigma_\Delta** and **V_\Delta** (in code they might be `Sdel` and `Vdel`, or even `Sdelta` if preferred)
- about non-singlet I'd propose to use meaningful names: **V_{u,3}**, **T_{u,3}**, **V_{u,8}**, **T_{d,8}**,  **V_{d,3}**, **T_{d,3}**, **V_{d,8}**, **T_{d,8}**, (in code `Vu3`, `Td8`, ...)
  - the idea of `Tu8` is that is doing the same of `T8`, but only with up-like quarks (has the same coefficients for the first, second, and third elements of the family)

#### Basis for operators

I'd like to keep the `anomalous_dimension_basis`, even if we can keep only the idea, and has to be extended to QEDxQCD.
In general, we have many more coupled elements, so to me it makes sense to name them consistently with a 2-tuple.

In order to have few complicated objects at bottom level, we can even replace strings with our beloved evolution pseudo-PIDs and similar:
https://github.com/N3PDF/eko/blob/e98971aa2e6457cfd4bb834cf6192a79f5a4aa07/src/eko/basis_rotation.py#L62-L67

I propose to use 11xx, 12xx, 13xx, 14xx for the new families (and leaving a gap on 0xxx, you never know).

The only problem is how to name sectors like `nsp`, `nsm`, and `nsv`. I would say it's not clever to name out of a single random representative, so I would assign their own names.
Since we are looking for numbers, we can assign something in the 1xxxx range, keeping it well separated from the linear basis, and still fitting in two bytes (<6.5e4), if ever needed.
Names to be assigned to the diagonal elements in the decoupled sector are simple: `nspu`, `nspd`, `nsmu`, ... .

#### Specs

All these mappings between names/objects and numbers have to be specified both in the docs, and in the code (namely in `eko.basis_rotation`, that we might consider to rename `eko.bases` or `eko.flavor` at some point), in order to allow for automatic conversions later.

niclaurenti:
@giacomomagni @AleCandido We are proposing to change the identification of the matrix of the element in the DGLAP singlet matrix by a tuple of strings: for example `("S", "g")` to select the singlet from gluon contributions. 
We believe it to be more readable than having an arbitrary number assigned. 
This is a conflict with pull request #91 .
What should we do?
We also ask if we should split the pull request in two parts: the preparation here and the actual extension to QED.

src/eko/matching_conditions/__init__.py
```diff
@@ -80,11 +80,11 @@ def split_ad_to_evol_map(
...
-                            # f"S.{ihq}+": ome_members["S_qH"],
-                            f"g.{ihq}+": ome_members["S_gH"],
-                            f"{ihq}-.{ihq}-": ome_members["NS_HH"],
-                            # f"V.{ihq}-": ome_members["NS_qH"],
+                            f"{ihq}+.{ihq}+": ome_members[("H+", "H+")],
+                            # f"S.{ihq}+": ome_members[("S", "H+")],
+                            f"g.{ihq}+": ome_members[("g", "H+")],
+                            f"{ihq}-.{ihq}-": ome_members[("H-", "H-")],
+                            # f"V.{ihq}-": ome_members[("V", "H-")],
                         }
```
AleCandido, line 88: Since we're actually changing, I'd use a lowercase `h` for the heavy quark, because in general is a placeholder for `c`, `b`, or `t`, so it's more consistent,

giacomomagni:
> @giacomomagni @AleCandido We are proposing to change the identification of the matrix of the element in the DGLAP singlet matrix by a tuple of strings: for example `("S", "g")` to select the singlet from gluon contributions. We believe it to be more readable than having an arbitrary number assigned. This is a conflict with pull request #91 . What should we do? We also ask if we should split the pull request in two parts: the preparation here and the actual extension to QED.

Hi @niclaurenti , I agree that the best is to spit this PR and implement this update separately to the QED ad (so will be faster). This method is more general than what I did in #91, so I can drop that one and merge the updated `develop` there. 
Can we pass integers (`pids`) to `numba` instead of `string`? This will optimise our usage of `numba`.

AleCandido:
About `numba` optimization: can you provide a snippet showing that the thing with `str` is slower that the same one with `u1`? (and maybe even `u2`) @giacomomagni 

Furthermore, as foreseen, it is not a great idea to merge everything in `feature/qed`, because for example this addition @giacomomagni will like to merge into its branch, that of course is based on `develop`, and not `feature/qed`.
Both the changes in this PR and in #89 are non-breaking, so let's merge this in `feature/qed`, and immediately `feature/qed` in `develop`.

Even the introduction of anomalous dimensions will be non-breaking, so also the next PR will be merged into `develop`.
The only problem will come when we'll try to add an actual solution for QED, and so that PR will have to be self-contained.

giacomomagni:
> About `numba` optimization: can you provide a snippet showing that the thing with `str` is slower that the same one with `u1`? (and maybe even `u2`) @giacomomagni
>
I read it [here](https://stackoverflow.com/questions/46708708/compare-strings-in-numba-compiled-function)

AleCandido:
> I read it [here](https://stackoverflow.com/questions/46708708/compare-strings-in-numba-compiled-function)

Thank you for the reference, it contains the snippet I was looking for. Since it looks like that `str` comparison is one order of magnitude slower than `int` comparison, it might be worth to replace `str` with `u1`, keeping the translation in `basis_rotation` (so it's easy to interpret the moment we need).

Most likely it's not the most expensive operation we're doing, but either we profile and we confirm that it is the case, or if we're not going to investigate, better to replace in case of doubts.

felixhekhorn:
> Most likely it's not the most expensive operation we're doing, but either we profile and we confirm that it is the case, or if we're not going to investigate, better to replace in case of doubts.

yes, let's replace with integers ...

AleCandido:
Since #89 has been merged, I rebased on `develop`

src/eko/anomalous_dimensions/__init__.py
```diff
@@ -116,6 +116,7 @@ def gamma_ns(order, mode, n, nf):
         elif mode in [10201, 10200]:
             gamma_ns_1 = nlo.gamma_nsm_1(n, nf)
         else:
+            print(mode)
```
AleCandido, line 119: Don't make a separate print, if you believe this information to be relevant, put it in the error message.

Use an [f-string](https://docs.python.org/3/tutorial/inputoutput.html#tut-f-strings).

AleCandido (changes requested):

src/eko/anomalous_dimensions/__init__.py
niclaurenti: Numba doesn't allow for fstrings

src/eko/constants.py
```diff
@@ -2,15 +2,6 @@
...
 
-
-The constants are:
-
-    - :math:`N_C` the number of colors - defaults to :math:`3`
-    - :math:`T_R` the normalization of fundamental generators - defaults to :math:`1/2`
-    - :math:`C_A` second Casimir constant in the adjoint representation - defaults to
-      :math:`N_C = 3`
-    - :math:`C_F` second Casimir constant in the fundamental representation - defaults to
-      :math:`\frac{N_C^2-1}{2N_C} = 4/3`
```
AleCandido, line 13: I agree to remove duplication, however formulas could be preserved in the documentation of the individual variables :)

AleCandido:
Before reviewing, please provide full test coverage (strange that I'm complaining about this, usually is @felixhekhorn's role...).

AleCandido:
It's just one error:
https://github.com/N3PDF/eko/blob/de4d4eb1630357d617fe80d4283261527585cda7/src/eko/anomalous_dimensions/__init__.py#L119
plus one simple file:
https://github.com/N3PDF/eko/blob/de4d4eb1630357d617fe80d4283261527585cda7/src/eko/anomalous_dimensions/aem1.py

felixhekhorn:
> Before reviewing, please provide full test coverage (strange that I'm complaining about this, usually is @felixhekhorn's role...).

it's because we simply haven't had a look yet ... (as you saw, we just fixed test/bench_*)

AleCandido:
The moment you fix the test, remember to remove the "Draft" attribute from the PR, by clicking "Ready for review" down below.

src/eko/evolution_operator/__init__.py
```diff
@@ -242,16 +243,16 @@ def labels(self):
             logger.warning("Evolution: skipping non-singlet sector")
         else:
             # add + as default
-            labels.append("NS_p")
+            labels.append((br.non_singlet_pids_map["ns+"], 0))
```
giacomomagni, line 246: All these tuples `(br.non_singlet_pids_map["ns+"], 0)` are already in `non_singlet_labels `. Can't you use the second one directly?

src/eko/anomalous_dimensions/aem1.py
```diff
@@ -0,0 +1,88 @@
+# -*- coding: utf-8 -*-
+import numba as nb
+
+from .. import constants
+from . import as1
```
giacomomagni, line 5: Can you make all the anomalous dimension files consistent with this new convention? 
I would say `qcd_lo` and similar is more clear.

felixhekhorn: most likely we'll go back to `lo` - without prefix - since this also becomes ambiguous at `nlo` as there are three possibilities ... instead, maybe we go for `lo1, lo2`, `nlo1`, `nlo2`, `nlo3` as is usually done in the EW literature ...

AleCandido: No wait, let's stick to powers, they are much more straightforward than `lo1234` and so on.

If it is ambiguous to have a default to apply when the power it's not specified, let's always specify the power:
- `as1a0`
- `as0a1`
- `as2a0`
- `as1a1`
- ...

felixhekhorn: well that's what we agreed on in the meeting on Monday (and that's why we asked in the first place)

AleCandido: Why don't you like the fully explicit thing?

`nlo1`, `nlo2`, `nlo3` is certainly far more confusing...

giacomomagni: yes I agree, let's specify the powers, but in all files!

felixhekhorn: as said on Monday... we give up on `gamma_ns_0` something we don't like

src/eko/evolution_operator/__init__.py
niclaurenti: Yes of course. Anyway I think that the first one is a bit more readable than the second.

giacomomagni: you are right, I just found it quite long, since it appears many times... in the end to singlet you're using `singlet_labels` straight

src/eko/anomalous_dimensions/aem1.py
AleCandido: At least for the modules, let's stick to the powers. Concerning the name of the functions, we can discuss, but I guess there is no need for suffixes, since they are scoped in modules.

giacomomagni (approved):

src/eko/anomalous_dimensions/__init__.py
AleCandido: If `numba` does not support f-strings, try something else: there are several ways to manipulate a string, e.g. concatenation with `+`, and this [is supported by `numba`](https://numba.pydata.org/numba-doc/dev/reference/pysupported.html#str) ;)

In general, I highlighted a problem and proposed a solution. But if the solution I proposed is not available, that's not a good reason to dismiss the problem ^^

niclaurenti:
We want to close this pull request to have this in develop and continue our implementation

src/eko/anomalous_dimensions/__init__.py
```diff
@@ -52,9 +53,9 @@ def exp_singlet(gamma_S):
 
     See Also
     --------
-        eko.anomalous_dimensions.lo.gamma_singlet_0 : :math:`\gamma_{S}^{(0)}(N)`
-        eko.anomalous_dimensions.nlo.gamma_singlet_1 : :math:`\gamma_{S}^{(1)}(N)`
-        eko.anomalous_dimensions.nnlo.gamma_singlet_2 : :math:`\gamma_{S}^{(2)}(N)`
+        eko.anomalous_dimensions.as1.gamma_singlet_0 : :math:`\gamma_{S}^{(0)}(N)`
```
AleCandido, line 56: So, the final decision has been to don't write the power explicit when it is 0, is it correct?
If yes, for me it's also fine.

src/eko/anomalous_dimensions/__init__.py
```diff
@@ -80,7 +81,7 @@ def gamma_ns(order, mode, n, nf):
     ----------
         order : int
             perturbative order
-        mode : "m" | "p" | "v"
+        mode : "-" | "+" | "V"
```
AleCandido, line 84: Since it's written above that mode has to be a number (specifically `u2`), then it can not be `"-" | "+" | "V"`.

src/eko/anomalous_dimensions/aem1.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+@nb.njit("c16(c16)", cache=True)
+def gamma_phq_0(N):
```
AleCandido, line 9: Here (and related) the agreement was to drop `_0`, since there is no consistent way to assign it, and rely on the module scope.

AleCandido: Related: all the functions in this subpackage, not only the same module (i.e. the same file).

src/eko/constants.py
```diff
@@ -20,7 +11,13 @@
 """the normalization of fundamental generators"""
```
AleCandido, line 11: For consistency, write the default also here.

src/eko/constants.py
```diff
@@ -20,7 +11,13 @@
...
 
 CA = float(NC)
-"""second Casimir constant in the adjoint representation"""
+"""second Casimir constant in the adjoint representation - defaults to :math:`N_C = 3"""
 
 CF = float((NC * NC - 1.0) / (2.0 * NC))
-"""second Casimir constant in the fundamental representation"""
+"""second Casimir constant in the fundamental representation - defaults to :math:`\frac{N_C^2-1}{2N_C} = 4/3"""
```
AleCandido, line 17: Since there has always been written that these numbers are only the default, let's add a small function to update them consistently:

```py
def update_colors(nc):
    global NC, CA, CF
    NC = int(nc)
    CA = float(NC)
    CF = float((NC * NC - 1.0) / (2.0 * NC))
```

tests/eko/test_ev_operator.py
```diff
@@ -30,7 +30,8 @@ def test_quad_ker(monkeypatch):
         res_ns = quad_ker(
             u=0,
             order=0,
-            mode="NS_p",
+            mode0=br.non_singlet_pids_map["ns+"],
```
AleCandido, line 33: Since all occurrences of these `br.non_singlet_pids_map[...]` are quite verbose for a number, I wonder if we should provide even further aliases in `br`, like `br.nsp` and related.

AleCandido (changes requested):

src/eko/anomalous_dimensions/__init__.py
felixhekhorn: the solution was https://github.com/N3PDF/eko/pull/95/commits/9b8c1310eab032542e6c42b11f0b7b64cfacdcb8

src/eko/anomalous_dimensions/__init__.py
niclaurenti: We were wondering whether to drop also the 1 and 2 from as2 and as3, since once they are scoped there's no need to specify them. What should we do?

AleCandido: According to me, drop them all. @felixhekhorn?

giacomomagni: I would drop them

src/eko/constants.py
felixhekhorn: I wonder if this function is ever user ... however, if you add the function, you also have to provide a unit test

AleCandido: > I wonder if this function is ever user ...

It's just to allow to change consistently, since we provide them in full generality, it's silly not to provide such a simple function.
However, it's not used of course, the proper question is if it *will be* used, and this you never know (e.g. in principle QCD with one color and suitable charges should be QED, so you might use it for testing).

>

felixhekhorn: @niclaurenti please also add a docstring

tests/eko/test_ev_operator.py
felixhekhorn: Maybe we can do (or not do) this later ...

src/eko/anomalous_dimensions/__init__.py
```diff
@@ -94,37 +95,39 @@ def gamma_ns(order, mode, n, nf):
...
-            gamma_ns_1 = nlo.gamma_nsp_1(n, nf)
+        if mode == 10101:
+            gamma_ns_1 = as2.gamma_nsp(n, nf)
         # To fill the full valence vector in NNLO we need to add gamma_ns^1 explicitly here
-        elif mode in ["m", "v"]:
-            gamma_ns_1 = nlo.gamma_nsm_1(n, nf)
+        elif mode in [10201, 10200]:
+            gamma_ns_1 = as2.gamma_nsm(n, nf)
+        else:
+            raise NotImplementedError("Non-singlet sector is not implemented")
```
giacomomagni, line 119:
```suggestion
            raise NotImplementedError(f"Non-singlet sector {mode} is not implemented")
```

AleCandido: It looks like f-strings are not yet supported by numba, but you can do:

```suggestion
            raise NotImplementedError("Non-singlet sector " + str(mode) + " is not implemented")
```

src/eko/evolution_operator/__init__.py
```diff
@@ -32,25 +32,27 @@
...
 
     Parameters
     ----------
-        mode : str
-            sector element
         ker : numpy.ndarray
             singlet integration kernel
-
+        mode0 : int
+            first sector element
```
giacomomagni, line 45:
```suggestion
            pid for first sector element
```
same for `mode1`, to be consistent

tests/eko/test_ad_nlo.py
```diff
@@ -73,7 +73,37 @@ def test_gamma_1():
...
     )  # gg
-    gS1 = ad_nlo.gamma_singlet_1(4, NF)
+    gS1 = ad_as2.gamma_singlet(4, NF)
     np.testing.assert_allclose(
         gS1[0, 1], (-56317.0 / 18000.0 * CF + 16387.0 / 9000.0 * CA) * NF
     )  # qg
+
+    update_colors(4)
+    from eko.constants import CA as ca
+    from eko.constants import CF as cf
```
AleCandido, line 83: Here to get the actual thing updated, the best thing is to keep the scope when imported top level:

```py
from eko import constants as const

assert const.CA == 3
update_colors(4)
assert const.CA == 4
```

If possible, it would be better to do it everywhere, but for the time being if you update here is just fine.

AleCandido (changes requested):

src/eko/constants.py
```diff
@@ -2,25 +2,37 @@
...
+def update_colors(nc):
+    """
+    Updates the number of colors to :math:`NC = nc` and the Casimirs for a generic value of :math:`NC`
+
+    Parameters
+    ----------
+      nc : int
+        Number of colors
+    """
+    global NC, CA, CF
```
giacomomagni, line 35: disable `pylint(global-statement)` since this is really what you want here...

src/eko/constants.py
```diff
@@ -2,25 +2,37 @@
...
-    - :math:`C_F` second Casimir constant in the fundamental representation - defaults to
-      :math:`\frac{N_C^2-1}{2N_C} = 4/3`
 """
 
 NC = 3
 """the number of colors"""
 
 TR = float(1.0 / 2.0)
-"""the normalization of fundamental generators"""
+"""the normalization of fundamental generators - defaults to :math:`T_F = 1/2`"""
```
giacomomagni, line 11:
```suggestion
"""the normalization of fundamental generators - defaults to :math:`T_R = 1/2`"""
```

src/eko/evolution_operator/__init__.py
AleCandido: Maybe here I'd write `id` more generically: the `pid` is usually taken to be one of those listed in PDG, but here we're making up `id`s ourselves (like for singlet), so no need to generate more confusion, let's just drop one letter.

tests/eko/test_ad_lo.py
```diff
@@ -12,39 +13,63 @@ def test_number_conservation():
...
+
+def test_gamma_qph_0():
+    N = complex(1.0, 0.0)
+    res = complex(-20.0 / 3.0, 0.0) * 3 / 0.5
+    np.testing.assert_almost_equal(ad_aem1.gamma_qph(N, NF), res)
+
+
+def test_gamma_phph_0():
+    res = complex(-4.0 / 3, 0.0)
+    np.testing.assert_almost_equal(ad_aem1.gamma_phph(), res)
```
giacomomagni, line 75: Here we can't test the photon conservation?  
I see that `gamma_qph` contains the factor NC, I was wondering if this has to be included also in `gamma_phph`
according to eq 2.4 of `Carrazza:2015dea`

src/eko/anomalous_dimensions/__init__.py
```diff
@@ -94,37 +95,41 @@ def gamma_ns(order, mode, n, nf):
...
+        if mode == 10101:
+            gamma_ns_1 = as2.gamma_nsp(n, nf)
         # To fill the full valence vector in NNLO we need to add gamma_ns^1 explicitly here
-        elif mode in ["m", "v"]:
-            gamma_ns_1 = nlo.gamma_nsm_1(n, nf)
+        elif mode in [10201, 10200]:
+            gamma_ns_1 = as2.gamma_nsm(n, nf)
+        else:
+            raise NotImplementedError(
+                "Non-singlet sector " + str(mode) + " is not implemented"
```
AleCandido, line 120: Okay, now it has been tested and it's [not working](https://github.com/N3PDF/eko/runs/5508976685), so let's revert this.

AleCandido (changes requested):

tests/eko/test_ad_lo.py
niclaurenti: I agree with you: 
in fact we had included NC in the def of gamma_qph so that we can remove it from the def of theta+- (see eq. 2.27 of Carrazza:2015dea). But the same applies gamma_phph with e_\Sigma.
In my opinion either we put both NC factors inside gamma_qph and gamma_phph or we put them in theta+- and e_\Sigma.

giacomomagni: If think that `gamma_phph` should be:
```
- 2 / 3 * NC * 2 * nf
```

this way the photon momentum conservation test will work

giacomomagni: @niclaurenti. Just for the records, don't know if this is already in the references: 
[paper](https://arxiv.org/pdf/1512.00612.pdf), but might come useful also in the future.

AleCandido (approved):

tests/eko/test_ad_lo.py
AleCandido: Thanks, @giacomomagni, this appeared at the same time as Stefano's thesis, so they might have a common reference, but it's spelling out the full "algorithm to obtain QED ad from QCD ones", that might become handy sooner or later :)

giacomomagni (approved):

PR #105, opened 2022-03-13
Composite Output

### LINKED ISSUE #96: Memory saving output
We realized that most `eko` operations do not depend on multiple `Q2` at the same time.

The full `OperatorGrid` is a rank-5 tensor, by using only one `Q2` at a time we reduce the problem to a rank-4 tensor, with much less memory consumption.

In order to get a usable and flexible structure, a few capabilities are needed for the new structure:
- [x] separate `Q2` storage: it will be `{q2}.npy`
- [x] lazy loading: loads one (or a subset of) `Q2` at a time, and drop them once consumed
  - even the subset might be useful, for cases in which the consumer only accepts a rank-5 tensor
- [ ] merge separately computed (but input compatible) outputs
- [ ] split a single output into multiple ones
  - *not strictly needed, but it's dual to the former one, and so nice to have*

This new structure will be the upgraded version of the current `Output` object.

Moreover, a few more things might be implemented as related, and later used to support separate computation of `Q2` elements:
- [ ] replace/upgrade `OperatorGrid`
  - we need a manager for the computation, but it has not to hold the data, that has to be dumped as soon as possible
  - the easiest is to have the current `OperatorGrid` to hold the reference to a new `Output` object, and store everything in there
  - everything will include threshold operators, and partial `Q2` results (those obtained before combining with threshold operators), together with full results
- [ ] support separately threshold operators and partial `Q2` in `Output`
  - they are also dumped on disk, with their own names: `thresholds.npz` (containing all the threshold elements) and `{q2}.part.npy`

The idea is that, an object supporting these features, can be computed separately, in a completely independent way (1 process for thresholds, plus one for each `Q2`, for example), and then merged together.
In order to make it easier to merge and compute the final one:
- the `thresholds.npz` is never removed from the saved output
  - if other `{q2}.part.npy` come later, they can always be consumed
  - [ ] unless explicitly stated: provide an `optimize()` or `clean()` method, to get rid of it
- everything can be merged together, with or without thresholds, but it's simply checking the input compatibility, and adding the arrays to the archive
- if it contains both thresholds and partial objects, compute final ones
  - [ ] provide a `combine()` method
  - partial objects are removed after combination

AleCandido:
This should close #96.

- [x] split the 5-dim array into 4-dim ones in the `operators` folder

In this PR I'd like to provide even a CLI in `ekobox`:

- [x] kickstart `eko` CLI

Further:

- ~add `product` subcommand (log operations into `history.yaml` and runcards to `{n+1}.theory.yaml` and `{n+1}.operator.yaml`)~ -> postponed to #107 
- [x] autoload operators on `__getitem__`
- [x] unload (replace with `None`) and dump operators on `__delitem__` (if already `None`, do nothing)
- [x] call `__delitem__` for each operator on `EKO.close`, and call this last in `EKO.__del__`, for auto unloading
  - notice that
    > It is not guaranteed that [`__del__()`](https://docs.python.org/3/reference/datamodel.html?highlight=__del__#object.__del__) methods are called for objects that still exist when the interpreter exits.

    so when the user program ends, `EKO.close` has to be called explicitly
- [x] make an iterator for `EKO` that loads operator one by one, and when loading the next one will dump the former (useful for applications: manipulation, consumption)
  - pay attention to dump even the last element before closing iteration
  - make a generator, `yield` elements one by one, `raise StopIteration` manually at the end
- [x] while manipulating, instead of skipping unloaded operators, load them one by one and manipulate (rely on the former)
- [x] context manager for item retrieval, with automated drop
- [x] provide docstrings (at the very least for the content of `struct.py`)
- [x] try to remove `PathLike` (in favor of the native `os.PathLike`)
- [x] implement deep copy (since it has to copy even the disk part, it has to request an explicit path `eko.deepcopy(path)`)
- [x] fix https://github.com/N3PDF/eko/pull/105#discussion_r924382850
- [x] fix https://github.com/N3PDF/eko/pull/105#discussion_r924381781 and https://github.com/N3PDF/eko/pull/105#discussion_r924385487
- [x] create `compatibility` layer https://github.com/N3PDF/eko/pull/105#issuecomment-1197929649


AleCandido:
In this PR I definitely found out two things:

1. `eko` is too coupled: changes in `output` module and runcards format have propagated everywhere
  - a good way to reduce coupling would be to avoid passing around runcards to the `from_dict` method of everything, and maybe just to kill those methods: not only `runner`, but potentially everything is aware of the structure of the input runcards (or part of it)
  - we can think on more ways to decouple
2. Test suite is definitely useful, even for refactoring: I changed a lot (and also tests had to change a bit), but I'm pretty sure everything is still working
  - we're not running benchmarks in the CI, so they are potentially broken
  - we're lacking tests for `ekobox` and `ekomark`, so they are potentially broken

AleCandido:
@andreab1997 please help :sos: :pray: 

I fixed all the old `ekobox` tests, but for the product.
While investigating it, I tried to fix it by simplifying (thanks to the new `EKO` class and related sugar). After the first few fixes it was just numerically failing the operator comparison, and after the heavy simplification is still numerically failing the same way...

AleCandido:
@andreab1997 yesterday evening I was too tired, it looks like...

Actually, the problem has been solved in [6af70f0](https://github.com/N3PDF/eko/pull/105/commits/6af70f0db0be67977f062bb3b76e087642aff965): the product code (most likely after my update) was recomputing an operator that was already part of the "initial" one.

AleCandido:
Since we should improve the product to guarantee that nothing weird will happen, I opened a dedicated issue #107.

AleCandido:
I left intentionally a couple of `pdb` statement, because I'm trying to bisect the current issue in "isolated benchmarks".

AleCandido:
@felixhekhorn I removed the usage of `unittest` in [e7be4bb](https://github.com/N3PDF/eko/pull/105/commits/e7be4bb5199dcbadd3083802869195b2989ed564) for two reasons:

1. it's true that `unittest` is standard library, nevertheless it is a further testing framework; since we're already using `pytest`, let's be as much consistent as possible and always use only that one (the thing it was tested by `unittest`, the mode in which the file was opened, it was unnecessarily detailed... if you wish we can improve current status by testing file existence after writing, that is the interesting thing for the end user)
2. since it was mocking a lot, it was even breaking the usage of simple functions like `tarfile.is_tarfile()` (I'm using to test if the associated file is reliable, required for the new output, since it's half on-disk), I guess because being a stream it was not possible to reread it, but it would have been a nightmare to fix it for the test

felixhekhorn:
> @felixhekhorn I removed the usage of `unittest` in [e7be4bb](https://github.com/N3PDF/eko/pull/105/commits/e7be4bb5199dcbadd3083802869195b2989ed564) for two reasons:

fine

>     1. it's true that `unittest` is standard library, nevertheless it is a further testing framework; since we're already using `pytest`, let's be as much consistent as possible and always use only that one (the thing it was tested by `unittest`, the mode in which the file was opened, it was unnecessarily detailed... if you wish we can improve current status by testing file existence after writing, that is the interesting thing for the end user)

that we should do

felixhekhorn:
As I realized in https://github.com/N3PDF/pineappl/pull/138 it is stupid to store the alpha_s values along side (since you don't know the renomalization scale of the process), so if you wish you can stop propagating that information here ...

AleCandido:
> As I realized in [N3PDF/pineappl#138](https://github.com/N3PDF/pineappl/pull/138) it is stupid to store the alpha_s values along side (since you don't know the renomalization scale of the process), so if you wish you can stop propagating that information here ...

Changing things add layers over layers of refactoring. I'm just out of a few, I'd like to close this one without any other big change, if possible.

Not that I don't want to do it, but simply open a separate issue for it. I'll do the same with the `from_dict` extermination campaign.

felixhekhorn:
> From now on, we should refrain to address multiple big tasks in a single PR

while I agree with that in theory, in practice this is much more difficult ... (as typically problems appear along the way)

AleCandido:
> while I agree with that in theory, in practice this is much more difficult ... (as typically problems appear along the way)

In the case of this PR this is actual refactoring, not addition, and refactoring the ground of such a big project it's rather expected to be messy, since it's breaking compatibility of everything (in case you didn't properly designed your internal interfaces...).

Hopefully, I'll be able to merge right after #83

AleCandido:
Ok, I finally rebased this branch, I expect everything to break now.

In any case, it was already broken (as usual...), so now the project is:
- [ ] recover it once more (as soon as possible)
- [ ] implement a bunch of quick refinements
- [ ] merge before #115

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+import numpy.lib.npyio as npyio
+import yaml
+
+from .. import basis_rotation as br
+from .. import interpolation
+from .. import version as vmod
+
+logger = logging.getLogger(__name__)
+
+PathLike = typing.Union[str, os.PathLike]
```
felixhekhorn, line 26: is there no standard type? this should appear all over the place ...

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+@dataclass
+class Operator(DictLike):
+    alphas: float
+    operator: np.ndarray
+    error: Optional[np.ndarray] = None
+
+    # IO works with streams in memory, in order to avoid intermediate write on
+    # disk (keep read from and write to tar file only)
+
+    def save(self, compress: bool = True) -> Tuple[io.BytesIO, bool]:
```
felixhekhorn, line 64: why is stream not coming as argument here? I would expect the method to write to it (think of `<<` in c++)

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+        return dictionary
+
+
+@dataclass
+class Operator(DictLike):
+    alphas: float
+    operator: np.ndarray
+    error: Optional[np.ndarray] = None
+
+    # IO works with streams in memory, in order to avoid intermediate write on
```
felixhekhorn, line 61: comments are good - docstrings are better ;-)

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+            dictionary[field.name] = value
+
+        return dictionary
+
+
+@dataclass
+class Operator(DictLike):
+    alphas: float
+    operator: np.ndarray
+    error: Optional[np.ndarray] = None
```
felixhekhorn, line 59: I don't have a better suggestion, but "error" is quite a general name ...

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+                value = float(value)
+
+            dictionary[field.name] = value
+
+        return dictionary
+
+
+@dataclass
+class Operator(DictLike):
+    alphas: float
```
felixhekhorn, line 57: to be dropped, as said in #113

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+        debug configurations
+    version : str
+        library version used to create the corresponding file
+    data_version : str
+        specs version, to which the file adheres
+
+    """
+
+    # operators cache, contains the Q2 grid information
+    _operators: Dict[float, Optional[Operator]]
```
felixhekhorn, line 188: - `Dict` [is deprected](https://docs.python.org/3/library/typing.html#typing.Dict)
- making the value type optional seems wrong ...

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+        if not tarfile.is_tarfile(self.path):
+            raise ValueError("EKO: the corresponding file is not a valid tar archive")
+
+    def __getitem__(self, q2: float):
+        # TODO: autoload
+        op = self._operators[q2]
+        if op is not None:
+            return op
+
+        start = f"operators/{q2:8.2f}"
```
felixhekhorn, line 216: delegate to a method `def op_name(q2)` (since it is repeated and crucial) - or actual you can also just expose as (module-wide) global format string

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+            if stream is None:
+                raise ValueError
+
+            op = Operator.load(stream, compressed=compressed)
+
+        self._operators[q2] = op
+        return op
+
+    def __setitem__(self, q2: float, op: Optional[Operator], compress: bool = True):
+        if op is not None:
```
felixhekhorn, line 241: Why would it be `None`?

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+        if close.size == 1:
+            return close[0]
+        if close.size == 0:
+            return None
+        raise ValueError(f"Multiple values of Q2 have been found close to {q2}")
+
+    @staticmethod
+    def bootstrap(tdir: PathLike, theory: dict, operator: dict):
+        tdir = pathlib.Path(tdir)
+        (tdir / "theory.yaml").write_text(yaml.dump(theory), encoding="utf-8")
```
felixhekhorn, line 347: same as before -> module constants

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+        Returns
+        -------
+        EKO
+            the generated structure
+
+        """
+        path = pathlib.Path(
+            path if path is not None else tempfile.mkstemp(suffix=".tar")[1]
+        )
+        path.unlink()
```
felixhekhorn, line 456: this needs to be confirmed by an external argument

src/eko/output/struct.py
```diff
@@ -0,0 +1,502 @@
...
+        tuple
+            couples of ``(q2, operator)``, loaded immediately before, unloaded
+            immediately after
+
+        """
+        for q2 in self.Q2grid:
+            yield q2, self[q2]
+            del self[q2]
+
+    def __contains__(self, q2: float) -> bool:
```
felixhekhorn, line 331: this is complicated because `__contains__` and `items()` do not correlate (on in memory the other on disk)

src/eko/output/struct.py
```diff
@@ -254,6 +255,17 @@ def __setitem__(self, q2: float, op: Optional[Operator], compress: bool = True):
             #  info.uname = os.getlogin()
             #  info.gname = os.getlogin()
 
+            # TODO: unfortunately Python has no native support for deleting
```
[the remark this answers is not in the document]
AleCandido: If you use "w" you will overwrite the whole file, completely erasing whatever was there.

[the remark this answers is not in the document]
AleCandido: I thought enough on this issue. The alternative solution is written there. The other two solutions are:
- make our package, wrapping some C library to manipulate tar files (if exists in Rust is just easier)
- add an issue on Python repository, and wait for them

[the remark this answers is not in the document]
felixhekhorn: it seems the only thing you *can* actually do is to rewrite the archive minus the file you want to delete (and yes we can add a short wrapper for that) - this is even what [the officicial GNU implementation does](https://www.gnu.org/software/tar/manual/html_node/delete.html#delete): "This operation will rewrite the archive."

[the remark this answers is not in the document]
AleCandido: Consider that a `tar` is just a bunch of consequential bytes...
Maybe there are clever things you can do, but it's reasonable that you can't do everything.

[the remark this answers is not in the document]
AleCandido: Actually, this suggests me that maybe we should not use an archive for an actual operator: the moment we connect maybe we should use an actual temporary folder. Blown up from the archive, if we're connecting an existing operator, or just created from scratch, and then finally generate the archive only at saving time.

I will definitely do this, but I will postpone to a separate PR.  Keep tracking in #139

felixhekhorn:
Re `test_to_evol`:
1. this https://github.com/N3PDF/eko/blob/822d552fef955f4274df7db05fa39c9912c263c0/src/eko/output/struct.py#L265 seems not true - if you run `poe test -k test_to_evol  -s` (note the additional `-s`) you get `tar: operators/    2.00.npz.lz4: Not found in archive
tar: Exiting with failure status due to previous errors` - the commit contains a simple fix for this
2. I also added a non-fix for the remaining `test_to_evol` - it is working, but I'm cheating (I'm tricking the autodelete function ...)
3. However, I believe I just realized the true source of the problem: `copy.deepcopy` copies the whole object (who ẃhould have thought ^^) - including *the file reference* and so instead of looking at 4 different objects in that specific test we're actually looking at one and the same rewritten over and over; you need to implement a `copy` or we need to trick the test in some way ... however, all that is not so trivial ...
4. actually, I wonder how all the other tests did work out (e.g. `test_xgrid_reshape` should have a similar problem ...

felixhekhorn:
And now I know how to bypass pre-commit - that is dangerous ... :see_no_evil:

src/eko/output/manipulate.py
```diff
@@ -137,7 +137,9 @@ def flavor_reshape(
         inv_inputbasis = np.linalg.inv(inputbasis)
 
     # build new grid
-    for q2, elem in eko.items():
+    # for q2, elem in eko.items():
+    for q2 in eko.Q2grid:
+        elem = eko[q2]
```
AleCandido, line 142: Wait, why not using `items()`?
(moreover, `Q2grid` here is not needed, since you only need to generate the array, `EKO` has already `__iter__` defined that does the exact same job)

src/eko/output/struct.py
```diff
@@ -262,13 +262,16 @@ def __setitem__(self, q2: float, op: Optional[Operator], compress: bool = True):
             # at the moment, an implicit dependency on `tar` command has been
             # introduced -> dangerous for portability
             # since it's not raising any error, it is fine to run in any case:
-            # if the file is not there, nothing happens
-            subprocess.run(
-                f"tar -f {self.path.absolute()} --delete".split() + [info.name]
-            )
+            has_file = False
```
AleCandido, line 265: in case you're doing for performances, at least remove both lines of the comment (not just the second one) and replace with the reason

otherwise in a few weeks we'll wonder why we are doing this

AleCandido: if it's not for performances, then it's actually doing nothing...

src/eko/output/struct.py
AleCandido: This one I solved in another way, but I guess it does what you wanted.
In any case, a more stable solution will come with #139

AleCandido:
> And now I know how to bypass pre-commit - that is dangerous ... :see_no_evil:

Extremely easy, you just need to append `-n`:

```sh
git commit -m "My message" -n
```

src/eko/output/struct.py
AleCandido: I don't know, I thought `os.PathLike` was the standard type, but most likely the static analyzer complained, I will try again without and check for the issues.

AleCandido:
> 3\. you need to implement a `copy` or we need to trick the test in some way ... however, all that is not so trivial ...

Added

src/eko/output/struct.py
AleCandido: I guess in case you want to erase it? but that's just stupid, I'll drop

src/eko/output/struct.py
AleCandido: Think namespaced: `Operator.error`, it is rather specific:

> Namespaces are one honking great idea -- let's do more of those!

src/eko/output/struct.py
AleCandido: It has been

src/eko/output/struct.py
AleCandido: I know, is deprecated, but only since 3.9. I need this to support 3.8.
It is not, as written somewhere else, the in-memory operator **is** optional. Because it might simply be on disk.

src/eko/output/struct.py
AleCandido: do you mean `"theory.yaml"`, it might be a good idea. I agree about getting rid of duplicated strings (it was part of the idea of this module in the first place)

src/eko/output/struct.py
felixhekhorn: see [above](https://github.com/N3PDF/eko/pull/105#issuecomment-1190523057)

src/eko/output/manipulate.py
felixhekhorn: see [above](https://github.com/N3PDF/eko/pull/105#issuecomment-1190523057)

src/eko/output/struct.py
felixhekhorn: postponed to #139

src/eko/output/struct.py
felixhekhorn: hm? also f-strings can be constants - simply without the f and then you "call" them with `var.format(a=b)` (the f-syntax is just sugar)

src/eko/output/struct.py
felixhekhorn: agreed

src/eko/output/struct.py
AleCandido: Solved in 1b8bb42bfa5a9f98670ca007dd78eb67ee9c4dda

src/eko/output/struct.py
AleCandido: Solved in a8a5818eb007cd1cc3b8815216a7784f001f7e0a

src/eko/output/struct.py
AleCandido: Solved in a8a5818eb007cd1cc3b8815216a7784f001f7e0a

src/eko/output/struct.py
AleCandido: Solved in a621d847da996f9498c2a32bf0658d4d8aa28432

src/eko/output/struct.py
AleCandido: Solved in 1be6fa4e8faedfa49065fbacfa5adcb86b3554d0

AleCandido:
@felixhekhorn replacing this:
https://github.com/N3PDF/eko/blob/21f7b430662a2a029193b3f6e74d995eaca31afe/src/eko/output/manipulate.py#L138-L141
with `eko.items()` (as in the comment) makes a single test failing (`tests/eko/test_output.py::TestManipulate::test_to_evol`).

I'm not fully understanding what is going on, but for sure is a bug: the two should have the same behavior. I know that the difference is that done in this way, the `EKO` is not unloading operators, so they are all kept in memory.
Thus, there is a problem while reloading them (I don't know if it's in the code, or in the test itself).

Can you investigate?

felixhekhorn:
- Commit https://github.com/N3PDF/eko/pull/105/commits/a57be9d459b02cbcc03f5aceee3abd8209d1070e brings back the former tests and it was quite tough to get them back - as you do have a design flaw in you implementation
- currently there is no clear way to distinguish the requested and the computed operators; e.g. the implementation of `load_tar` was challenging the native constructor; this also broke the `fake_output` implementation - I saw somewhere (don't remember where) you made an attempt, but this was not really clear and for sure not sufficient; my implementation is also not very good, but it was the smallest effort to not rewrite all tests ... however consider to fix this in #138 
- before merging you *must* fix the benchmarks which are currently *all* broken - reasons are these lines https://github.com/N3PDF/eko/blob/a57be9d459b02cbcc03f5aceee3abd8209d1070e/src/eko/compatibility.py#L35-L36 which is just crazy - the compatibility layer is exactly made to transform into the new layout; i.e. here you bring the flat opcard into your sections

AleCandido:
> which is just crazy - the compatibility layer is exactly made to transform into the new layout; i.e. here you bring the flat opcard into your sections

It's not crazy: it's simply not done.

The error is raised to have an explicit reason, instead of crashing somewhere else for an unknown reason. It's true that I could have used `NotImplementedError`.

EKO never had a compatibility layered: it has been just initialized by @niclaurenti to support QED configs reasonably, and I added that bit when I rebased this PR a few days ago.

AleCandido:
@felixhekhorn now we should be finally back to the situation of this morning (tests working). Plus, we also have `compatibility` in place, and we got rid of `InterpolatorDispatcher.from_dict`.

Now, we should complete coverage (if you can take care, I'd be grateful) and docstrings (I'll do myself).

Once done, we can merge (everything else is postponed, if possible to many smaller PR).

AleCandido:
Docstrings for `struct.py` provided, for the rest, let's go incrementally.
We knew that fully documenting `eko` would have been a massive effort (even though it is already pretty good, but not uniform in style).

felixhekhorn:
The only remaining lines are the lovely dead code you promised to take care of ...

felixhekhorn:
while https://github.com/N3PDF/eko/pull/105/commits/44b581cc497008e0c0da33162cb94e26a818a584 brings LHA LO back running, I'm afraid we're hit by #123 again, because now I'm trying to run NLO and since 10 mins nothing is happening (except my computer taking 100% energy) and I believe he is compiling matching ... @giacomomagni can you confirm that you can run NLO (e.g. on master meaning in your FK table generation)? this PR should have no impact on compiling/matching ...

felixhekhorn:
- I didn't recheck everything, but I'm relying on the tests, benchmarks, and I rerun the lha benchmark (eventually discovering another inconsistency)
- since the output tutorial is now outdated we need to rewrite it
- we also may want to delay the merge to develop until the resubmission of the paper which may trigger a 0.10.3

src/eko/evolution_operator/__init__.py
```diff
@@ -436,8 +436,11 @@ def run_op_integration(
                 temp_dict[label] = res[:2]
             column.append(temp_dict)
         logger.info(
-            f"{self.log_label}: computing operators: - {k+1}/{self.grid_size}"
-            f" took: {(time.perf_counter() - start_time):6f} s"
+            "%s: computing operators - %u/%u took: %6f s",
+            self.log_label,
+            k + 1,
+            self.grid_size,
+            (time.perf_counter() - start_time),
```
AleCandido, line 443: Actually, why? .-.

felixhekhorn: I can't find the blog post where I read it first, but lazy logging is actually recommanded both for performance and security reasons, because this way you ensure lazy evaluation ... it's also written here: https://medium.com/flowe-ita/logging-should-be-lazy-bc6ac9816906, https://medium.com/swlh/why-it-matters-how-you-log-in-python-1a1085851205

AleCandido: Good to know

felixhekhorn:
> * since the output tutorial is now outdated we need to rewrite it

https://github.com/N3PDF/eko/pull/105/commits/47e30e518ae0aa1b34bd8d2b2d0bb7d0b7b479de makes the tutorial back running with a minimal presentation

Issue #94, opened 2022-02-11
Minimize the amount of initial imports


AleCandido:
**Motivation**: As soon as you install `eko`, you'll need to wait a while on first import. This is needed for the `eko` *application*, but unnecessary and completely counter-intuitive for the `eko` library.

At the moment importing `eko` means importing `eko.runner`, and also `eko.output`.

Then, as it makes sense, importing runner means importing everything (see [dependency graph](https://eko.readthedocs.io/en/latest/code/Dependency.html)).

### How to improve

From my point of view:
- there is no way to improve on `eko.runner`: it's quite intuitive that uses all (or almost all) of the library
- but I'd say that we should not wait on the compilation of all the anomalous dimensions and so on, if you only need interpolation or strong coupling (or even masses, basis rotation, ...)

In particular, having to wait every time you install anew it's already a bit unpleasant in `yadism`, but much more in `banana` or `runcards`.

### In practice

I'd say we just need to:
- [ ] shuffle back `run_dglap` inside `eko.runner`
- [ ] at this point we can keep the main `__init__.py` just for the version (no need for a separate module)
- [ ] update related docs

AleCandido:
I believe we don't need this any longer, because of:
- mainly the `numba` rearrangement
- and the forthcoming runner-less structure (in which runner will be optional, and thus by default importing much less)

Can we close this? @giacomomagni @felixhekhorn

giacomomagni:
okay, fine

felixhekhorn:
Agreed

Issue #78, opened 2021-10-20
Speed up first import


AleCandido:
Compilation is needed only once per installation, nevertheless it is a bit annoying that actually the first time you try to run you don't have any feedback for a while.

There is an easy way to speed up compilation, since at the moment we are doing it on a single core. We can just add to the decorators:
https://github.com/N3PDF/eko/blob/47ad7d011bd853a86fe2cf7239931e28addb958b/src/eko/kernels/evolution_integrals.py#L18
the simple kwarg `parallel=True`.

https://numba.pydata.org/numba-doc/latest/user/threading-layer.html

> If a code base does not use the threading or multiprocessing modules (or any other sort of parallelism) the defaults for the threading layer that ship with Numba will work well, no further action is required!

scarrazza:
Sorry @AleCandido, I was talking to Felix and what you say is completely wrong XD.
Parallel true, allows you to use numba.prange and he automatic parallelization through omp, tbb, etc. 
This has nothing to do with AOT or JIT compilation.

AleCandido:
Got it.

After investigation, looks like there is no way in `numba` to control compilation options (and yes, it seems to compile on a single core).
Most likely the multithreaded compilation is still lacking in `llvm`: https://reviews.llvm.org/D52193 (it looks like a lot of compilers are still single threaded, e.g. even `rustc` is based on `llvm`, so most likely the parallelization is at the level of build tool).

To improve, there are still multiple ways available:
1. post install scripts (to compile code at install time)
2. [distribute cache](https://numba.pydata.org/numba-doc/latest/developer/caching.html#cache-sharing) built in a workflow
3. hope for `numba` doing something

Comments on the three:
1. poetry does not support this (see this python-poetry/poetry/issues/265), and it is fine, fewer things to debug (that might even be more annoying to)
2. this would make `eko` architecture dependent, and we should distribute multiple package with built extensions
3. actually, I didn't find anywhere (neither in [`cache` introduction](https://numba.pydata.org/numba-doc/latest/user/jit.html#cache), nor in [related notes](https://numba.pydata.org/numba-doc/latest/developer/caching.html)) that the function should be cached on `import`, and it would be meaningful to cache it on first run since it is still [`jit`](https://numba.pydata.org/numba-doc/latest/user/jit.html) and not [AOT](https://numba.pydata.org/numba-doc/latest/user/pycc.html); if it would really behave like this it'd be much more intuitive: non-blocking until an actual computation is expected

[...]

PR #108, opened 2022-03-18
Recover numba

felixhekhorn:
Here I want to recover the "true" numba behaviour, i.e. compilation just-in-time (and not ahead-of-time)
- I think it is just a matter of removing all signatures (as is done in this commit)
- then `import eko` is immediate and instead running, e.g. the LHA benchmark, has some initialization cost
- tests are still running* and the LHA benchmark is still working -  so I think this works

Please have a look and then we can merge back to #83 

*actually, they are not - because https://github.com/N3PDF/eko/blob/606138fcfc31e0364457ab49e9e9557597e49a42/src/eko/evolution_operator/__init__.py#L243 forces me to do a manual `CTRL+C` in the test shell - and hence I strongly recommend to revert that

PS: I think we introduced the signatures because at some point we had problems with types, i.e. we were not sufficiently casting - hopefully this is gone now ...

### DIFF
--- a/src/eko/evolution_operator/__init__.py
+++ b/src/eko/evolution_operator/__init__.py
@@ -240,7 +240,6 @@ class Operator:
     """
 
     log_label = "Evolution"
-    n_pools = os.cpu_count()
     # complete list of possible evolution operators labels
     full_labels = br.full_labels
 
@@ -255,6 +254,13 @@ class Operator:
         self.op_members = {}
 
     @property
+    def n_pools(self):
+        n_pools = self.config["n_integration_cores"]
+        if n_pools > 0:
+            return n_pools
+        return os.cpu_count() + n_pools
+
+    @property
     def fact_to_ren(self):
         r"""Returns the factor :math:`(\mu_F/\mu_R)^2`"""
         return self.config["fact_to_ren"]
@@ -264,8 +270,7 @@ class Operator:
         """Returns the scale variation mode"""
         if self.config["ModSV"] is not None:
             return sv.Modes[self.config["ModSV"]]
-        else:
-            return sv.Modes.unvaried
+        return sv.Modes.unvaried
 
     @property
     def int_disp(self):
@@ -466,11 +471,18 @@ class Operator:
         tot_start_time = time.perf_counter()
 
         # run integration in parallel for each grid point
-        with Pool(self.n_pools) as pool:
-            res = pool.map(
+        # or avoid opening a single pool
+        if self.n_pools == 1:
+            res = map(
                 self.run_op_integration,
                 enumerate(np.log(self.int_disp.xgrid_raw)),
             )
+        else:
+            with Pool(self.n_pools) as pool:
+                res = pool.map(
+                    self.run_op_integration,
+                    enumerate(np.log(self.int_disp.xgrid_raw)),
+                )
 
         # collect results
         for k, row in enumerate(res):
--- a/src/eko/evolution_operator/__init__.py
+++ b/src/eko/evolution_operator/__init__.py
@@ -26,7 +26,7 @@ from ..member import OpMember
 logger = logging.getLogger(__name__)
 
 
-@nb.njit("c16(c16[:,:],u2,u2)")
+@nb.njit(cache=True)
 def select_singlet_element(ker, mode0, mode1):
     """
     Select element of the singlet matrix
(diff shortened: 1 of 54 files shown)


giacomomagni:
Okay. It's working so fine for me. 
Regarding:
https://github.com/N3PDF/eko/blob/606138fcfc31e0364457ab49e9e9557597e49a42/src/eko/evolution_operator/__init__.py#L243
I think we should just set it to one in all the test instances of `Operator` and `OperatorMatrixElement` inside tests. It's useless to have more pools there.

felixhekhorn:
> I think we should just set it to one in all the instances of `Operator` and `OperatorMatrixElement` inside tests. It's useless to have more pools there.

I think that would work - you need to make that option accessible from the outside though

I still think this parallelization is a good a idea - since it is an immediate help for us developers as it speeds up the computation of a single eko (as we typically do), instead #105 is for users who typically request several Q2. This can addressed here or back in #83 where it was introduced

felixhekhorn:
this should correctly address #78. We can still keep #94 - or close it.

giacomomagni:
> I think that would work - you need to make that option accessible from the outside though

Shall I add an option to the operator card? something like: `compute_parellel: True/False`

felixhekhorn:
> Shall I add an option to the operator card? something like: `compute_parellel: True/False`

maybe yes - the name has to be more specific though I think (or maybe it's just fine, since as said #105 plays on a different level and is parallelize on the outside) ... or maybe pass directly the number of processes? with 0=all or something like this (and -1 = all-1)

AleCandido:
The most important question: is the cache still working?

AleCandido:
> maybe yes - the name has to be more specific though I think (or maybe it's just fine, since as said #105 plays on a different level and is parallelize on the outside) ... or maybe pass directly the number of processes? with 0=all or something like this (and -1 = all-1)

Definitely, I'd put in the runcard the number of processor, this is something you always want to control.

In #105, I started splitting the operator card in several chunks, so this option can now be scoped in `configs:` section, that is dedicated to the internal configurations of `eko`.
At the moment contain part of what we believe should be in theory (`ev_op...` and `backward_method`), but eventually it should contain things like the Mellin paths (and maybe its parameters), quad precision and these kind of things.

AleCandido:
Actually, there will be no clash with parallelization itself w.r.t. to #105 and related, since `eko` will never provide `Q2` parallelization itself, but it will just split the jobs.

Jobs management should be done:
- directly by the user, or
- in a dedicated runner in `ekobox`

We have to mandatory split the jobs in `eko`, otherwise no one would have a handle on them, but managing the jobs is beyond `eko`'s scope.

felixhekhorn:
> The most important question: is the cache still working?

Yes, I think so - but please do a simple run yourself

(not that this has an interplay with the parallelization, since it seems you need to compile numpy in numba (`numba.np`) on each core)

AleCandido:
> this should correctly address #78. We can still keep #94 - or close it.

If there is not any longer the problem of initial slow down (i.e. slow down on `import`, not on run), for me, we can close even #94 (i.e. link the issue to this PR)

giacomomagni (approved):
I'm happy to merge.
(All tests run on single core, while benchmarks are still in parallel)

AleCandido (approved):
Nothing but signatures and `multiprocessing.Pool` related stuffs. I'm happy to merge as well.

In case we'll want the signatures back, they are still in `git` anyhow, so we can recover them with reasonable to little effort in the next future (more changes more effort, of course, but even less likely we'd like to change).
Not that we should, I believe you this is working as expected, just to point that we *could*.

src/ekomark/data/operators.py
```diff
@@ -18,6 +19,7 @@
     ev_op_max_order=10,
     ev_op_iterations=10,
     backward_inversion="expanded",
+    n_integration_cores=os.cpu_count(),
```
felixhekhorn, line 22: The way you set it up - you can just pass 0

src/eko/evolution_operator/__init__.py
```diff
@@ -466,11 +471,18 @@ def integrate(
         tot_start_time = time.perf_counter()
 
         # run integration in parallel for each grid point
-        with Pool(self.n_pools) as pool:
-            res = pool.map(
+        # or avoid opening a single pool
+        if self.n_pools == 1:
```
felixhekhorn, line 475: If you wish you can be even stricter and define before `args = (a,b)` and then `map(*args)`

felixhekhorn (commented):
Two minor things - else we can merge and move on ...

PR #109, opened 2022-03-21
ekobox unittests

felixhekhorn:
- moved some code from benchmarks back to tests (those unrelated to lhapdf)

### DIFF
--- a/benchmarks/ekobox/benchmark_evol_pdf.py
+++ b/benchmarks/ekobox/benchmark_evol_pdf.py
@@ -6,7 +6,7 @@ import pytest
 
 from eko import basis_rotation as br
 from ekobox import evol_pdf as ev_p
-from ekobox import gen_op as g_o
+from ekobox import operators_card as oc
 from ekobox import theory_card as tc
 from ekobox.genpdf import load
 
@@ -21,7 +21,7 @@ def benchmark_evolve_single_member(tmp_path, cd, lhapdf_path):
         100.0,
         10000.0,
     ]
-    op = g_o.gen_op_card(q2grid)
+    op = oc.generate(q2grid)
     theory = tc.generate(
         0,
         5.0,
@@ -77,7 +77,7 @@ def benchmark_evolve_single_member(tmp_path, cd, lhapdf_path):
 
 @pytest.mark.isolated
 def benchmark_evolve_more_members(tmp_path, cd, lhapdf_path):
-    op = g_o.gen_op_card([10, 100], update={"xgrid": [1e-7, 0.01, 0.1, 0.2, 0.3]})
+    op = oc.generate([10, 100], update={"xgrid": [1e-7, 0.01, 0.1, 0.2, 0.3]})
     theory = tc.generate(0, 1.0)
     with lhapdf_path(test_pdf):
         pdfs = lhapdf.mkPDFs("myMSTW2008nlo90cl")
--- a/benchmarks/ekobox/benchmark_evol_pdf.py
+++ b/benchmarks/ekobox/benchmark_evol_pdf.py
@@ -1,11 +1,9 @@
 # -*- coding: utf-8 -*-
 import pathlib
 
-import lhapdf
 import numpy as np
 import pytest
 
-import eko.output.legacy
 from eko import basis_rotation as br
 from ekobox import evol_pdf as ev_p
 from ekobox import gen_op as g_o
@@ -13,6 +11,7 @@ from ekobox import gen_theory as g_t
 from ekobox.genpdf import load
 
 test_pdf = pathlib.Path(__file__).parent / "fakepdf"
+lhapdf = pytest.importorskip("lhapdf")
 
 
 @pytest.mark.isolated
@@ -91,21 +90,3 @@ def benchmark_evolve_more_members(tmp_path, cd, lhapdf_path):
         _ = lhapdf.mkPDFs("Debug")
         info = load.load_info_from_file("Debug")
     assert info["XMin"] == op["xgrid"][0]
-
-
-@pytest.mark.isolated
-def benchmark_gen_and_dump_out(tmp_path):
-    op = g_o.gen_op_card([100.0], update={"xgrid": [1e-7, 0.01, 0.1, 0.2, 0.3]})
-    theory = g_t.gen_theory_card(0, 1.0)
-
-    out = ev_p.gen_out(theory, op, path=tmp_path)
-
-    ops_id = f"o{op['hash'][:6]}_t{theory['hash'][:6]}"
-    outpath = f"{tmp_path}/{ops_id}.tar"
-    loaded_out = eko.output.legacy.load_tar(outpath)
-    assert list(out.xgrid) == list(loaded_out.xgrid)
-    for el, load_el in zip(out[100.0].operator, loaded_out[100.0].operator):
-        np.testing.assert_allclose(
-            out[100.0].operator,
-            loaded_out[100.0].operator,
-        )
--- a/benchmarks/ekobox/benchmark_evol_pdf.py
+++ b/benchmarks/ekobox/benchmark_evol_pdf.py
@@ -7,7 +7,7 @@ import pytest
 from eko import basis_rotation as br
 from ekobox import evol_pdf as ev_p
 from ekobox import gen_op as g_o
-from ekobox import gen_theory as g_t
+from ekobox import theory_card as tc
 from ekobox.genpdf import load
 
 test_pdf = pathlib.Path(__file__).parent / "fakepdf"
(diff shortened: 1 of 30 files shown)


felixhekhorn:
@andreab1997 please take a look to `ekobox` and tell me what you think ... I did some renaming (e.g. `gen_theory.gen_theory_card` -> `theory_card.generate`) so you would most likely need to adjust the n3fit fit side ...

also I'm not really convinced of providing default runcards from banana: 1. are we really sure we want to do this? (instead of sticking to our default-less scheme - and in NNPDF we always have full runcards) and 2. if still, I don't think we should inherit from banana: our default theory there is FFNS3 which is good for debugging, but stupid in production ...

andreab1997:
> @andreab1997 please take a look to `ekobox` and tell me what you think ... I did some renaming (e.g. `gen_theory.gen_theory_card` -> `theory_card.generate`) so you would most likely need to adjust the n3fit fit side ...
> 
Yes I'll take a look. Probably is only a matter of changing some names in evolven3fit. 

> also I'm not really convinced of providing default runcards from banana: 1. are we really sure we want to do this? (instead of sticking to our default-less scheme - and in NNPDF we always have full runcards) and 2. if still, I don't think we should inherit from banana: our default theory there is FFNS3 which is good for debugging, but stupid in production ...

I have not a strong opinion on that. I believe that for users having some defaults is good because most of them would only use a few of the runcards options. However maybe we can provide less and more "standard" defaults.

andreab1997:
> @andreab1997 please take a look to `ekobox` and tell me what you think ... I did some renaming (e.g. `gen_theory.gen_theory_card` -> `theory_card.generate`) so you would most likely need to adjust the n3fit fit side ...
> 
> also I'm not really convinced of providing default runcards from banana: 1. are we really sure we want to do this? (instead of sticking to our default-less scheme - and in NNPDF we always have full runcards) and 2. if still, I don't think we should inherit from banana: our default theory there is FFNS3 which is good for debugging, but stupid in production ...

Anyway, I really like the new names. The old ones were very confusing

felixhekhorn:
> Anyway, I really like the new names. The old ones were very confusing

you know whom to blame :rofl: :innocent:

andreab1997:
> > Anyway, I really like the new names. The old ones were very confusing
> 
> you know whom to blame 🤣 😇

yes, myself! xD

andreab1997:
> @andreab1997 please take a look to `ekobox` and tell me what you think ... I did some renaming (e.g. `gen_theory.gen_theory_card` -> `theory_card.generate`) so you would most likely need to adjust the n3fit fit side ...
> 
> also I'm not really convinced of providing default runcards from banana: 1. are we really sure we want to do this? (instead of sticking to our default-less scheme - and in NNPDF we always have full runcards) and 2. if still, I don't think we should inherit from banana: our default theory there is FFNS3 which is good for debugging, but stupid in production ...

I changed the names on evolven3fit and everything should work again

felixhekhorn:
> I changed the names on evolven3fit and everything should work again

Mmm that might have been premature ... you maybe want to stash that commit until this branch is released on main, which might take a while ...

andreab1997:
> > I changed the names on evolven3fit and everything should work again
> 
> Mmm that might have been premature ... you maybe want to stash that commit until this branch is released on main, which might take a while ...

Oh I see. Yes I will do that

felixhekhorn:
@AleCandido maybe consider to merge this PR first before continuing on #105

benchmarks/ekobox/benchmark_evol_pdf.py
```diff
@@ -1,18 +1,17 @@
...
 from eko import basis_rotation as br
 from ekobox import evol_pdf as ev_p
-from ekobox import gen_op as g_o
-from ekobox import gen_theory as g_t
+from ekobox import operators_card as oc
+from ekobox import theory_card as tc
 from ekobox.genpdf import load
 
 test_pdf = pathlib.Path(__file__).parent / "fakepdf"
+lhapdf = pytest.importorskip("lhapdf")
```
AleCandido, line 14: I guess here you can even do
```suggestion
import lhapdf
```

It should be provided in the testing environment. However, even fine like this.

src/ekobox/__init__.py
```diff
@@ -1,4 +1,4 @@
 # -*- coding: utf-8 -*-
 from ekomark import apply
 
-from . import evol_pdf, gen_info, gen_op, gen_theory
+from . import evol_pdf, info_file, operators_card, theory_card
```
AleCandido, line 4: I guess `info_file` might even be just `info`, but also fine like this.

src/ekobox/evol_pdf.py
```diff
@@ -100,28 +90,5 @@ def evolve_pdfs(
...
-        : eko.output.Output
-            eko output
-    """
-    eko_output = eko.run_dglap(theory_card, op_card)
-    if path is not None:
-        ops_id = f"o{op_card['hash'][:6]}_t{theory_card['hash'][:6]}"
-        path = f"{path}/{ops_id}.tar"
-        eko.output.legacy.dump_tar(eko_output, path)
-    return eko_output
+def ekofileid(theory_card, operators_card):
```
AleCandido, line 93: Missing docstring

src/ekobox/info_file.py
```diff
@@ -22,8 +22,8 @@ def create_info_file(theory_card, operators_card, num_members, info_update):
 
     Returns
     -------
-        : dict
-        info file in lhapdf format
+        dict :
```
AleCandido, line 25: Since you're not specifying a name, you don't need a colon at all, neither before nor after.

Please, follow the reference https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html

src/ekobox/operators_card.py
```diff
@@ -7,27 +7,27 @@
...
 
     Parameters
     ----------
-    Q2grid : list(float)
-        grid for Q2
-    update : dict
-        dictionary of info to update in op. card
-    name : str
-        name of exported op.card (if name not None)
+        Q2grid : list(float)
```
AleCandido, line 20: Again: this indentation and the above newline for the first sentence is non-standard.

Please, follow the referene: https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html

src/ekobox/operators_card.py
```diff
@@ -46,40 +46,39 @@ def gen_op_card(Q2grid, update=None, name=None):
...
-    name : str
-        name of the op. card to export
-
-    op : dict
-        op card
-
+        name : str
+            name of the operators card to export
+        op : dict
+            operators card
```
AleCandido, line 62: Once more: no indentation, no newline at the beginning.

I'm not going to update all at once all the docstrings, but do not revert the correct ones.

src/ekobox/operators_card.py
```diff
@@ -46,40 +46,39 @@ def gen_op_card(Q2grid, update=None, name=None):
...
     ----------
         path : str
-            path to op. card in yaml format
+            path to operators card in yaml format
 
     Returns
     -------
-        : dict
-            op card
+        dict :
```
AleCandido, line 80: One more useless colon.

src/ekobox/theory_card.py
```diff
@@ -76,7 +72,7 @@ def import_theory_card(path):
 
     Returns
     -------
-        : dict
+        dict :
```
AleCandido, line 75: Useless colon.

AleCandido (changes requested):
I reviewed half, and never run. I'll do both, but for the time being I submit current comments.

benchmarks/ekobox/benchmark_evol_pdf.py
felixhekhorn: let's keep like this

src/ekobox/__init__.py
felixhekhorn: let's keep like this

src/ekobox/operators_card.py
felixhekhorn: as you maybe noticed, in pineko I don't indent, but here I wanted to stick to the scheme - you know this is due to VS code, however I'm not strongly opinioned any longer (since never using); still do you want to mix conventions?

src/ekobox/evol_pdf.py
felixhekhorn: agreed - still I hope that function will soon vanish completely

src/ekobox/operators_card.py
AleCandido: If there is already a convention, and you start applying a new one, or you spend a lot of energy in enforcing it (manually checking or spending the time to automate), or it will be broken at some point.

Since there is a convention, and it has not been decided by us, let's stick to it, with an incremental transition.

AleCandido (approved):

AleCandido:
For me this is fine, I'm going to merge and continue the development on the other branch.

PR #110, opened 2022-03-29
Fix harmonics

giacomomagni:
* [x] Move harmonics to a sub-package
* [x] Use the cached harmonics in the NLO anomalous dimension (to remove explict `mellin_g3` calls, need to rewrite all the ad expressions probably not worth... )
* [x] Use the cached harmonics in the NNLO  matching (remove explict `mellin_g3` calls )
* [x] More clean harmonics cache ordering inside N3LO matching
* [x] Expand docs section about harmonics


AleCandido:
@giacomomagni thank you very much for taking care of properly dealing with harmonics.

Actually, @niclaurenti was asking me about:

> * [ ] Use the cached harmonics in the NLO anomalous dimension (remove explict `mellin_g3` calls )

this same afternoon. It seems like cached is not used at all in NLO QCD.

The moment you finish, please ask for a review also @niclaurenti.

felixhekhorn:
> It seems like cached is not used at all in NLO QCD.

Just to confirm, this is true - moreover

https://github.com/N3PDF/eko/blob/341e6323e680d2320ae4400da61c92a445160416/src/eko/anomalous_dimensions/nlo.py#L5-L7

this is(/was) because NLO is also a special case inside PEGASUS - and I didn't wanted to duplicate that structure

AleCandido:
I guess at that time there was no point in worrying about a different and slower NLO, we had to write NNLO in the first place.

But now that we have `eko` working, do you still believe that we don't want to use cached harmonics? @felixhekhorn

felixhekhorn:
> do you still believe that we don't want to use cached harmonics? @felixhekhorn
> 
> It might be an optimization for the NLO runs.

definitely we want them! (Something we wanted to do since long ago ...) (I was just explaining why I decided like this back then)

src/eko/harmonics/__init__.py
```diff
@@ -0,0 +1,100 @@
...
+
+from .w1 import S1, Sm1
+from .w2 import S2, Sm2
+from .w3 import S3, S21, S2m1, Sm2m1, Sm3, Sm21
+from .w4 import S4, S31, S211, Sm4, Sm22, Sm31, Sm211
+from .w5 import S5, S23, S41, S221, S311, S2111, S2m3, S21m2, Sm5, Sm23, Sm221, Sm2111
+
+
+@nb.njit(cache=True)
+def base_harmonics_cache(n, max_weight=5, n_max_sums_weight=7):
```
giacomomagni, line 17: Hi @felixhekhorn, I'm trying to sort the harmonics in a more elegant way, but the thing is a bit involved:
* the best way would be to store them by weight.
* you also want the possibility to select the single index easily, since are used to compute more complex ones.
* the number of harmonics needed with a given weight depends also in the `PTO` and you don't want to compute unnecessary ones.

The solution I came up with is a matrix `(weight, max_n_harmonics_for_each_weight)` such that the first column always contains  the list of `S_{n}` and the last `S_{-n}` and the multi indices harmonics are stored in the middle always sorted by weight and index.  
For instance the complete cache will look: 
````
[
[S1,0,0,0,0,0,S-1],
[S2,0,0,0,0,0,S-2],
[S3,S21,S2-1,S-21,S-2-1,0,S-3],
[S4,S31,S211,S-22,S-211,S-31,S-4],
[S5,0,0,0,0,0,S-5],  # No need to store multi indices w5 harmonics 
]
````

It still looks a bit involved, maybe you have abetter idea?

File: src/eko/matching_conditions/as2.py [hunk 2]
```diff
@@ -268,7 +261,7 @@ def A_hg(n, sx, L):
                 * (n + 1)
                 * (n + 2)
                 * (2 + n + n**2)
-                * (10 * S1m**2 - 9 * Sp2m + 26 * S2m + 9 * Sp2p)
+                * (10 * S1m**2 + 18 * (-1) ** n * (2 * Sm2 + zeta2) + 26 * S2m)
```
felixhekhorn, line 264: Error: `(-1)**n` - I believe this can **not** appear in the implementation! I will appear in the mathematical expression, but then gets eliminated upon choosing ns+/ns-

src/eko/matching_conditions/as2.py
```diff
@@ -74,7 +74,7 @@ def A_qq_ns(n, sx, L):
 
 
 @nb.njit(cache=True)
-def A_hq_ps(n, sx, L):
+def A_hq_ps(n, S2, L):
```
felixhekhorn, line 77: I think one of the goal of this PR should be to unify the signatures* - or are you going to do this in a separate step? such that you do not need to know for e.g. `A_hq_ps` I do need exactly `S2` and not anything else ...

*do you actually agree?

src/eko/matching_conditions/as2.py
```diff
@@ -458,11 +450,11 @@ def A_singlet(n, sx, L, is_msbar=False):
         A_gq : :math:`A_{gq, H}^{S,(2)}`
         A_gg : :math:`A_{gg, H}^{S,(2)}`
     """
-    A_hq_2 = A_hq_ps(n, sx, L)
-    A_qq_2 = A_qq_ns(n, sx, L)
-    A_hg_2 = A_hg(n, sx, L)
-    A_gq_2 = A_gq(n, sx, L)
-    A_gg_2 = A_gg(n, sx, L)
+    A_hq_2 = A_hq_ps(n, sx[1, 0], L)
```
felixhekhorn, line 453: Re "unique signature": I think I prefer the old code - in this case I would sacrifice the aim of "only take those arguments you truly need" over "unique signature"; here you already did compute all elements in `sx` so the expensive cost has already been paid ... what do you think?

src/eko/matching_conditions/as2.py
giacomomagni: true true...

src/eko/matching_conditions/as2.py
giacomomagni: Okay, we can leave it as it was before...

src/eko/harmonics/__init__.py
```diff
@@ -98,3 +100,63 @@ def get_smx(n, max_weight=5):
     if max_weight >= 5:
         smx[4] = Sm5(n)
     return smx
+
+
+@nb.njit(cache=True)
+def get_s3x(n, sx, smx):
```
AleCandido, line 106: Usually, is more pythonic to call things (mainly properties, but even functions) without prepended `get_`.

Thing like `srx(...)` it's just the object, that simply depends on some parameters.

src/eko/matching_conditions/as3/__init__.py
```diff
@@ -49,15 +49,21 @@ def A_singlet(n, sx_all, nf, L):
...
-            List of harmonic sums containing:
-                [S1 ... S5, Sm1 ... Sm5, S21, S2m1, Sm21, Sm2m1, S31, S221, Sm22, Sm211, Sm31]
+            harmonic sums cache containing:
+                [
+                    [S1, 0,   0,    0,    0,     0,   Sm1],
+                    [S2, 0,   0,    0,    0,     0,   Sm2],
+                    [S3, S21, S2m1, Sm21, Sm2m1, 0,   Sm3],
+                    [S4, S31, S221, Sm22, Sm211, Sm31 Sm4],
+                    [S5, 0,   0,    0,    0,     0,   Sm5],
+                ]
```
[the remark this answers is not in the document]
felixhekhorn: agreed - but that's not exactly what I had in mind ... instead I would do something like:
```py
S1,Sm1 = sw1
S2,Sm2 = sw2
S3, S21, S2m1, Sm21, Sm2m1, Sm3 = sw3
```
again, I know, it is not much better, but
- you avoid a bunch of zeros
- you acknowledge the fact that the different weight do have intrinsically different number of member

felixhekhorn:
If I may suggest a crazy idea - and you can tell me whether you like it or not and whether it's feasable ...

we could go even a more radical step and enforce more the caching of harmonics (which we do believe is our biggest time consumer):
- we could even try to catch and cache such a line: https://github.com/N3PDF/eko/blob/d782a61ccc952e394533ccd20e2c81f49eea1f86/src/eko/harmonics/g_functions.py#L92-L93
- because we actually only need to compute `S1(N)` once and for all - and that's the complicated part
- but then `S1(N+1)` can be computed via the usual recurrence relations
- since it will be complicated at this point we would need a class as a manager
- so it would become something like `h.get("S1", N, j)`

src/eko/harmonics/g_functions.py
```diff
@@ -143,13 +148,19 @@ def mellin_g5(N):
     for k, ak in enumerate(a1):
         Nk = N + k + 1
         g5 -= ak * (
-            (k + 1) / Nk**2 * (zeta2 + cern_polygamma(Nk + 1, 1) - 2 * w1.S1(Nk) / Nk)
+            (k + 1)
+            / Nk**2
+            * (
+                zeta2
+                + cern_polygamma(Nk + 1, 1)
```
felixhekhorn, line 155: Can we trade the polygamma for an S? and so `harmonic_sum`?

src/eko/matching_conditions/as2.py
```diff
@@ -157,14 +157,14 @@ def A_hg(n, sx, L):
         A_hg : complex
             |NNLO| heavy-gluon |OME| :math:`A_{Hg}^{S,(2)}`
     """
-    S1 = sx[0, 0]
-    S2 = sx[1, 0]
-    S3 = sx[2, 0]
+    S1 = sx[0][0]
+    S2, Sm2 = sx[1]
+    if len(sx[2]) == 3:
```
felixhekhorn, line 162: This sounds a bit dangerous ... can't we unify the structure? Where does the branching come from?

giacomomagni: It's due to the different number of harmonics needed at NNLO and N3LO

felixhekhorn: I see ...

felixhekhorn:
[ce590bb](https://github.com/N3PDF/eko/pull/110/commits/ce590bb0ea3ce6ffd223f2c831c5d39d58ada763) gave a 40% speed up to NLO ad and 29% to NNLO; shall we add a similar stupid test for OME? (I guess we didn't touch those here, so nothing should change, I guess ...)

felixhekhorn:
@AleCandido converning coverage we have a similar problem here (truly in the parent branch) as in PineAPPL: isolated benchmarks are genuinely contributing now to coverage as they *do run* in parallel (and hence the `Pool` line in `ev_op/__init__`) - can we do a similar fix? i.e. [introducing flags](https://github.com/N3PDF/pineappl/commit/3789a89d007615691b022cb174f0077f8c348d1a)?

AleCandido:
> @AleCandido concerning coverage we have a similar problem here (truly in the parent branch) as in PineAPPL: isolated benchmarks are genuinely contributing now to coverage as they _do run_ in parallel (and hence the `Pool` line in `ev_op/__init__`) - can we do a similar fix? i.e. [introducing flags](https://github.com/N3PDF/pineappl/commit/3789a89d007615691b022cb174f0077f8c348d1a)?

The difference here is that unit tests are supposed (and actually are) finishing before, and we'd like to have full coverage on their own.
Moreover: they are already flagged:
https://github.com/N3PDF/eko/blob/4e5c2b1768162bab5c89df328efd8e53b9349821/.github/workflows/isobench.yml#L51-L55

So, I'm not sure of what you actually want to add...

felixhekhorn:
> we'd like to have full coverage on their own.

but then we have to run part of them in parallel (or fake parallelism) to catch those lines: https://github.com/N3PDF/eko/blob/4e5c2b1768162bab5c89df328efd8e53b9349821/src/eko/evolution_operator/__init__.py#L479-L480

AleCandido:
Actually, the `-F` is wrong in PineAPPL, I'll fix it.
https://github.com/codecov/codecov-action/blob/d8cf6802a833acb75bafac8995c5fbaafa9a5a14/README.md?plain=1#L108-L116

felixhekhorn:
@AleCandido I added `asv` as dev-dependency by hand in `pyproject.toml` ([4e5c2b1](https://github.com/N3PDF/eko/pull/110/commits/4e5c2b1768162bab5c89df328efd8e53b9349821)), but that apparently was breaking - do I need to run `poetry update` in addition? (I already ran for 10min and then gave up ...)

AleCandido:
> @AleCandido I added `asv` as dev-dependency by hand in `pyproject.toml` ([4e5c2b1](https://github.com/N3PDF/eko/pull/110/commits/4e5c2b1768162bab5c89df328efd8e53b9349821)), but that apparently was breaking - do I need to run `poetry update` in addition? (I already ran for 10min and then gave up ...)

If you don't tell `poetry` to update the lockfile in any way, then you will end up in an inconsistent state between your `pyproject.toml` and `poetry.lock`, thus it's fair that it complaints...

You can do one of three things:
1. if you want to update to the latest compatible versions all your dependencies, then yes, you have to do `poetry update`
2. if you just want to generate a new lockfile, you can just use `poetry lock`
3. if you just want to add a dependency, please consider not doing it by hand, but using `poetry add` (or `poetry add -D` for dev dependencies)

doc/source/theory/Mellin.rst
```diff
@@ -130,10 +130,10 @@ the Mellin inverse.
 - for the sums :math:`S_{-m}(N)` and :math:`m > 0` we use:
 
   .. math ::
-    S_{-m}(N) = \frac{\eta}{2} (S_{m}(N / 2) - S_{m}((N - 1) / 2)) - c_{m}
+    S_{-m}(N) = \frac{\eta}{2} (S_{m}(N / 2) - S_{m}((N - 1) / 2)) - d_{m}
 
   where formally :math:`\eta = (-1)^N` but in all singlet-like quantities it has to be analytically continued with 1
-  and with -1 elsewise and :math:`c_{m}= \left [ log(2), 1/2 \zeta_{2}, 3/4 \zeta_{3}, 7/8 \zeta_{4}, 15/16 \zeta_{5} \right]`
+  and with -1 elsewise and :math:`d_{m}= \left [ \log(2), 1/2 \zeta_{2}, 3/4 \zeta_{3}, 7/8 \zeta_{4}, 15/16 \zeta_{5}, \ldots \right]`
```
felixhekhorn, line 136: @giacomomagni this definition does not match with [MuselliPhD, Eq. (B.5.4d)] (zeta is evaluated at -k there); can you please confirm (maybe you have a second source or just a simple example, in case Muselli is wrong)? are we using this relation somewhere in the code?

felixhekhorn:
@giacomomagni in [44759d0](https://github.com/N3PDF/eko/pull/110/commits/44759d0ba93db67d8addf796f251dc5764ed7252) I wanted to add a few tests to check the `d_m` - but then I realized, the way I implemented *should not work*, because S_{-1} really is a complicated object, i.e. it depends on being singlet-like or not ...

and again: https://github.com/N3PDF/eko/blob/44759d0ba93db67d8addf796f251dc5764ed7252/src/eko/harmonics/w1.py#L62
this should *never* happen! (as far as I understand)

Please double check the mathematics in this branch! Actually even looking at #83 I can see some (-1)**n

doc/source/theory/Mellin.rst
giacomomagni: So this relation should be equivalent to eq 5.4d

giacomomagni:
I thought we discussed this in the past and we agreed that the factors `(-1)**N` couldn't appear explicitly inside the expressions of the ad/ome, but where only inside the harmonics...

doc/source/theory/Mellin.rst
giacomomagni: I think I actually used: [eq.510](https://arxiv.org/pdf/0901.3106.pdf)

niclaurenti (approved):

AleCandido:
>     * [ ]  Expand docs section about harmonics

I will review, but I would tick also the last item before merging. You were right in putting it in :)

benchmarks/performance/harmonics.py
```diff
@@ -0,0 +1,39 @@
...
+    def time_as2_sing(self):
+        for n in self.ns:
+            ad.gamma_singlet(1, n, NF)
+
+    def time_as3_sing(self):
+        for n in self.ns:
+            ad.gamma_singlet(2, n, NF)
+
+
+# class MemorySuite:
```
AleCandido, line 32: Remove the commented classes

AleCandido (commented):
Benchmarks and documentation are fine (as usual: we should stop using benchmarks in sandbox mode as soon as possible...)

tests/eko/test_matching_n3lo.py
```diff
@@ -173,47 +163,46 @@ def test_Blumlein_3(get_sx):
...
 
     # odd numbers of qqNS
     # Limited accuracy due to F functions
     ref_qqNS_odd = [-40.94998646588999, -21.598793547423504, 6.966325573931755]
-    for N, ref in zip([3.0, 15.0, 101.0], ref_qqNS_odd):
-        sx = get_sx(N)
-        smx = get_smx(N)
-        s3x = get_s3x(N, sx, smx)
-        s4x = get_s4x(N, sx, smx)
+    rtols = [4e-4, 3e-3, 21e-2]
```
AleCandido, line 189: The last accuracy is really bad...

AleCandido (commented):
Tests are mostly good (I'm just suggesting some aesthetic and readability), they even run in a reasonable amount of time (even though `ome` takes a bit, but that's fine).

tests/eko/test_matching_n3lo.py
giacomomagni: Indeed there was a dot missing so it's `2.1e-2`, not perfect but better than before...

doc/source/theory/Mellin.rst
felixhekhorn: I guess then we trust in your reference ...

giacomomagni: I added a test to check the consistency of these 2 formulas: 

https://github.com/N3PDF/eko/blob/9606618337c08df4e2118ce3c10be0862b1562c6/tests/eko/test_harmonics.py#L127

tests/eko/test_matching_n3lo.py
```diff
@@ -173,47 +163,46 @@ def test_Blumlein_3(get_sx):
             # as non singlet. The accuracy is worst for large N
             # due to the approximations of F functions.
             np.testing.assert_allclose(
-                aS3[1, 1], ref_val_qqNS[L][idx] + ref_val_qqPS[L][idx], rtol=3e-2
+                aS3[1, 1], ref_val_qqNS[L][idx] + ref_val_qqPS[L][idx], rtol=8e-2
```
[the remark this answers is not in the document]
AleCandido: Ok, I get the technical limitation. At this point I just wonder if we should drop the test.
However, we're always in time to do it.

PR #115, opened 2022-04-13
Feature/qed couplings

niclaurenti:


src/eko/couplings.py
```diff
@@ -60,66 +59,108 @@ def as_expanded(order, as_ref, nf, scale_from, scale_to):
...
-                        + 2
-                        * beta0
-                        * b1
-                        * log_fact
-                        * (b2 + 2 * (b1**2 - beta0 * b2) * lmu * as_ref)
-                        - beta0**2
-                        * lmu
-                        * as_ref
+    if order[1] == 0:
+        beta_qcd0 = beta_qcd(0, nf)
```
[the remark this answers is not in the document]
felixhekhorn: I think between the two sectors there will always be a symmetry, with just the references swapped and beta adjusted

src/eko/couplings.py
```diff
@@ -477,7 +477,7 @@ def compute(self, a_ref, nf, scale_from, scale_to):
         """
         key = (float(a_ref[0]), float(a_ref[1]), nf, scale_from, float(scale_to))
         try:
-            return self.cache[key]
+            return self.cache[key].copy()
```
AleCandido, line 480: At the moment the cache is just empty, so I'm pretty sure you want to fill it.

In case the content will be a structured object, pay attention to the difference between a deep and shallow copy. In any case, sometimes even a shallow copy is just fine, but you have to be aware and do it on purpose. If it's not structured, never mind.

src/eko/beta.py
```diff
@@ -46,21 +68,94 @@ def beta_1(nf):
...
+    nd = nf - nu
+    beta_aem_3 = (
+        -4.0 * constants.NC * (nu * constants.eu2**2 + nd * constants.ed2**2)
+    )
+    return beta_aem_3
 
 
 @nb.njit("f8(u1)", cache=True)
-def beta_2(nf):
+def beta_as_2aem1(nf):
```
felixhekhorn, line 109: Can we make the names more consistent? e.g. `beta_as2_aem1` (or all underscore?)

AleCandido: There is a reason for this name choice: this is the beta function for as, thus `beta_as`, while if it were for a it would have been called `beta_a`. After the second underscore, it follows the order as a series of numbers, each prefixed with the coupling, but the first one, for which it is implicit (explicitly it would be `beta_as_as2aem1`).
You can decide to put the order on the other side of the underscore, but it is not more consistent.

Think about the two cases with the same order:
- in your proposal, they would be `beta_as2_aem1` (the beta function for as) and `beta_aem1_as2` (the beta function for the qed coupling): this looks more like a mistake, why the two things are not the same?
- with current convention instead they are: `beta_as_2aem1` and `beta_a_1as2`, and if you truncate at the second underscore, you clearly read for what the beta function is, while the second piece is spelling out the order

It looks cleaner to me (even if from the point of view of the single one your own looks more natural), but I'm not especially keen in any. Both work, it's just a matter of taste at this point.

felixhekhorn: accepted - though the correct explicit name would be `beta_qcd_as2aem1`

AleCandido: I agree

src/eko/couplings.py
```diff
@@ -0,0 +1,729 @@
...
+                * (
+                    5 * b1**2 * log_fact
+                    + 3 * b1**2 * log_fact**2
+                    + beta0 * (b2 + 2 * (b1**2 - beta0 * b2) * lmu * ref)
+                )
+            )
+        )
+        return res
+
+    if order[1] == 0:
```
felixhekhorn, line 128: is there anything shared between the (if-) branches? if not make them a seperate function each

felixhekhorn:
I guess you can now attempt to merge `develop` here - an important change that happened meanwhile in `develop` that the `numba` usage has reverted to it's "original state", i.e. just-in-time compilation: for that all numba calls just need to be `nb.njit(cache=True)`

niclaurenti:
> I guess you can now attempt to merge `develop` here - an important change that happened meanwhile in `develop` that the `numba` usage has reverted to it's "original state", i.e. just-in-time compilation: for that all numba calls just need to be `nb.njit(cache=True)`

So you mean that I should change all the arguments of njit for example like:
@nb.njit("f8(f8,u1,u1,f8)", cache=True) -> @nb.njit(cache=True)

src/eko/couplings.py
```diff
@@ -265,96 +351,171 @@ def from_dict(cls, theory_card, masses=None):
...
-                if self.order >= 3:
-                    b_vec.append(beta_b(3, nf))
+                if self.order[0] >= 4:
+                    beta_qcd_vec.append(beta_qcd((5, 0), nf))
+        if self.order[1] >= 2:
+            beta_qcd_mix = beta_qcd((2, 1), nf)
+        beta_qed_vec = [beta_qed((0, 2), nf)]
+        beta_qed_mix = 0
+        # NLO
+        if self.order[1] >= 2:
```
giacomomagni, line 446: These two if are just the same as before, can you put them together?

src/eko/couplings.py
```diff
@@ -265,96 +351,171 @@ def from_dict(cls, theory_card, masses=None):
...
+            )
+        if self.order[0] in [0, 1]:
+            # return expanded solution for a_s and exact for a_em
+            a_s = couplings_expanded(
+                self.order, a_ref, nf, scale_from, float(scale_to)
+            )[0]
+            beta0_qed = beta_qed((0, 2), nf)
+            b_qed_vec = [1.0]
+            # NLO
+            if self.order[1] >= 2:  # I think that at this point this if is always true
```
giacomomagni, line 404: I agree, this should be true at this point

src/eko/couplings.py
```diff
@@ -265,96 +351,171 @@ def from_dict(cls, theory_card, masses=None):
...
+                rge,
+                (0, beta0 * u),
+                (a_ref,),
+                args=[b_vec],
+                method=method,
+                rtol=rtol,
+            )
+            return res.y[0][-1]
+
+        if self.order in [(0, 0), (0, 1), (1, 0), (1, 1)]:
```
giacomomagni, line 392: Is the case `(0,0)` meaningful?

giacomomagni (commented):
Hi @niclaurenti, 
I have added some minor comments. 

I noticed you haven't merged the most updated `develop` into this branch.
Just pull before merging here... (If you are using VS code there is this [extension](https://marketplace.visualstudio.com/items?itemName=mhutchie.git-graph) which is very useful)

Since this PR is quite big and introduced some changes in the notation, would you mind adding also some documentation and fix the names (`beta` ... ) also there? 
I think an explanation about the available pto expansions will be useful for our users.

src/eko/couplings.py
AleCandido: Nope, `(0,0)` just means there is no evolution, so in principle there is no reason to run `eko`.

Nevertheless, we might decide to gracefully fallback on some trivial values (identity for the EKO, flat for running couplings, and related).

src/eko/couplings.py
niclaurenti: No it's not. It's because previously the names where given factorizing a term as1 from beta_qcd and aem1 from beta_qed but then i reinserted it but i forgot to change the names inside the functions. I'm going to fix it

src/eko/couplings.py
niclaurenti: In the code in the case (0,0) the reference values of a_s,a_em are returned (no evolution).

src/eko/couplings.py
```diff
@@ -195,7 +195,7 @@ def couplings_expanded(order, couplings_ref, nf, scale_from, scale_to):
 
 class Couplings:
     r"""
-        Computes the strong and electromagnetic coupling constants :math:`a_s,a`.
+        Computes the strong and electromagnetic coupling constants :math:`a_s, a_em`.
```
felixhekhorn, line 322: Check the docstring also in practice! I'm quite sure `a_em` will be interpreted as `{a_e}m` which is not what you want ...

src/eko/couplings.py
```diff
@@ -36,102 +35,177 @@ def strong_coupling_mod_ev(mod_ev):
...
+@nb.njit("f8(f8,f8,f8)", cache=True)
+def expanded_lo(ref, beta0, lmu):
+    den = 1.0 + beta0 * ref * lmu
+    return ref / den
+
+
+@nb.njit("f8(f8,f8,f8,f8)", cache=True)
+def expanded_nlo(ref, beta0, beta1, lmu):
+    den = 1.0 + beta0 * ref * lmu
+    b1 = beta1 / beta0
```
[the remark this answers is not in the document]
niclaurenti: Maybe I dindn't understand what you mean. Here beta0, beta1, ecc are passed as parameters beacuse I wanted to use the same function for both QCD and QED since they have the same form but the beta functions are different

src/eko/couplings.py
```diff
@@ -37,12 +37,48 @@ def couplings_mod_ev(mod_ev):
 
 @nb.njit(cache=True)
 def expanded_lo(ref, beta0, lmu):
+    """Compute expanded solution at LO.
```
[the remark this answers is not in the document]
felixhekhorn: (of course here and below)

src/eko/couplings.py
```diff
@@ -37,12 +37,48 @@ def couplings_mod_ev(mod_ev):
...
+
+    Parameters
+    ----------
+        ref: float
+            reference value of the coupling
+        beta0: float
+            LO beta function
```
felixhekhorn, line 47: it's not the "beta function" but the first coefficient of the function (of course here and below)

src/eko/couplings.py
```diff
@@ -37,12 +37,48 @@ def couplings_mod_ev(mod_ev):
...
+        lmu: float
+            log(scale_to / scale_from)
```
felixhekhorn, line 49: A better name would be scaling log

src/eko/couplings.py
```diff
@@ -37,12 +37,48 @@ def couplings_mod_ev(mod_ev):
...
+
+    Returns
+    -------
+        a : float
+            coupling at target scale :math:`a(scale_to^2)`
```
felixhekhorn, line 54: please use a proper math name (like µ_r^2)

src/eko/couplings.py
```diff
@@ -62,6 +62,8 @@ def expanded_lo(ref, beta0, lmu):
 def expanded_nlo(ref, beta0, beta1, lmu):
     """Compute expanded solution at NLO.
 
+    From apfel
```
felixhekhorn, line 65: maybe write that this is the default expression found everywhere, e.g. in |APFEL| eq. ...

src/eko/couplings.py
```diff
@@ -90,6 +92,8 @@ def expanded_nlo(ref, beta0, beta1, lmu):
 def expanded_nnlo(ref, beta0, beta1, beta2, lmu):
     """Compute expanded solution at NNLO.
 
+    From apfel
```
felixhekhorn, line 95: please mention that this is *not* the default expression, but instead is coming from |APFEL| eq. ...

src/eko/couplings.py
```diff
@@ -123,6 +127,8 @@ def expanded_nnlo(ref, beta0, beta1, beta2, lmu):
 def expanded_n3lo(ref, beta0, beta1, beta2, beta3, lmu):
     """Compute expanded solution at N3LO.
 
+    From Luca Rottoli
```
felixhekhorn, line 130: - maybe we should add a bibtex entry á là "Luca Rottoli, private communication" and cite that one here
- also mention that is obtained via ordinary iterated solution of the RGE

src/eko/couplings.py
felixhekhorn: actually, why? `order` is coming from the outside ...

src/eko/gamma.py
```diff
@@ -10,7 +10,7 @@
 
 
 @nb.njit(cache=True)
-def gamma_0():
+def gamma_as1():
```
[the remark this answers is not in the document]
niclaurenti: I was thinking that maybe it's more clear if we drop this convention and we just rename beta_as_2aem1 -> beta_qcd_as2aem1. @AleCandido @giacomomagni what do you think?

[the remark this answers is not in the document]
felixhekhorn: fine with me - as said before this is actually the fully explicit name

src/eko/couplings.py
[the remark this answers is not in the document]
felixhekhorn: I guess he meant: take as argument `ref, beta0, b1, lmu` and compute the actual `b1` in the calling function - this way you can *reuse* `b1` instead of *recompute*

src/eko/couplings.py
```diff
@@ -46,12 +46,12 @@ def expanded_lo(ref, beta0, lmu):
...
             first coefficient of the beta function
         lmu: float
-            log(scale_to / scale_from)
+            scaling log
 
     Returns
     -------
         a : float
-            coupling at target scale :math:`a(scale_to^2)`
+            coupling at target scale :math:`a(µ_r^2)`
```
felixhekhorn, line 54: Please don't follow me blindly :see_no_evil: use your own brain :brain:  I meant this more in a "global" sense - follow the notation which is already there: e.g. https://eko.readthedocs.io/en/latest/theory/pQCD.html you see the default notation is `\mu_R`; same with the "scaling log" - you know what it is and if ever you find a better name or explanation (like ratio between reference scale and target scale or log expansion parameter or ...) use it - us your physics intuition :rocket: ;-)

src/eko/gamma.py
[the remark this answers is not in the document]
AleCandido: Essentially fine, but for `beta` actually `beta_as` it is even more explicit than `beta_qcd` (because it is the derivative of `as` in the first place), while `gamma_as` makes not so much sense, `gamma_qcd` is much more meaningful (since it's coming from the contribution of QCD interaction).

If you want to stick to only one of the two (because at the end the theory and the coupling are one-to-one), then prefer `qcd`, since `gamma_as` is way less meaningful.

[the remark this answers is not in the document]
giacomomagni: I would go for `gamma_qcd` as well.

src/eko/couplings.py
niclaurenti: Do you know where I can find a reference that I can cite?? In the Apfel paper and documentation i couldn't find anything

felixhekhorn: mmm you're right - that makes me worry that it seems there is official statement ... when we first came across this (meaning implementing NNLO) we asked to the whole collaboration how to obtain this, but we only got alternative solutions ... I can see two options: or we trust APFEL and just cite it (without specifying a specific equation), or we fall back to the ordinary, iterated solution

src/eko/couplings.py
```diff
@@ -0,0 +1,795 @@
...
+        "iterate-expanded",
+        "decompose-expanded",
+        "perturbative-expanded",
+    ]:
+        return "expanded"
+    raise ValueError(f"Unknown evolution mode {mod_ev}")
+
+
+@nb.njit(cache=True)
+def expanded_lo(ref, beta0, lmu):
```
felixhekhorn, line 39: Actually, this function should be called `exact_lo` as we can solve the RGE at LO exact and so we do always

src/eko/beta.py
```diff
@@ -25,17 +24,40 @@ def beta_0(nf):
...
+    beta_qcd_as2 = 11.0 / 3.0 * constants.CA - 4.0 / 3.0 * constants.TR * nf
+    return beta_qcd_as2
 
 
 @nb.njit(cache=True)
-def beta_1(nf):
+def beta_qed_aem2(nf):
+    """Computes the first coefficient of the QED beta function.
+
+    Implements Eq. (3.10) of :cite:`Herzog:2017ohr`.
```
felixhekhorn, line 39: Are you sure the Herzog paper discusses QED?

src/eko/couplings.py
AleCandido: In lack of better references, cite APFEL

felixhekhorn:
- @niclaurenti try to merge `develop` here (i.e. resolve the conflicts)
- please double check all LHA settings (including SV) are still passing
- then I guess we can merge ...
- just to remind everybody: this is a massive change both in code and interface and it will e.g. require a rewrite of all runcards (e.g. in pineko @andreab1997 )

niclaurenti:
> What is still missing here? @niclaurenti @felixhekhorn
> 
> Otherwise I would merge, and most likely release v0.10.0 (such that the moment we'll be able to adapt yadism for this as soon as we want to, but it is not going to happen automatically with `poetry update`).

It is complete. I thought that you wanted to review it again before merging.

felixhekhorn:
> Ok. Actually I put a function in compatibility.py that should interface the old runcards with the one that is used in this branch

Actually, this does not cover the shift of `PTO` -> `PTO+1` - maybe you should add this (and change all the benchmark cards again)

benchmarks/CT14_bench.py
```diff
@@ -62,15 +63,16 @@ def benchmark_llo_NF3(self, Q0=5, Q2grid=(100,)):
             "T15",
             "V15",
         ]
-        self.run([theory_card], [operator_card], ["CT14llo_NF3"])
+        new_theory, new_operators = compatibility.update(theory_card, operator_card)
```
felixhekhorn, line 66: this has to be lifted inside `ekomark` otherwise the others (LHA, etc.) won't benefit from that - more specific it should move just before the `eko` invocation (since the DB is unchanged)

src/eko/evolution_operator/__init__.py
```diff
@@ -286,10 +286,12 @@ def grid_size(self):
     def a_s(self):
         """Returns the computed values for :math:`a_s`"""
         sc = self.managers["strong_coupling"]
-        a0 = sc.a_s(
+        a0 = sc.a(
```
AleCandido, line 289: Maybe here and below I'd rename variables to `as0` and `as1`, such that it is clear at first inspection that we are still talking about the strong coupling.

src/eko/evolution_operator/grid.py
```diff
@@ -111,7 +111,7 @@ def from_dict(
                 created object
         """
         config = {}
-        config["order"] = int(theory_card["PTO"])
+        config["order"] = theory_card["order"]
```
AleCandido, line 114: Maybe make them explicitly integers as before (I don't remember if or which, but if `int` was there, there might have been a reason).
```suggestion
        config["order"] = tuple(int(o) for o in theory_card["order"])
```

src/eko/runner.py
```diff
@@ -7,11 +7,13 @@
 
 import numpy as np
 
+from eko import compatibility
```
AleCandido, line 10: Never use `eko` eplicitly inside `eko`. Use relative imports:
```suggestion
from . import compatibility
```

src/eko/evolution_operator/grid.py
```diff
@@ -34,8 +34,8 @@ class OperatorGrid:
             configuration dictionary
```
AleCandido, line 34: Even in this file, above, there is an absolute `import` from `eko`, please make it relative (as all the others)

AleCandido (changes requested):
I reviewed everything but the actual `couplings.py` file and `compatibility.py`, I'll do as soon as possible.

src/eko/couplings.py
```diff
@@ -0,0 +1,795 @@
...
+                    a_ref.astype(float), nf, scale_from, scale_to
+                )
+            else:
+                a_new = couplings_expanded(
+                    self.order, a_ref.astype(float), nf, scale_from, float(scale_to)
+                )
+            self.cache[key] = a_new.copy()
+            return a_new
+
+    def a(
```
felixhekhorn, line 625: I'd say let's provide some syntactical sugar: `def a_s(...): return a(...)[0]`

AleCandido:
I didn't review line by line the whole `couplings.py`, but what I read was fine.

If you can fix the last two comments, I believe you can merge right after.

src/eko/compatibility.py
```diff
@@ -0,0 +1,38 @@
...
+            upgraded observable runcard
+    """
+    new_theory = copy.deepcopy(theory)
+    new_operators = copy.deepcopy(operators)
+    if "alphaqed" in new_theory:
+        new_theory["alphaem"] = new_theory.pop("alphaqed")
+    if "QED" in new_theory:
+        new_theory["order"] = (new_theory.pop("PTO") + 1, new_theory.pop("QED"))
+    else:
+        if "PTO" in new_theory:
```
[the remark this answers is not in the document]
felixhekhorn: mh? but all old cards have a QED field - it was in APFEL in the first place

src/eko/compatibility.py
```diff
@@ -0,0 +1,38 @@
...
+            new_theory["order"] = (new_theory.pop("PTO") + 1, 0)
+    if "ev_op_max_order" in new_operators:
```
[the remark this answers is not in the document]
niclaurenti: Here I had put it since in the tests we don't always provide the operator card so I hat to define an empty card and pass it. But now it should be fixed

src/eko/compatibility.py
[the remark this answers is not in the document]
niclaurenti: Yes yes, but for example some tests have not it and kept failing. This is why I implemented it that way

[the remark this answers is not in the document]
AleCandido: "PTO" is always there, this is for free :)

src/eko/couplings.py
```diff
@@ -679,6 +679,9 @@ def a(
             final_a = new_a
         return final_a
 
+    def a_s(self, scale_to, fact_scale=None, nf_to=None):
```
felixhekhorn, line 682: You can copy the docstring

niclaurenti:
@felixhekhorn @AleCandido Now everything should be fixed. Can we merge?

src/eko/compatibility.py
```diff
@@ -0,0 +1,36 @@
...
+        new_obs : dict
+            upgraded observable runcard
+    """
+    new_theory = copy.deepcopy(theory)
+    new_operators = copy.deepcopy(operators)
+    if "alphaqed" in new_theory:
+        new_theory["alphaem"] = new_theory.pop("alphaqed")
+    if "QED" in new_theory:
+        new_theory["order"] = (new_theory.pop("PTO") + 1, new_theory.pop("QED"))
+    if operators is not None:
```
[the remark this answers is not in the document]
AleCandido: The best would be to have two decoupled methods (one for theory, the other for operators) that is not possible at the moment.

Most likely the best way would be to lift `ev_op_max_order` on theory side (and actually make it part of the `ModEv`). At that point you won't need any longer.

But ok, let's keep like this for the time being, and do it in a second step. What do you think @felixhekhorn?

src/eko/compatibility.py
```diff
@@ -0,0 +1,36 @@
...
+        if isinstance(new_operators["ev_op_max_order"], int):
+            new_operators["ev_op_max_order"] = (
+                new_operators["ev_op_max_order"],
+                new_theory["order"][1],
+            )
+        return new_theory, new_operators
+    return new_theory
```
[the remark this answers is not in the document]
AleCandido: Something like:

```py
def update_theory(theory):
    return update(theory, None)[0]
```

AleCandido (changes requested):

src/eko/compatibility.py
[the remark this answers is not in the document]
felixhekhorn: > But ok, let's keep like this for the time being, and do it in a second step. What do you think @felixhekhorn?

definitely, this PR is already too big ...

AleCandido (approved):

felixhekhorn:
fine to merge!

PR #123, opened 2022-05-27
Generate callables for each PTO

### LINKED ISSUE #119: Matching failed to compile
Running `eko` resulted in
```
2022-05-19 17:58:25,999 eko.evolution_operator/INFO: Evolution: computing operators 2.722500e+00 -> 1.000000e+01, nf=4
2022-05-19 17:58:25,999 eko.evolution_operator/INFO: Evolution: µ_R^2 distance: 6.806250e-01 -> 2.500000e+00
2022-05-19 17:58:25,999 eko.evolution_operator/INFO: Scale Variation: (µ_F/µ_R)^2 = 4.000000e+00, mode: expanded
2022-05-19 17:58:27,223 eko.evolution_operator/INFO: Evolution: a_s distance: 2.494975e-02 -> 1.910111e-02
2022-05-19 17:58:27,223 eko.evolution_operator/INFO: Evolution: order: 1, solution strategy: truncated
2022-05-19 18:00:52,979 eko.evolution_operator/INFO: Evolution: computing operators: - 9/50 took: 145.698243 s
2022-05-19 18:00:53,748 eko.evolution_operator/INFO: Evolution: computing operators: - 15/50 took: 146.446427 s
2022-05-19 18:00:54,008 eko.evolution_operator/INFO: Evolution: computing operators: - 5/50 took: 146.740913 s
2022-05-19 18:00:54,355 eko.evolution_operator/INFO: Evolution: computing operators: - 3/50 took: 147.093093 s
2022-05-19 18:00:54,446 eko.evolution_operator/INFO: Evolution: computing operators: - 13/50 took: 147.152215 s
2022-05-19 18:00:54,620 eko.evolution_operator/INFO: Evolution: computing operators: - 1/50 took: 147.362054 s
2022-05-19 18:00:54,905 eko.evolution_operator/INFO: Evolution: computing operators: - 11/50 took: 147.619450 s
2022-05-19 18:00:56,091 eko.evolution_operator/INFO: Evolution: computing operators: - 7/50 took: 148.816877 s
2022-05-19 18:00:56,770 eko.evolution_operator/INFO: Evolution: computing operators: - 10/50 took: 3.791091 s
2022-05-19 18:00:57,047 eko.evolution_operator/INFO: Evolution: computing operators: - 16/50 took: 3.298314 s
2022-05-19 18:00:58,106 eko.evolution_operator/INFO: Evolution: computing operators: - 14/50 took: 3.660176 s
2022-05-19 18:00:58,106 eko.evolution_operator/INFO: Evolution: computing operators: - 6/50 took: 4.098238 s
...
2022-05-19 18:01:07,163 eko.evolution_operator/INFO: Evolution: Total time 159.940010 s
```
which is fine - because he needs to compile on each core after that its just normal; after that it is quick:
```
2022-05-19 18:01:07,234 eko.evolution_operator/INFO: Evolution: computing operators 2.722500e+00 -> 2.000000e+01, nf=4
...
2022-05-19 18:01:25,740 eko.evolution_operator/INFO: Evolution: Total time 18.505267 s
```
but then I reached the threshold and I get
```
2022-05-19 18:01:25,809 eko.evolution_operator/INFO: Evolution: computing operators 2.722500e+00 -> 2.420640e+01, nf=4
...
2022-05-19 18:01:44,208 eko.evolution_operator/INFO: Evolution: Total time 18.397738 s
```
and after that nothing is happening for 45min (but the CPU running at full power) when I decided to kill the process

I haven't investigated yet - @giacomomagni @AleCandido any ideas?

felixhekhorn:
Closes #119 

- in my first attempt I only checked tests are actually running
- it still has some repetitions which maybe can be avoided
- we can *not* cache the "dispatcher" - so the functions we generate dynamically since `nb` wants a definitive address (which we haven't since we generate on the fly); however, this is maybe acceptable since they are short and the hard work (the actual elements) are fully cached

we should assign this some priority to make the package usable again from the outside


felixhekhorn:
Actually, I think
https://github.com/N3PDF/eko/blob/143fe85e8ed3678da043e04694cbfe6df405aa8e/src/eko/matching_conditions/operator_matrix_element.py#L104-L107
the if condition could even be removed since the new compilation guard fulfils the same purpose ...

felixhekhorn:
> * we can _not_ cache the "dispatcher" - so the functions we generate dynamically since `nb` wants a definitive address (which we haven't since we generate on the fly); however, this is maybe acceptable since they are short and the hard work (the actual elements) are fully cached

unfortunately in practice this is visible: for the first step in matching we pay 20s - compared to the ~7-8s for the second step; but for the moment I don't see a way around ...

felixhekhorn:
@AleCandido this is my second attempt; I'm not too happy, because there is too much copy and paste still in the game ... to be specific: these 11 lines https://github.com/N3PDF/eko/blob/61068792c2c53cc9d719dfd054fde4d769af14f8/src/eko/matching_conditions/operator_matrix_element.py#L119-L135 are copied almost exact 3 times ...

I guess, I should start introducing function pointers ... what do you think?

AleCandido:
Wrap the three assignments into a single function:
https://github.com/N3PDF/eko/blob/61068792c2c53cc9d719dfd054fde4d769af14f8/src/eko/matching_conditions/operator_matrix_element.py#L119-L124
the early return can be just propagated (you early return in the function inside, and then check again in the one outside).

Also these two lines can be packed together in a single function:
https://github.com/N3PDF/eko/blob/61068792c2c53cc9d719dfd054fde4d769af14f8/src/eko/matching_conditions/operator_matrix_element.py#L133-L135

You know, they are small, so you might wonder if it's worth to bundle them into a function. I tell you that is worth: if you need to change, a single place will be affected.

In this way you go down to 8 lines. And I know they are repeated, but code is not only measured in lines: they contain almost nothing, the differences are relevant.
It's true that in principle you can get most of it in a single line, passing around the function pointers, but there is an extra value in simplicity: function pointers are complicated objects, then you might get closer to signatures, and we are screwed again.
In any case, you'd spend your time to make a more complicated call that would span 3 lines in any case, so just to remove 5 lines.

Factorize what is simple to gather, keep explicit everything else. We are always in time to improve.

src/eko/quad_ker_base.py
```diff
@@ -0,0 +1,143 @@
...
+        # OME mode
+        if self.ker_mode == MODE_MATCHING:
+            if self.is_singlet:
+                indices = {21: 0, 100: 1, 90: 2}
+            else:
+                indices = {200: 0, 91: 1}
+            return ker[indices[self.mode0], indices[self.mode1]]
+        # AD mode
+        k = 0 if self.mode0 == 100 else 1
+        l = 0 if self.mode1 == 100 else 1
```
AleCandido, line 97: I had a look at code factor, and most of the remaining issues are with poor names.

On one side, the immediate feeling is that it is too picky, and for sure it is not able to recognize long living variables from local indices and so on. But on the other, it is only (or mostly) complaining about one letter variables, that we might really consider too little, since it is for sure not meaningful at all (almost anonymous).

Here, e.g., I would make the minimal effort to improve, and just call the two of them `fl1` and `fl2` (or `fli` and `flo`, for I/O).

giacomomagni:
Should we close this?

AleCandido:
> Should we close this?

At some point, definitely. But @felixhekhorn mentioned that he wanted to resume the project, when possible.
We can close and restart from scratch later on, or wait for the replacement to close.

I'm fine with both.

felixhekhorn:
> At some point, definitely. But @felixhekhorn mentioned that he wanted to resume the project, when possible.

yes, with more and more math coming in (QED, N3LO, polarized, FF), i.e. more and more complicated compilation, I consider this will become relevant again.
 
> We can close and restart from scratch later on, or wait for the replacement to close. I'm fine with both.

since a **lot** of things changed meanwhile, I guess we can convert the idea into an issue (mentioning the alternative proposal of using C/Rust) and close this PR ...

AleCandido:
This is outdated, and by now *on hold*.

Since we might want to step out Numba, maybe it will also be solved in a different way.

felixhekhorn:
Shifted as issue in #204

Issue #121, opened 2022-05-24
problems with scheme B scale variations


felixhekhorn:
in pineko we have a problem with scheme B scale variations https://github.com/N3PDF/pineko/pull/19 (we should move pineko to NNPDF to make @giacomomagni able to see it ...)

- our first guess was to remove the scale variations of the strong coupling to "no correction is applied here to the "resummation" scale," [as said by @AleCandido] - which we didn't do before (see [here](https://github.com/N3PDF/eko/blob/341e6323e680d2320ae4400da61c92a445160416/src/eko/evolution_operator/__init__.py#L278) ) and indeed we know that for scheme A = LHA paper shifting was correct (since we're matching there); this fix turns out to be not sufficient
- out of fun I tried to *invert* the scale ratio (xiev -> 1/xiev) and surprisingly enough this brought scheme B and C much closer: from 90% off to mostly O(10%) off
- then I started looking at the scheme B implementation of eko and there is something I don't understand: why is the [expanded correction](https://github.com/N3PDF/eko/blob/341e6323e680d2320ae4400da61c92a445160416/src/eko/scale_variations/expanded.py#L30) not similar to the [exponentiated correction](https://github.com/N3PDF/eko/blob/341e6323e680d2320ae4400da61c92a445160416/src/eko/scale_variations/exponentiated.py#L51) ? i.e. one has beta0 the other not?
- however, looking at Eq. 3.33 of MHOU I can't see an immediate error ...

PS: side note: https://github.com/N3PDF/eko/blob/341e6323e680d2320ae4400da61c92a445160416/doc/source/theory/pQCD.rst#L126
there is no such thing as a PDF set in eko - the "whole operator" is meant

felixhekhorn:
- [ ] most likely a missing minus sign from Eq. 3.33
- [ ] in scheme B the starting scale has to be kept fixed (so no *effective* shifting - in practice we might do and undo the thing); in scheme A the current implementation is correct
- [ ] `pineko` need to request for scheme B the *central* scale (and not the shifted one as at the moment of writing is doing); effectively eko is takining fully care of the SV

AleCandido:
> (we should move pineko to NNPDF to make @giacomomagni able to see it ...)

What if we made `pineko` public to begin with?

giacomomagni:
It's already public! I'm able to see it :)

AleCandido:
> (we should move pineko to NNPDF to make @giacomomagni able to see it ...)

[...]

PR #124, opened 2022-05-31
Exclude N3LO matching from compilation

felixhekhorn:
in order to prioritize #121 over #119 I suggest to do this quick fix here now and then later do the proper job in #123

### DIFF
--- a/src/eko/matching_conditions/operator_matrix_element.py
+++ b/src/eko/matching_conditions/operator_matrix_element.py
@@ -57,10 +57,10 @@ def compute_harmonics_cache(n, order, is_singlet):
     if order == 2:
         # Add Sm21 to cache
         sx[2, 1] = harmonics.Sm21(n, sx[0, 0], sx[0, -1], is_singlet)
-    if order == 3:
-        # Add weight 3 and 4 to cache
-        sx[2, 1:-2] = harmonics.s3x(n, sx[:, 0], sx[:, -1], is_singlet)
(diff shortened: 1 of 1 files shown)


felixhekhorn:
We have to sacrifice coverage, but I'll promise to recover in #123 - let's close this and make a new release!

src/eko/matching_conditions/operator_matrix_element.py
```diff
@@ -46,19 +48,22 @@ def compute_harmonics_cache(n, order, is_singlet):
 
     """
     # max harmonics sum weight for each qcd order
-    max_weight = {1: 2, 2: 3, 3: 5}
+    max_weight = {1: 2, 2: 3, 3: 3}
+    # _N3LO_ max_weight = {1: 2, 2: 3, 3: 5}
     # max number of harmonics sum of a given weight for each qcd order
-    n_max_sums_weight = {1: 1, 2: 3, 3: 7}
+    n_max_sums_weight = {1: 1, 2: 3, 3: 3}
```
AleCandido, line 54: Actually, why you changed the numbers?

AleCandido: Ok, but I believe here `3` is actually N3LO, so you would never use in any case.

felixhekhorn: True, but this would require further adjusting of the tests

AleCandido (approved):

PR #125, opened 2022-05-31
Fix scale variations scheme B

### LINKED ISSUE #121: problems with scheme B scale variations

felixhekhorn:
closes #121

- [x] ~~most likely a missing minus sign from Eq. 3.33~~
- [x] in scheme B the starting scale has to be kept fixed (so no *effective* shifting - in practice we might do and undo the thing); in scheme A the current implementation is correct

_Originally posted by @felixhekhorn in https://github.com/N3PDF/eko/issues/121#issuecomment-1137310300_


felixhekhorn:
@giacomomagni you correctly said here: https://github.com/N3PDF/eko/blob/039beec06aec5f14be82b696817bd6b5b4ed2b6f/src/eko/scale_variations/expanded.py#L170
but then why didn't you reflect that? do you agree on my fix in https://github.com/N3PDF/eko/pull/125/commits/039beec06aec5f14be82b696817bd6b5b4ed2b6f ?

the unit test is now failing ... I'm not entirely sure on what it should test ... also the non-commutativity in scheme B should now be taken care of ... but actually it seems the discrepency is worse now ... can you please have a look?

giacomomagni:
Hi @felixhekhorn, 
I'm not really convinced about the motivation of the missing minus sign. 

If I look at eq. 3.32 (scale variation exponentiated) you can see that the sign convention holds. 
This equations is exactly the same as eq 2.8 of the Pegasus paper and is what we are doing in the code.
However the only term that contains an odd number of beta*gamma occurs at NNLO, so it's still possible that we are missing a sign there, but this would imply that the LHA bench are not enough a good check of our implementation...  

While doing the work I annotated that there is also a minus sign in the definition of `k` eq 3.5.

AleCandido:
> @andreab1997 could you please double check my updated version - @giacomomagni convinced me that he was right in the first place ...

I'd recommend implementing the second point (about the starting scale), and once it is in place, check that scheme B and C are actually sufficiently compatible (or that expanded and exponentiated are compatible, using TRN)

src/eko/evolution_operator/__init__.py
```diff
@@ -195,9 +195,9 @@ def quad_ker(
         )
         # scale var expanded is applied on the kernel
         if sv_mode == sv.Modes.expanded and not is_threshold:
-            ker = np.ascontiguousarray(ker) @ np.ascontiguousarray(
+            ker = np.ascontiguousarray(
                 sv.expanded.singlet_variation(gamma_singlet, a1, order, nf, L)
-            )
+            ) @ np.ascontiguousarray(ker)
```
giacomomagni, line 200: uhh well spotted! Did this solved the issue?

felixhekhorn: Unfortunately no ... still I believe this error was mathematically at the same order order as the scale difference, just in practice with the actual numbers it is tiny

AleCandido:
@felixhekhorn do you have a candidate now, or still investigating?

andreab1997:
> @felixhekhorn do you have a candidate now, or still investigating?

Last time we talked, we noticed again that a change of a minus sign would lead to compatible results with scheme C. However, we checked together basically all the part of the code related to scale variations and we did not find where this minus sign could be wrong. I don't know if @felixhekhorn has news about this.

felixhekhorn:
Let's discuss this later ... (I **really** need to prepare the meeting later **now** :see_no_evil: :scream:  )

felixhekhorn:
Since we're about to close a bunch of PRs, let's close also this one (despite that there is no definite conclusion on scheme B)

Changes should be:
- many docstrings
- introduce `ModeMixin`
- fix non-commutativity in singlet SV
- put SV kernel to the very left
- introduce `mur2_shift`

this PR does not include the split of the SV kernel from the evolution kernel, but I suggest to postpone this to a separate PR

AleCandido:
@giacomomagni may I ask you to review it? This week I'll be pretty busy...

giacomomagni (commented):
okay, looks good. When also the tests are updated we can merge this.

felixhekhorn:
> okay, looks good. When also the tests are updated we can merge this.

@giacomomagni you're absolutely correct and I should apologize to have asked for a review before the thing was actually finished ... as you can see the merge has been actually non-trivial;

please give me an explicit approval when you're happy (note that (as usual) we get the green tick only after the benchmarks finished)

giacomomagni (approved):
Sure no problem. Thanks for taking care of updating this.

Issue #130, opened 2022-06-15
Evolve pdfs with many replicas


giacomomagni:
When using `evolve_pdfs` with more than one replica, all the evolved replicas are the same.
Ploblably there is a mistake in how the blocks are appended: see 

https://github.com/N3PDF/eko/blob/16b4606d877099ebf7b806f225a2f35adf881049/src/ekobox/evol_pdf.py#L94

This is the script I'm using to evolve the pdf set:
````
import numpy as np
import pathlib

from eko.interpolation import make_lambert_grid
from banana.benchmark.runner import get_pdf


from ekobox.evol_pdf import evolve_pdfs

here = pathlib.Path(__file__).parent

th_card = {
    "HQ": "POLE",
    "IB": 0,
    "IC": 1,
    "MaxNfAs": 6,
    "MaxNfPdf": 6,
    "ModEv": "EXA",
    "ModSV": None,
    "NfFF": 4,
    "PTO": 2,
    "Q0": 1.65,
    "Qmb": 4.92,
    "Qmc": 1.51,
    "Qmt": 172.5,
    "Qref": 9.1187600e01,
    "alphas": 0.1180,
    "fact_to_ren_scale_ratio": 1.0,
    "kbThr": 1.0,
    "kcThr": 1.0,
    "ktThr": 1.0,
    "mb": 4.92,
    "mc": 1.51,
    "mt": 172.5,
    "nf0": 4,
    "nfref": 5,
    "MZ": 9.1187600e01,
}

op_card = {
    "Q2grid": [10,100,1e3,1e4,1e5],
    "backward_inversion": "exact",
    "debug_skip_non_singlet": False,
    "debug_skip_singlet": False,
    "ev_op_iterations": 10,
    "ev_op_max_order": 10,
    "interpolation_is_log": True,
    "interpolation_polynomial_degree": 4,
    "interpolation_xgrid": make_lambert_grid(60).tolist(),
    "n_integration_cores": 8,
}

pdf_name = "NNPDF40_nnlo_as_01180_nf_4"


def run_ffns4(pto):
    """Evolve a pdf in FFNS 4"""
    th_card["kbThr"] = np.inf
    th_card["ktThr"] = np.inf
    th_card["nfref"] = 4
    th_card["PTO"] = pto

    cached_path = here.joinpath(f"ffns4_pto_{pto}.tar")
    pdf = get_pdf(pdf_name, full_set=True)
    evolve_pdfs(
        pdf,
        th_card,
[...]

PR #131, opened 2022-06-15
Fixing bug of ekobox evolve_pdfs with multiple replicas

andreab1997:
Solve #130 
The bug was due to the fact that, even if `all_blocks` was supposed to contain all and only the blocks of a certain replica, it was not initialized to zero at the beginning of each loop step on `evolved_PDFs`.

### DIFF
--- a/benchmarks/ekobox/benchmark_evol_pdf.py
+++ b/benchmarks/ekobox/benchmark_evol_pdf.py
@@ -91,8 +91,14 @@ def benchmark_evolve_more_members(tmp_path, cd, lhapdf_path):
             ev_p.evolve_pdfs(pdfs, theory, op, install=True, name="Debug")
         # ev_pdfs
         _ = lhapdf.mkPDFs("Debug")
+        new_pdf_1 = lhapdf.mkPDF("Debug", 0)
+        new_pdf_2 = lhapdf.mkPDF("Debug", 1)
         info = load.load_info_from_file("Debug")
     assert info["XMin"] == op["interpolation_xgrid"][0]
+    for Q2 in [10, 100]:
+        for x in [1e-7, 0.01, 0.1, 0.2, 0.3]:
+            for pid in [21, 1, 2]:
+                assert new_pdf_1.xfxQ2(pid, x, Q2) != new_pdf_2.xfxQ2(pid, x, Q2)
 
 
 @pytest.mark.isolated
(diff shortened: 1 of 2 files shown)


felixhekhorn:
Closes #130 

Github has some [keywords](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests) to automatically close an associated issue

andreab1997:
I believe that now the problem should be solved. Thanks to @giacomomagni to have spotted this. :)

benchmarks/ekobox/benchmark_evol_pdf.py
```diff
@@ -91,8 +91,14 @@ def benchmark_evolve_more_members(tmp_path, cd, lhapdf_path):
             ev_p.evolve_pdfs(pdfs, theory, op, install=True, name="Debug")
         # ev_pdfs
         _ = lhapdf.mkPDFs("Debug")
```
AleCandido, line 93: Maybe now you don't need any longer this line

AleCandido, line 93: Maybe now you don't need any longer this line

andreab1997: I would keep it because it is testing if lhapdf is able to load the complete set

AleCandido: Is it doing something different than loading the info and the individual replicas?

AleCandido: If so, don't discard the return value, but test what you expect it to be.

andreab1997: No it is only loading the info and all the replicas and constructing the lhapdf object. Maybe I can just check if the number of final replicas is the same of the initial number

giacomomagni:
Okay thanks for fixing it! Can I add also the feature to store the computed eko ?
Right now evolve_pdfs loads the eko but has not the option to store it...

benchmarks/ekobox/benchmark_evol_pdf.py
andreab1997: Have a look if now it is ok

AleCandido (approved):

felixhekhorn (approved):

felixhekhorn (approved):

PR #135, opened 2022-07-05
Implement QED exact solution

niclaurenti:
- [x] if `math.isnan(couplings.alphaem.scale)` then deactivate the $\alpha$ running
    - to be traded for the current `alphaem_running` flag


src/eko/anomalous_dimensions/as1aem1.py
```diff
@@ -355,3 +356,63 @@ def gamma_nsm(N, sx):
...
+    nd = nf - nu
+    vu = nu / nf
+    vd = nd / nf
+    e2avg = (nu * constants.eu2 + nd * constants.ed2) / nf
+    e2m = constants.eu2 - constants.ed2
+    e2delta = (nd * constants.eu2 + nu * constants.ed2) / nf
+    gamma_S_11 = np.array(
+        [
+            [
+                nf * e2avg * gamma_gg(),
```
felixhekhorn, line 373: Please start to factorize right away: find out the most efficient number of independent variables and immediately compute the in a separate function; e.g. I can see
```
abra = vu * e2m
kadabra = vd * e2m
```
is repeated here and below - and it is the atomic ingredient; so only comute them once (you can put, e.g., next to the `uplike_flavors` function) (remember you can return a tuple)

PS: I hope you can find better names then the ones I suggested :see_no_evil:

src/eko/anomalous_dimensions/as1.py
```diff
@@ -140,3 +140,48 @@ def gamma_singlet(N, s1, nf):
         [[gamma_qq, gamma_qg(N, nf)], [gamma_gq(N), gamma_gg(N, s1, nf)]], np.complex_
     )
     return gamma_S_0
+
+
+@nb.njit(cache=True)
+def gamma_QEDsinglet(N, s1, nf):
+    r"""
```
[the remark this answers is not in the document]
AleCandido: I'm not strict about existing ones, we decided to go in that direction incrementally, but I am on new additions :wink:

src/eko/anomalous_dimensions/as1.py
```diff
@@ -140,3 +140,48 @@ def gamma_singlet(N, s1, nf):
         [[gamma_qq, gamma_qg(N, nf)], [gamma_gq(N), gamma_gg(N, s1, nf)]], np.complex_
     )
     return gamma_S_0
+
+
+@nb.njit(cache=True)
+def gamma_QEDsinglet(N, s1, nf):
```
AleCandido, line 146: I'm not sure this is the best name ever, but as @felixhekhorn keeps saying, "we're not good at naming".

In any case, let's brainstorm a bit about it :)

niclaurenti: Yes I agree. I didn't think much about it but we can always change it as we find a good name

src/eko/anomalous_dimensions/as1aem1.py
felixhekhorn: - how about calling `nfe2avg` just `e2tot`?
- and then of course the important part is to split the code from here and put in an extra function that you reuse, e.g. in `gamma_valence`

niclaurenti:
@felixhekhorn @AleCandido Question about names: in the nonsinglet evolution for the pure QCD we had the terms j00, j01 etc where the j functions are the integrals of as^k/beta_N^iLO(as). In that case k starts from 1.
However in the QED case we have also the case k=0. Should we rename the already implemented j_ki by shifting by 1 the first index (i.e. j00 -> j10, j01 -> j11, etc) such that k=0 is the integral of 1/beta(as)? Or maybe we should find another name?

src/eko/basis_rotation.py
```diff
@@ -61,6 +61,22 @@
 corresponding |PDF| : :math:`\gamma, \Sigma, g, V, V_{3}, V_{8}, V_{15}, V_{24},
 V_{35}, T_{3}, T_{8}, T_{15}, T_{24}, T_{35}`
 """
+intrinsic_evol_basis = (
```
felixhekhorn, line 64:
```suggestion
unified_evol_basis = (
```

src/eko/basis_rotation.py
```diff
@@ -175,19 +191,19 @@
...
+    (22, 21): ["ph.g"],
+    (22, 22): ["ph.ph"],
+    (22, 100): ["ph.S"],
+    (22, 101): ["ph.Sdelta"],
     (100, 21): ["S.g"],
-    (100, 20): ["S.ph"],
+    (100, 22): ["S.ph"],
     (100, 100): ["S.S"],
     (100, 101): ["S.Sdelta"],
     (101, 21): ["Sdelta.Sdelta"],
```
felixhekhorn, line 205:
```suggestion
    (101, 21): ["Sdelta.g"],
```

src/eko/evolution_operator/physical.py
```diff
@@ -75,6 +75,76 @@ def ad_to_evol_map(cls, op_members, nf, q2_final, intrinsic_range):
...
+            "ph.S": op_members[(22, 100)],
+            "ph.Sdelta": op_members[(22, 101)],
+            "S.g": op_members[(100, 21)],
+            "S.ph": op_members[(100, 22)],
+            "S.S": op_members[(100, 100)],
+            "S.Sdelta": op_members[(100, 101)],
+            "V.V": op_members[(10200, 10200)],
+            "V.Vdelta": op_members[(10200, 10204)],
+            "Vdelta.V": op_members[(10204, 10200)],
+            "Vdelta.V": op_members[(10204, 10204)],
```
felixhekhorn, line 117:
```suggestion
            "Vdelta.Vdelta": op_members[(10204, 10204)],
```

src/eko/evolution_operator/physical.py
```diff
@@ -75,6 +75,76 @@ def ad_to_evol_map(cls, op_members, nf, q2_final, intrinsic_range):
...
+            "g.Sdelta": op_members[(21, 101)],
+            "ph.g": op_members[(22, 21)],
+            "ph.ph": op_members[(22, 22)],
+            "ph.S": op_members[(22, 100)],
+            "ph.Sdelta": op_members[(22, 101)],
+            "S.g": op_members[(100, 21)],
+            "S.ph": op_members[(100, 22)],
+            "S.S": op_members[(100, 100)],
+            "S.Sdelta": op_members[(100, 101)],
+            "V.V": op_members[(10200, 10200)],
```
felixhekhorn, line 114: Choose your convention: either NS stuff as 0 as second argument or you repeat the first label

niclaurenti: For V.V i didn't use the notation (pid, 0) because it belongs to the 2x2 sector (even if it is a nonsinglet). In fact we have also V.Vdelta and Vdelta.Vdelta

tests/eko/test_ad.py
```diff
@@ -35,6 +35,23 @@ def test_eigensystem_gamma_singlet_0_values():
     assert_allclose(e_m, res[4])
 
 
+def test_exp_matrix():
```
felixhekhorn, line 38: also add a (or several) test(s) with known mathematical properties such as identity or rotation or 0s etc. ...

src/eko/evolution_operator/physical.py
felixhekhorn: you're right

tests/eko/test_ad.py
```diff
@@ -50,6 +50,11 @@ def test_exp_matrix():
     res = expm(gamma_v_0_qed)
     res2 = ad.exp_matrix(gamma_v_0_qed)[0]
     assert_allclose(res, res2)
+    diag = np.diag([1, 2, 3, 4])
+    assert_allclose(np.diag(np.exp([1, 2, 3, 4])), ad.exp_matrix(diag)[0])
+    Id = np.identity(4, np.complex_)
```
felixhekhorn, line 55: stick to the python naming convention `Id` -> `id_` (since the name `id` is already taken by a built-in function)

AleCandido: You see that otherwise even the syntax highlighter is confused :)

felixhekhorn:
@niclaurenti since your touching solutions here and #137 was doing as well, maybe make sure by merging that your fine with the changes there ...

src/eko/evolution_operator/physical.py
```diff
@@ -47,96 +47,75 @@ def ad_to_evol_map(cls, op_members, nf, q2_final, intrinsic_range):
...
         if intrinsic_range is not None:
             hqfl = "cbt"
-            op_id = member.OpMember.id_like(
-                op_members[
-                    (br.non_singlet_pids_map["nsV"], 0)
-                ]  # TODO : Probably it is wrong
-            )
+            if not is_qed:
+                op_id = member.OpMember.id_like(
+                    op_members[(br.non_singlet_pids_map["nsV"], 0)]
```
felixhekhorn, line 113: just use, e.g., the `g.g` which is always there

src/eko/evolution_operator/physical.py
```diff
@@ -47,96 +47,75 @@ def ad_to_evol_map(cls, op_members, nf, q2_final, intrinsic_range):
...
-            m["Vu3.Vu3"] = op_members[(br.non_singlet_pids_map["ns-u"], 0)]
-        if nf >= 5:
-            m["Td8.Td8"] = op_members[(br.non_singlet_pids_map["ns+d"], 0)]
-            m["Vd8.Vd8"] = op_members[(br.non_singlet_pids_map["ns-d"], 0)]
-        if nf >= 6:
-            m["Tu8.Tu8"] = op_members[(br.non_singlet_pids_map["ns+u"], 0)]
-            m["Vu8.Vu8"] = op_members[(br.non_singlet_pids_map["ns-u"], 0)]
+        if not is_qed:
+            m = {
+                "S.S": op_members[(100, 100)],
```
felixhekhorn, line 52: the four singlet lines are actually shared between the two branches

src/eko/evolution_operator/physical.py
felixhekhorn: the idea was to drop the if condition :upside_down_face: - since, as said, it is always there and here you need just a reference operator to get the size

src/eko/evolution_operator/flavors.py
```diff
@@ -130,7 +131,7 @@ def rotate_pm_to_flavor(label):
     return l
 
 
-def rotate_matching(nf, inverse=False):
+def rotate_matching(nf, is_qed=False, inverse=False):
```
AleCandido, line 134: I would also go for the short version, just `qed`. It is a flag, so it should be non-ambiguous :)

benchmarks/CT18_bench.py
```diff
@@ -46,13 +46,13 @@ def benchmark_nnlo(self, Q0=1.295, Q2grid=(1e4,)):
...
-            "T35",
-            "V35",
+            # 6,
+            # 22,
+            # "ph",
+            "Tu8",
+            "Vu8",
         ]
-        self.run([theory_card], [operator_card], ["CT18NNLO"])
+        self.run([theory_card], [operator_card], ["NNPDF31_nnlo_as_0118_luxqed"])
```
felixhekhorn, line 55: Keep the CT18 benchmark for CT18 - we also have the NNPDF benchmark with the default settings used there available ...

niclaurenti: Yes sorry I had changed it for running a bench but then I forgot to change it and I pushed it

src/eko/couplings.py
```diff
@@ -206,23 +206,29 @@ def expanded_qcd(ref, order, nf, lmu):
         strong coupling at target scale :math:`a_s(\mu_R^2)`
     """
     res_as = ref
-    if order >= 1:
+    if order == 1:
```
felixhekhorn, line 209: why did you change from `>=` to `==`? this way you have to repeat the definitions in all later branches

niclaurenti: I changed it because we were computing all the lower orders that we didn't need (e.g. for order==3 we compute as_LO, as_NLO and in the end as_NNLO)

AleCandido: Well, then split the two things: maintain the original behavior, i.e. `<=`, to compute the beta coefficients, and introduce and second switch (a second block of `if ... elif ... else`) with the `==` you are proposing just for the solution :)

AleCandido: In this way it will be as efficient as you obtained, but you don't need to duplicate the beta coefficients call

felixhekhorn: Or you can just compute the coefficients always (they are very cheap) and only defer the solvers ...

benchmarks/NNPDF_bench.py
```diff
@@ -35,6 +35,7 @@ def skip_pdfs(self, _theory):
 
 base_theory = {
     "Qref": 91.2,
+    "Qedref": 91.2,
```
AleCandido, line 38: QED you can also spell uppercase ^^

AleCandido: In any case, it is not the same `Q`, so also `Qqed` is fine, maybe even better. Just don't collapse the two Qs

felixhekhorn: there is the theoretical constraint that the two things _have to be given at the same scale_, isn't it? so we don't need this (otherwise they can never evolve together ...)

niclaurenti: It's how it's called in banana. We have to change banana in order to change it

AleCandido: Perfect, then change banana please :)
(unless is the name used by APFEL, check it - but I would change in any case, and translate for APFEL)

niclaurenti: @felixhekhorn yes but for some reason in banana the two Qref are separated. In eko I implemented it to be the same

felixhekhorn: as said above: then banana is just wrong ;-) it is used nowhere else but in eko. I'd suggest to revert

tests/eko/test_beta.py
```diff
@@ -44,7 +44,7 @@ def test_beta_aem3():
     """Test second beta function coefficient"""
     # from hep-ph/9706430
     np.testing.assert_approx_equal(
-        beta.beta_qed_aem3(5), -4.0 * 3 * (2 * 16 / 81 + 3 * 1 / 81)
+        beta.beta_qed_aem3(5), -4.0 * 3 * (3 + 2 * 16 / 81 + 3 * 1 / 81)
```
felixhekhorn, line 47: are you sure the reference is in that paper?

src/eko/kernels/QEDnon_singlet.py
```diff
@@ -0,0 +1,364 @@
+# -*- coding: utf-8 -*-
```
felixhekhorn, line 1: first trivial comment: rename this file to `non_singlet_qed.py` and similar for the other files here (`_qed` to the end)

src/eko/kernels/QEDnon_singlet.py
```diff
@@ -0,0 +1,364 @@
...
+                )
+                # res = np.prod(
+                #   [
+                #       nnlo_aem2_exact(
+                #           gamma_ns, a_steps[i],
+                #           a_steps[i-1], aem_list[step],
+                #           nf
+                #       ) for i in range(1, 1 + ev_op_iterations)
+                #   ]
+                # )
```
felixhekhorn, line 362:
```suggestion
```

drop if you don't need

src/eko/kernels/QEDnon_singlet.py
```diff
@@ -0,0 +1,364 @@
...
+import numba as nb
+import numpy as np
+
+from .. import beta
+from . import evolution_integrals as ei
+from . import non_singlet, utils
+
+
+@nb.njit(cache=True)
+def lo_aem1_exact(gamma_ns, a1, a0, aem, nf):  # lo refers to the order in as1
```
felixhekhorn, line 12: let's enforce our new naming scheme also here: `as1_aem1_exact` (and similar of course)

felixhekhorn: (of course this also cancels the comment)

src/eko/anomalous_dimensions/__init__.py
```diff
@@ -21,18 +21,18 @@
...
+from .. import constants, harmonics
 from . import aem1, aem2, as1, as1aem1, as2, as3, as4
 
 
 @nb.njit(cache=True)
-def exp_singlet(gamma_S):
+def exp_matrix_2D(gamma):
     r"""
-    Computes the exponential and the eigensystem of the singlet anomalous dimension matrix
+    Compute the exponential and the eigensystem of the singlet anomalous dimension matrix.
```
felixhekhorn, line 31: drop reference here and below to singlet and just say 2d

src/eko/anomalous_dimensions/__init__.py
```diff
@@ -60,22 +60,54 @@ def exp_singlet(gamma_S):
...
+        e : numpy.ndarray
+            projectors on the eigenspaces of the matrix gamma :math:`\gamma(N)`
+    """
+    # compute Matrix of coefficients
+    w, v = np.linalg.eig(gamma)
+    tmp = np.linalg.inv(v)
+    dim = gamma.shape[0]
+    e = np.zeros((dim, dim, dim), np.complex_)
+    exp = np.zeros((dim, dim), np.complex_)
+    # TODO check if this loop can be entirely cast to numpy
```
felixhekhorn, line 101: maybe @AleCandido can double check, but I'm afraid the answer is no. Looking to [`outer`](https://numpy.org/doc/stable/reference/generated/numpy.outer.html?highlight=outer#numpy.outer) there is no `axis` or similar and `einsum` which is given as an alternative option [is not in numba](http://numba.pydata.org/numba-doc/dev/reference/numpysupported.html#linear-algebra) ... actually this is maybe even not needed since we're actually inside numba and so a loop should not be too expensive

src/eko/anomalous_dimensions/__init__.py
```diff
@@ -205,3 +237,270 @@ def gamma_singlet(order, n, nf):
...
+    Returns
+    -------
+        gamma_ns : numpy.ndarray
+            non-singlet anomalous dimensions
+    """
+    if mode in [10102, 10202]:
+        return constants.eu2 * aem1.gamma_ns(n, sx)
+    if mode in [10103, 10203]:
+        return constants.ed2 * aem1.gamma_ns(n, sx)
+    else:
```
felixhekhorn, line 340: you can drop the `else`

src/eko/kernels/QEDnon_singlet.py
felixhekhorn: actually the new scheme would be `as1aem1_exact` (without the underscore)

src/eko/anomalous_dimensions/__init__.py
AleCandido: The second line should be simple to do it with NumPy
```py
    for i in range(dim):
        e[i] = np.outer(v[:, i], tmp[i])

     exp = (e * np.exp(w)).sum(axis=0)
    return exp, w, e
```

AleCandido: > (not 100% sure about the axis, but I almost sure you are summing over the most external one, so 0)

And a trivial test at least shows it working

AleCandido: For `e` if I understood the einsum would be
```py
e = np.einsum("ij,jk->jik", v, tmp)
```
so the full one would be:
```py
exp = np.einsum("ij,jk,j->ik", v, tmp, np.exp(w))
```

niclaurenti: @AleCandido numba doesn't like it: ```Use of unsupported NumPy function 'numpy.einsum' or unsupported use of the function.```

AleCandido: This should be a solution, but it is not optimal:
```py
e = np.outer(v, tmp).reshape(dim, dim, dim, dim)[:, idx, idx, :]
exp = (e * np.exp(w)).sum(axis=0)
```

AleCandido: > @AleCandido numba doesn't like it: `Use of unsupported NumPy function 'numpy.einsum' or unsupported use of the function.`

I know, @felixhekhorn wrote it above. I just wanted to confirm it is the correct signature, and I used that to write Einstein notation without latex, nothing else...

AleCandido: If you want test the proposed solution, but profile it. I expect to be slower than yours, since it requires more operations (and more memory, but most likely it is irrelevant), and Numba should convert the loop to the same loops of NumPy.

If it is not slower, than it's a problem, i.e. Numba it is not doing enough.

niclaurenti: It still can't be compiled by numba. Moreover in this way both ```e``` and ```exp``` are different with respect to the original function. I would return to the ```for``` loop.

AleCandido: Ok, in my test they were the same, but we can check later on. How did you check it if it didn't compile?
In any case, it was more a test to benchmark the original solution, that I believe to be performant enough.

Only for my curiosity, can you tell me what Numba is not able to compile?

niclaurenti: I run the test ```test_ad.py``` where this function is compared with ```scipy.linalg.expm``` and it is broken.

Regarding numba, it stops at ```e = np.outer(v, tmp).reshape(dim, dim, dim, dim)[:, idx, idx, :]``` with the error
```
Failed in nopython mode pipeline (step: nopython frontend)
No implementation of function Function(<built-in function getitem>) found for signature:
 
 >>> getitem(array(complex128, 4d, C), Tuple(array(int64, 1d, C), slice<a:b>, array(int64, 1d, C), slice<a:b>))
```

AleCandido: Ah, `getitem` are the square brackets. This was unexpected...

AleCandido: Since you are testing them in `test_ad.py`, you can (also in the same test function) benchmark the performances of the two implementations with [`timeit`](https://docs.python.org/3/library/timeit.html) (standard library module).
You can also assert that one has not to be much worse than the other :)

AleCandido: Pass a lambda to `timeit.timeit`:
```py
scipy_time = timeit.timeit(lambda: scipy.linalg.expm(mymat))
numba_time = timeit.timeit(lambda: exp_matrix(mymat))
assert numba_time < 2*scipy_time
```
Where 2 is random value, pick your favorite...
(you can use also `timeit.repeat` to receive a bunch of attempts, and compute mean and variance for the comparison)

felixhekhorn:
@giacomomagni may I ask for your help in the review?
- [ ] can you please check the documentation? whether it is complete and comprehensible for someone who is familiar with splitting functions (such as you :upside_down_face: ), but not with the details of QED
- [ ] can you check that all LHA benchmarks are still okay?
- [ ] can you run a few simple APFEL benchmarks?

giacomomagni:
Hi @felixhekhorn, @niclaurenti. Okay I'm happy to do this, but I need some time since I'm now giving 3 "diffent" talks in the next two weeks XD, so it might take me a while

src/ekomark/benchmark/external/apfel_utils.py
```diff
@@ -34,8 +32,7 @@ def compute_apfel_data(
         ref : dict
             output containing: target_xgrid, values
     """
-
-    target_xgrid = operators["xgrid"]
+    target_xgrid = operators["interpolation_xgrid"]
```
giacomomagni, line 35: Is this correct?

src/eko/scale_variations/expanded.py
```diff
@@ -170,3 +170,178 @@ def singlet_variation(gamma, a_s, order, nf, L):
...
+
+    Returns
+    -------
+    numpy.ndarray
+        scale variation kernel
+    """
+    sv_ker = np.eye(2, dtype=np.complex_)
+    gamma = np.ascontiguousarray(gamma)
+    if not alphaem_running:
+        sv_ker = np.eye(2, dtype=np.complex_)
```
giacomomagni, line 310: these 2 lines are repeated here and below
```suggestion
```

src/eko/scale_variations/expanded.py
```diff
@@ -170,3 +170,178 @@ def singlet_variation(gamma, a_s, order, nf, L):
...
+        gamma = np.ascontiguousarray(gamma)
```
giacomomagni, line 311:
```suggestion
```

src/eko/scale_variations/expanded.py
```diff
@@ -170,3 +170,178 @@ def singlet_variation(gamma, a_s, order, nf, L):
...
+        logarithmic ratio of factorization and renormalization scale
+
+    Returns
+    -------
+    numpy.ndarray
+        scale variation kernel
+    """
+    sv_ker = np.eye(2, dtype=np.complex_)
+    gamma = np.ascontiguousarray(gamma)
+    if not alphaem_running:
```
giacomomagni, line 309: the two part of this if seems to be really similar except for some lines, maybe there is a way to avoid this repetition... (same for the singlet function)

src/eko/scale_variations/exponentiated.py
```diff
@@ -43,3 +43,63 @@ def gamma_variation(gamma, order, nf, L):
...
+        return gamma
+    else:
+        beta0qcd = beta.beta_qcd((2, 0), nf)
+        beta1qcd = beta.beta_qcd((3, 0), nf)
+        beta2qcd = beta.beta_qcd((4, 0), nf)
+        beta0qed = beta.beta_qed((0, 2), nf)
+        # beta1qed = beta.beta_qcd((0, 3), nf)
+        # beta01qcd = beta.beta_qcd((2, 1), nf)
+        if order[0] >= 4:
+            gamma[4, 0] -= (
```
giacomomagni, line 83: here we are repeating the same formulas of the QCD only scale variation, can you avoid this?

src/eko/kernels/valence_qed.py
```diff
@@ -0,0 +1,106 @@
...
+                betatot += betaQCD[i, j] * a_half ** (i + 1) * aem_list[step - 1] ** j
+                gamma += gamma_valence[i, j] * a_half**i * aem_list[step - 1] ** j
+        ln = gamma / betatot * delta_a
+        ek = np.ascontiguousarray(ad.exp_matrix_2D(ln)[0])
+        e = ek @ e
+    return e
+
+
+@nb.njit(cache=True)
+def dispatcher(  # pylint: disable=too-many-return-statements
```
giacomomagni, line 62:
```suggestion
def dispatcher(
```

src/eko/kernels/valence_qed.py
```diff
@@ -0,0 +1,106 @@
...
+            perturbative expansion order of U
+
+    Returns
+    -------
+        e_v : numpy.ndarray
+            singlet EKO
+    """
+    if method in ["iterate-exact", "iterate-expanded"]:
+        return eko_iterate(gamma_valence, a1, a0, aem_list, nf, order, ev_op_iterations)
+    raise NotImplementedError("Selected method is not implemented")
```
giacomomagni, line 106:
```suggestion
    raise NotImplementedError('Only "iterate-exact", "iterate-expanded" are implemented with QED')
```

src/eko/kernels/evolution_integrals.py
```diff
@@ -40,6 +41,64 @@ def j00(a1, a0, nf):
     return np.log(a1 / a0) / beta.beta_qcd((2, 0), nf)
 
 
+@nb.njit(cache=True)
+def j00_qed(a1, a0, aem, nf):
```
giacomomagni, line 45: Maybe you can split all these new integrals in a new file `evolution_integrals_qed`, this way should be simpler to read and maintain.

src/eko/harmonics/__init__.py
```diff
@@ -217,3 +218,50 @@ def compute_cache(n, max_weight, is_singlet):
...
+
+        .. math ::
+            [S_1(n/2),
+            S_2(n/2),
+            S_{3}(n/2),
+            S_1((n+1)/2),
+            S_2((n+1)/2),
+            S_{3}((n+1)/2),
+            g_3(n),
+            g_3(n+2)]
```
giacomomagni, line 247: I know this might be painful, so not sure if it's worth it: it should be possible to simplify this in as `g3(n) + something`. 
This is because `g3`  is `S[-2,1,n]` under the hood.

giacomomagni (commented):
Hi @niclaurenti, 
I left some minor comments. I haven't reviewed everything yet.
Sphinx documentation will be included in another PR? 
For instance what is the reference used for the scale variation formula?

src/ekomark/benchmark/external/apfel_utils.py
AleCandido: I would say no: it is the old standard, that has already been updated, and here reverted.

Most likely, it has been tested with something old, e.g. `ekomark` may still be old-styled.

src/eko/couplings.py
```diff
@@ -295,7 +291,9 @@ def couplings_expanded(order, couplings_ref, nf, scale_from, scale_to):
     lmu = np.log(scale_to / scale_from)
     res_as = expanded_qcd(couplings_ref[0], order[0], nf, lmu)
     res_aem = expanded_qed(couplings_ref[1], order[1], nf, lmu)
-    if order[0] >= 2 and order[1] >= 2:
+    # if order[0] >= 1 and order[1] >= 1:
+    # order[0] is always >=1
+    if order[1] >= 1:
```
giacomomagni, line 296: This if statement is different from the one before, can you remind why?

src/eko/couplings.py
```diff
@@ -380,10 +435,13 @@ def __init__(
             raise NotImplementedError("a_s beyond N3LO is not implemented")
         if order[1] not in [0, 1, 2]:
             raise NotImplementedError("a_em beyond NLO is not implemented")
+        if order[0] == 0:
```
giacomomagni, line 438: why do we need this?

src/eko/couplings.py
```diff
@@ -577,12 +644,244 @@ def rge(_t, a, beta_qcd_vec, beta_qed_vec):
...
+            beta_vec[0],
+            b_vec,
+            u,
+            a_ref[0],
+            method="Radau",
+            rtol=1e-6,
+        )
+        return np.array([rge_qcd, a_ref[1]])
+
+    # def compute_exact_decoupled_running(self, a_ref, nf, scale_qcd_from, scale_qed_from, scale_to):
```
giacomomagni, line 702: If you want to keep these 2 functions, add a comment explaining why they are commented

src/eko/evolution_operator/__init__.py
```diff
@@ -179,48 +254,130 @@ def quad_ker(
...
-
-    # compute the actual evolution kernel
-    if ker_base.is_singlet:
-        gamma_singlet = ad.gamma_singlet(order, ker_base.n, nf)
-        # scale var exponentiated is directly applied on gamma
-        if sv_mode == sv.Modes.exponentiated:
-            gamma_singlet = sv.exponentiated.gamma_variation(
-                gamma_singlet, order, nf, L
+    # TODO refactor calls
+    if order[1] == 0:
```
giacomomagni, line 258: Maybe also here it's easier to split the `quad_ker` in two different functions. All the cals have to be repeated, but they don't share much.

src/eko/anomalous_dimensions/aem1.py
```diff
@@ -93,4 +92,115 @@ def gamma_ns(N, s1):
...
+
+    Returns
+    -------
+    gamma_S_0 : numpy.ndarray
+        Leading-order singlet anomalous dimension matrix :math:`\gamma_{S}^{(0)}(N)`
+
+    See Also
+    --------
+    gamma_ns : :math:`\gamma_{qq}^{(0)}`
+    gamma_qg : :math:`\gamma_{qg}^{(0)}`
```
giacomomagni, line 191:
```suggestion
```

src/eko/anomalous_dimensions/aem1.py
```diff
@@ -93,4 +92,115 @@ def gamma_ns(N, s1):
...
+    )
+    return gamma_S_01
+
+
+@nb.njit(cache=True)
+def gamma_valence(N, nf, sx):
+    r"""
+    Compute the leading-order valence anomalous dimension matrix.
+
+    .. math::
```
giacomomagni, line 170: this docstring is still for singlet

niclaurenti:
> Hi @niclaurenti,
> I left some minor comments. I haven't reviewed everything yet.
> Sphinx documentation will be included in another PR?
> For instance what is the reference used for the scale variation formula?

Hi @giacomomagni, thanks, I will look through it.
Regarding the documentation, the answer is no, but I will add it (I think it's better in this documentation).
For the exponentiated scale variations I just applied the scale variation to alpha_em.
For the expanded I'm not completely sure about the QED implementation and I wanted to check it (but unfortunately i stll haven't done it).

src/ekomark/benchmark/external/apfel_utils.py
niclaurenti: I have changed it since ekomark was crashing after the merge. I think we should change ekomark to include the new name

src/eko/couplings.py
niclaurenti: In this case I was implementing the decoupled running (mostly for benchmarking against apfel) but I never did it to the end. Maybe we can just delete it.

src/eko/couplings.py
niclaurenti: I added it since we passed from PTO = 0, 1, 2 to order[0] = 1, 2, 3 in which the value order[0] = 0 now in meaningless (it would be no evolution)

niclaurenti: I added it since we passed from PTO = 0, 1, 2 to order[0] = 1, 2, 3 in which the value order[0] = 0 now in meaningless (it would be no evolution)

src/eko/couplings.py
niclaurenti: In this case i only check whether order[1]>=1 since i'm adding the mixed terms in the evolution: for a_s we have it if order[0>=1 (always) and order[1]>=1 ; for a_em we have it if order[1] >=1 and order[0] >=1 (always)

src/ekomark/benchmark/external/apfel_utils.py
giacomomagni: yes might be that this is the case

AleCandido: > I have changed it since ekomark was crashing after the merge. I think we should change ekomark to include the new name

Since you confirm it was crashing, I'm pretty sure that's the problem.

However, this is nice-to-have, but not really a required/prioritized feature: `ekomark` is not even running in the CI, so if needed can lack behind (but possibly not for too long, it is useful, so we don't want it to be broken).
In case you will merge this PR without recovering `ekomark`, remember to open a dedicated issue, in order to forget about the problem.

src/eko/harmonics/__init__.py
niclaurenti: Maybe yes, but I tried and I don't think it will be super easy. Anyway I don't think we would get a big improvement in speed being g3 an interpolated function.

src/eko/couplings.py
giacomomagni: I see then the fix is to remove the 0 in line 434 ( `if order[0] not in [0, 1, 2, 3, 4]: ` ) and remove also this check.

felixhekhorn: just to be sure: can we run eko with QED only using only the API? i.e. not via theory card? I think this is something we want

src/eko/harmonics/__init__.py
felixhekhorn: where are you actually using this? and when you generate the expression for the anomalous dimensions, I guess you have Mathematica in the game any ways, right? so there you can do the replacement easily ... since QED is slow already, we should be careful about calling harmonics - they are the most expensive part (and all of them are in some sense interpolated)

src/eko/anomalous_dimensions/__init__.py
felixhekhorn: if this is resolved - please remove the TODO

src/eko/harmonics/__init__.py
niclaurenti: I'm using it in the O(as1aem1) and O(aem2) anomalous dimensions and I'm generating it in anomalous_dimensions/__init__. Unfortunately ```S_i(n/2)``` and ```S_i(n/2+1/2)``` cannot be written as a function of ```S_i(n)``` so we need to compute all of them.

giacomomagni: From what I understand it should be possible to map them into `S_i` and `S_{-i}`, using relations reported [here](https://eko.readthedocs.io/en/latest/theory/Mellin.html#harmonic-sums), but still you will be left with 2 dof.

felixhekhorn: Okay, so
- you use them here
  https://github.com/NNPDF/eko/blob/88e7284b74e3899048b5eb6dfb50f16b583ed512/src/eko/anomalous_dimensions/as1aem1.py#L244
  https://github.com/NNPDF/eko/blob/88e7284b74e3899048b5eb6dfb50f16b583ed512/src/eko/anomalous_dimensions/as1aem1.py#L311
and no where else, correct?
- yes, $S_1(N)$ and $S_1((N+1)/2)$ are linearly independent
- still, from my intuition I would say, that it should be possible to recast all arguments to the list given in https://github.com/NNPDF/eko/blob/88e7284b74e3899048b5eb6dfb50f16b583ed512/src/eko/anomalous_dimensions/as2.py#L16
  because the underlying diagrams/topologies are the same

niclaurenti: > Okay, so
> 
> * you use them here
>   https://github.com/NNPDF/eko/blob/88e7284b74e3899048b5eb6dfb50f16b583ed512/src/eko/anomalous_dimensions/as1aem1.py#L244
>   https://github.com/NNPDF/eko/blob/88e7284b74e3899048b5eb6dfb50f16b583ed512/src/eko/anomalous_dimensions/as1aem1.py#L311
>   and no where else, correct?

They are also used in : 
https://github.com/NNPDF/eko/blob/feature/qed-solution/src/eko/anomalous_dimensions/aem2.py#L154
https://github.com/NNPDF/eko/blob/feature/qed-solution/src/eko/anomalous_dimensions/aem2.py#L193
and all the others non singlet at O(aem2)

> * still, from my intuition I would say, that it should be possible to recast all arguments to the list given in https://github.com/NNPDF/eko/blob/88e7284b74e3899048b5eb6dfb50f16b583ed512/src/eko/anomalous_dimensions/as2.py#L16
>   because the underlying diagrams/topologies are the same

Yes the same harmonic sums are used at O(as2) and I wanted to pass them also to these anomalous dimensions. But maybe it can be done in another PR since this one is already big enough.

felixhekhorn: > But maybe it can be done in another PR since this one is already big enough.

Agreed

felixhekhorn: Maybe keep a note somewhere

src/eko/anomalous_dimensions/__init__.py
felixhekhorn: why did you mark this resolved?

src/eko/anomalous_dimensions/__init__.py
```diff
@@ -205,3 +237,270 @@ def gamma_singlet(order, n, nf):
...
+        gamma_ns : numpy.ndarray
+            non-singlet anomalous dimensions
+    """
+    if mode == 10102:
+        return constants.eu2 * as1aem1.gamma_nsp(n, sx, sx_ns_qed)
+    elif mode == 10103:
+        return constants.ed2 * as1aem1.gamma_nsp(n, sx, sx_ns_qed)
+    elif mode == 10202:
+        return constants.eu2 * as1aem1.gamma_nsm(n, sx, sx_ns_qed)
+    elif mode == 10203:
```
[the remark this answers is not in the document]
felixhekhorn: why did you mark this resolved?

src/eko/anomalous_dimensions/aem1.py
```diff
@@ -1,96 +1,202 @@
...
+    --------
+    gamma_ns : :math:`\\gamma_{qq}^{(0)}`
+    gamma_qg : :math:`\\gamma_{qg}^{(0)}`
+    gamma_gq : :math:`\\gamma_{gq}^{(0)}`
+    gamma_gg : :math:`\\gamma_{gg}^{(0)}`
+    """
+    e2avg = constants.e2avg(nf)
+    vue2m = constants.vue2m(nf)
+    vde2m = constants.vde2m(nf)
+    e2delta = vde2m - vue2m + e2avg
```
felixhekhorn, line 131: since this is repeated at least twice (and potentially even more often), I wonder whether it is worth to wrap them into a single function:
`e2avg,vue2m,vde2m,e2delta = constants.collect_all_charges(nf)` (function name is subject to discussion)

src/eko/anomalous_dimensions/aem2.py
```diff
@@ -287,39 +295,153 @@ def gamma_nsmd(N, nf, sx):
...
+
+    Returns
+    -------
+    gamma_singlet : numpy.ndarray
+        :math:`O(a_{em}^2)` singlet anomalous dimension :math:`\\gamma_{S}^{(0,2)}(N,nf,sx)`
+    """
+    nu = constants.uplike_flavors(nf)
+    nd = nf - nu
+    vu = nu / nf
+    vd = nd / nf
```
felixhekhorn, line 359: again: if it is repeated let's wrap it (see comment on aem1)

src/eko/anomalous_dimensions/as1.py
```diff
@@ -104,38 +100,122 @@ def gamma_gg(N, s1, nf):
...
     gamma_qq = gamma_ns(N, s1)
     gamma_S_0 = np.array(
         [[gamma_qq, gamma_qg(N, nf)], [gamma_gq(N), gamma_gg(N, s1, nf)]], np.complex_
     )
     return gamma_S_0
+
+
+@nb.njit(cache=True)
+def gamma_singlet_qed(N, s1, nf):
+    r"""Compute the leading-order singlet anomalous dimension matrix.
```
felixhekhorn, line 141: mention QED somewhere

src/eko/anomalous_dimensions/as2.py
```diff
@@ -237,40 +232,127 @@ def gamma_gg(n, nf, sx):
...
     gamma_S_0 = np.array(
         [[gamma_qq, gamma_qg(n, nf, sx)], [gamma_gq(n, nf, sx), gamma_gg(n, nf, sx)]],
         np.complex_,
     )
     return gamma_S_0
+
+
+@nb.njit(cache=True)
+def gamma_singlet_qed(N, nf, sx):
+    r"""Compute the leading-order singlet anomalous dimension matrix.
```
felixhekhorn, line 275: Please fix docstring (mention QED and fix "returns" and "see also")

src/eko/anomalous_dimensions/as2.py
```diff
@@ -237,40 +232,127 @@ def gamma_gg(n, nf, sx):
...
+            [0.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j, gamma_ns_p],
+        ],
+        np.complex_,
+    )
+    return gamma_S
+
+
+@nb.njit(cache=True)
+def gamma_valence_qed(N, nf, sx):
+    r"""Compute the leading-order valence anomalous dimension matrix.
```
felixhekhorn, line 322: Fix docstring

src/eko/anomalous_dimensions/as3.py
```diff
@@ -546,42 +539,129 @@ def gamma_gg(n, nf, sx):
...
     gamma_S_0 = np.array(
         [[gamma_qq, gamma_qg(N, nf, sx)], [gamma_gq(N, nf, sx), gamma_gg(N, nf, sx)]],
         np.complex_,
     )
     return gamma_S_0
+
+
+@nb.njit(cache=True)
+def gamma_singlet_qed(N, nf, sx):
+    r"""Compute the leading-order singlet anomalous dimension matrix.
```
felixhekhorn, line 584: Fix docstring

src/eko/anomalous_dimensions/as3.py
```diff
@@ -546,42 +539,129 @@ def gamma_gg(n, nf, sx):
...
+            [0.0 + 0.0j, 0.0 + 0.0j, 0.0 + 0.0j, gamma_np_p],
+        ],
+        np.complex_,
+    )
+    return gamma_S
+
+
+@nb.njit(cache=True)
+def gamma_valence_qed(N, nf, sx):
+    r"""Compute the leading-order valence anomalous dimension matrix.
```
felixhekhorn, line 631: Fix docstring

src/eko/basis_rotation.py
```diff
@@ -68,15 +89,68 @@
 )
 """|pid| representation of :data:`evol_basis`."""
 
-non_singlet_pids_map = {"ns-": 10201, "ns+": 10101, "nsV": 10200}
+unified_evol_basis_pids = tuple(
+    [21, 22, 100, 101, 200, 201]
+    + [103 + 2, 203 + 2]  # d = +2
```
felixhekhorn, line 94: please make your choice of `+2` (which is just fine) more visible - e.g. in a docstring somewhere (likewise for Sdelta and Vdelta)

felixhekhorn (changes requested):
Some minor stuff

src/eko/kernels/evolution_integrals_qed.py
```diff
@@ -0,0 +1,303 @@
+r"""Compute evolution integrals needed for QED."""
```
felixhekhorn, line 1: This is sharing too much code with its QCD counterpart! I believe you can abstract the function there (e.g. write a function `j00` that takes beta0 as an argument and that can be used by both side

src/eko/kernels/evolution_integrals_qed.py
```diff
@@ -0,0 +1,303 @@
...
+    j00 : float
+        integral
+    """
+    beta0 = beta.beta_qcd((2, 0), nf) + aem * beta.beta_qcd((2, 1), nf)
+    return np.log(a1 / a0) / beta0
+
+
+@nb.njit(cache=True)
+def jm10(a1, a0, aem, nf):
+    r"""LO-LO QED exact evolution integral.
```
felixhekhorn, line 39: please find a different name for the "m1" ... no idea which one though ... a stupid one would be "(1/N)LO"

felixhekhorn (changes requested):
Rewriting the evolution integrals is a bit more complex I assume ...

src/eko/anomalous_dimensions/__init__.py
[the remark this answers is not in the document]
niclaurenti: Because the error is raised only for the order O(aem1), since it is useless to check it for all orders. If it doesn't fail at the beginning it will not fail later.

src/eko/kernels/evolution_integrals_qed.py
niclaurenti: One idea that I had was to call it 0 and to shift all of the others by one unit

src/eko/kernels/singlet_qed.py
```diff
@@ -0,0 +1,101 @@
...
+
+    Returns
+    -------
+    e_s^{order} : numpy.ndarray
+        singlet QEDxQCD iterated (exact) EKO
+    """
+    a_steps = utils.geomspace(a0, a1, 1 + ev_op_iterations)
+    e = np.identity(4, np.complex_)
+    betaQCD = np.zeros((4, 3))
+    for i in range(1, 3 + 1):
```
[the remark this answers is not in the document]
niclaurenti: I think that the 3 is just wrong. should be 2

src/eko/anomalous_dimensions/__init__.py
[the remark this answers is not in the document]
felixhekhorn: that's not what I meant! if you can't find the `mode` either always raise an error (maybe my preferred way and as you do in the function right above this one) or return `None` (as you do here). If you decide to raise at a different level, that is just fine, but please do it consistently all over the place

src/eko/constants.py
```diff
@@ -81,44 +82,13 @@ def e2avg(nf):
...
-    -------
-    vd * e2m : float
-
-    """
-    nd = nf - uplike_flavors(nf)
-    return nd / nf * (eu2 - ed2)
+    e2avg = (nu * eu2 + nd * ed2) / nf
+    vue2m = nu / nf * (eu2 - ed2)
+    vde2m = nd / nf * (eu2 - ed2)
+    return e2avg, vue2m, vde2m
```
felixhekhorn, line 94: why don't you add `e2delta`? it is often used in conjunction with these others?

src/eko/anomalous_dimensions/as1.py
```diff
@@ -104,38 +100,122 @@ def gamma_gg(N, s1, nf):
...
+        Mellin moment
+    s1 : complex
+        harmonic sum :math:`S_{1}`
+    nf : int
+        Number of active flavors
+
+    Returns
+    -------
+    gamma_S_0 : numpy.ndarray
+        Leading-order singlet anomalous dimension matrix :math:`\\gamma_{S}^{(0)}(N)`
```
felixhekhorn, line 163: Please check the full docstring :upside_down_face:

src/eko/kernels/evolution_integrals_qed.py
niclaurenti: On the other hand using the `-1` is the only way in which `jm1n` satisfies the definition of `j(m,n)` given at the top of the module

felixhekhorn: > One idea that I had was to call it 0 and to shift all of the others by one unit

maybe this is the way to go: we label them by $j^{u,v} = \int \alpha_s^u/\beta^{v-2}_{QCD}$ - so first index shifts by 1 and second by two (since beta_QCD start at this order) ... also I was more worried about the docstring ... and if we follow this suggestion, then maybe we just drop the mention of "LO" (there is no unique LO anymore) and write something like ```[insert details here] evolution integral :math:`\sim \alpha^0/\beta^{(0)}` ```

src/eko/kernels/evolution_integrals.py
```diff
@@ -14,12 +14,12 @@
 
 
 @nb.njit(cache=True)
-def jm10(a1, a0, beta0):
-    r"""LO-LO QED exact evolution integral.
+def j02(a1, a0, beta0):
+    r"""j^{(0,2)} exact evolution integral.
```
felixhekhorn, line 18: I think you might need a ```:math:`bla` ``` here - check sphinx

src/eko/kernels/non_singlet_qed.py
```diff
@@ -30,12 +30,12 @@ def as1aem1_exact(gamma_ns, a1, a0, aem, nf):
     """
     return np.exp(
         (gamma_ns[1, 0] + aem * gamma_ns[1, 1]) * ei.j12(a1, a0, aem, nf)
-        + aem * gamma_ns[0, 1] * ei.j02(a1, a0, aem, nf)
+        + aem * gamma_ns[0, 1] * np.log(mu2_from / mu2_to)
```
felixhekhorn, line 33: Should we drop in consequence the `j0x` from the code?

niclaurenti: I think yes

tests/eko/kernels/test_kernels_QEDns.py
```diff
@@ -159,7 +159,7 @@ def test_ode():
                     evmod,
                     masses,
                     hqm_scheme=QuarkMassSchemes.POLE,
-                    thresholds_ratios=MatchingScales(c=1.0, b=1.0, t=1.0),
+                    thresholds_ratios=None,
```
felixhekhorn, line 162: I believe this should be
```suggestion
                    thresholds_ratios=[1., 1., 1.],
```
and below of course (please check yourself)

tests/eko/kernels/test_kernels_QEDvalence.py
```diff
@@ -1,12 +1,13 @@
-# -*- coding: utf-8 -*-
 import warnings
 
 import numpy as np
 import pytest
 
-from eko import anomalous_dimensions as ad
 from eko.kernels import valence_qed as val
 
+# from ekore.anomalous_dimensions.unpolarized import space_like as ad
```
felixhekhorn, line 8:
```suggestion
```
if this is not needed

tests/ekore/anomalous_dimensions/unpolarized/space_like/test_init.py
```diff
@@ -38,18 +39,18 @@ def test_eigensystem_gamma_singlet_0_values():
...
-    gamma_S_0_qed = as1.gamma_singlet_qed(3, s1, NF)
+    gamma_S_0_qed = ad_us.as1.gamma_singlet_qed(3, s1, NF)
     res = expm(gamma_S_0_qed)
     res2 = ad.exp_matrix(gamma_S_0_qed)[0]
     assert_allclose(res, res2)
-    gamma_v_0_qed = as1.gamma_valence_qed(3, s1)
-    res = expm(gamma_v_0_qed)
-    res2 = ad.exp_matrix(gamma_v_0_qed)[0]
-    assert_allclose(res, res2)
+    # gamma_v_0_qed = ad_us.as1.gamma_valence_qed(3, s1)
```
felixhekhorn, line 50: is this needed?

tests/ekore/anomalous_dimensions/unpolarized/space_like/test_init.py
```diff
@@ -6,17 +6,18 @@
 from numpy.testing import assert_allclose, assert_almost_equal, assert_raises
 from scipy.linalg import expm
 
-from ekore import anomalous_dimensions as ad
+import ekore.anomalous_dimensions.unpolarized.space_like as ad_us
```
felixhekhorn, line 9: Are you sure to run pre-commit? isort should group `ekore` together, I think

tests/ekore/anomalous_dimensions/unpolarized/space_like/test_init.py
niclaurenti: Here I wanted to check that the exponenential that we implemented gives the same result of expm from scipy. However the function `exp_matrix_2D` is not able to compute the exponential of a matrix proportional to the identity (like in the case of gamma_V0_qed). That's why I removed it

tests/ekore/anomalous_dimensions/unpolarized/space_like/test_init.py
niclaurenti: Yes I am.

src/eko/evolution_operator/__init__.py
```diff
@@ -688,27 +688,27 @@ def compute_aem_list_as(self):
...
         else:
-            as_start = as0
-            aem_start = aem0
+            a_start = np.array([as0, aem0])
+            mu2_start = self.q2_from
         sc = self.managers["couplings"]
         ev_op_iterations = self.config["ev_op_iterations"]
-        as_steps = utils.geomspace(as0, as1, 1 + ev_op_iterations)
-        as_l = as_steps[0]
+        mu2_steps = np.linspace(self.q2_from, self.q2_to, 1 + ev_op_iterations)
```
felixhekhorn, line 701: Why did you change to `linspace`? the natural variable in evolution is $\ln(Q^2)$ (look at LO alpha evolution)

src/eko/kernels/non_singlet_qed.py
```diff
@@ -246,23 +246,13 @@ def running_alphaem_exact(
         non-singlet EKO
     """
     a_steps = utils.geomspace(a0, a1, 1 + ev_op_iterations)
+    mu2_steps = np.linspace(mu2_from, mu2_to, 1 + ev_op_iterations)
```
felixhekhorn, line 249: see [above](https://github.com/NNPDF/eko/pull/135/files/d5523e901ee8ace35b47b6ff04e9b88a831f3708..e21fbaf47e8f7d3690df7a9a75bcffe3ba57ddb3#r1115957206)

src/eko/scale_variations/expanded.py
```diff
@@ -201,7 +201,7 @@ def non_singlet_variation_qed(gamma, a_s, a_em, alphaem_running, order, nf, L):
     sv_ker = non_singlet_variation(gamma[1:, 0], a_s, order, nf, L)
     if alphaem_running:
         if order[1] >= 2:
-            sv_ker += a_em * variation_as1(gamma[0, 1:], L)
+            sv_ker -= a_em * variation_as1(gamma[0, 1:], L)
```
felixhekhorn, line 204: Just for my personal understanding: the minus here is the same as added in https://github.com/NNPDF/eko/pull/191 ?

src/eko/couplings.py
```diff
@@ -272,20 +269,19 @@ def expanded_qed(ref, order, nf, lmu):
...
+        res_aem = expanded_nlo(ref, beta0, b_vec[1], lmu)
     return res_aem
 
 
 @nb.njit(cache=True)
-def couplings_expanded_alphaem_running(order, couplings_ref, nf, scale_from, scale_to):
+def couplings_expanded_alphaem_running(
+    order, couplings_ref, nf, scale_from, scale_to, decoupled_running
+):
     r"""Compute coupled expanded expression of the couplings for running alphaem.
```
felixhekhorn, line 285: If this is already doing the decoupled case, please update the doc here and below

src/eko/scale_variations/expanded.py
niclaurenti: Honestly I was confused by the minus sign, but since in the qcd case it is there, then I put it.

src/eko/evolution_operator/__init__.py
niclaurenti: I agree, but I have to use it in `sc.compute` (few lines after) that wants a $\mu_2$

felixhekhorn: Mh? I don't understand ... `geomspace` is just saying how to distribute the values; so whether you have a list `[1,2,4]` or `[1,2.5,4]` ... the actual element has always the same unit ($\mu^2$ to be specific)

niclaurenti: Ah ok I thought that you wanted to divide in steps $\log(\mu^2)$ instead of $\mu^2$.
I did it since the i-th $a_{em}$ value must be in the interval $a_s^{(i+1)} - a_s^{(i)}$, and since the $a_s$ values are logspaced then $\mu^2$ should be linspaced.

src/eko/couplings.py
niclaurenti: Yes I added it but if you notice for the moment it is always set to `False`.

doc/source/theory/pQCD.rst
```diff
@@ -129,10 +129,11 @@ Order specification
...
 of :math:`a_s = \alpha_s/(4\pi)` and :math:`a_{em} = \alpha_{em}/(4\pi)`. The available perturbative expansions are the following:
 
-- ``order=(0,0)``: it is the non evolution case in which :math:`a_s` and :math:`a_{em}` are kept fixed and the splitting functions are null.
 - ``order=(n,0)``: with :math:`n=1,2,3,4` correspond to the pure |QCD| evolution at |LO|, |NLO|, |NNLO| and |N3LO| in which the |QCD| splitting functions are expanded up to :math:`\mathcal{O}(a_s^n)` and the strong coupling is evolved using the n-th coefficient of the beta function, i.e. :math:`\beta_{n-1}`.
 - ``order=(n,m)``; with :math:`n=1,2,3,4` and :math:`m=1,2` corresponds to the mixed |QED| :math:`\otimes` |QCD| evolution in which the splitting functions are expanded up to :math:`\mathcal{O}(a_s^na_{em}^m)`, the stromg coupling is evolved using up to the n-th coefficient of the beta function and the electromagnetic coupling is kept fixed.
 
+Observe that the case :math:`n=0` is not allowed, since it would correspond to the pure |QED| evolution or (if :math:`m > 0`) no evolution at all.
```
felixhekhorn, line 135: Since we need to close this PR ASAP, let's keep it here, but maybe we want to add an issue were we propose to revert that restriction

src/eko/evolution_operator/__init__.py
```diff
@@ -685,37 +682,46 @@ def a_em(self):
 
     def compute_aem_list(self):
         """Return the list of the couplings for the different values of :math:`a_s`."""
+        ev_op_iterations = self.config["ev_op_iterations"]
```
felixhekhorn, line 685: Can you please add a bit more documentation/comments to this function, which is fairly complicated?

Again, maybe open an issue: we may want to reconsider the split between `Operator` and `kernels` ... If I understand this function correctly here you're splitting the evolution integral, which so far was done below (see e.g. [here](https://github.com/NNPDF/eko/blob/45682d7775210f262529d0f639fd1d21a48b5457/src/eko/kernels/singlet.py#L331)), but now you no longer can't because you need to call a function (from `Couplings`). - Actually, why are you using `compute` (internal) and not `a` (public)? this way you never do the matching! - Coming back to the potential issue: now you're sensitive to how you choose the steps, and (by chance) you're currently consistent [here](https://github.com/NNPDF/eko/blob/47d46ce671f78fededb201604fc68e0a35532c06/src/eko/evolution_operator/__init__.py#L706) and [below](https://github.com/NNPDF/eko/blob/47d46ce671f78fededb201604fc68e0a35532c06/src/eko/kernels/non_singlet_qed.py#L254). Instead, we maybe want to compute a list of tuples `[(mu0,as0,a0),(mu1,as1,a1),...]` here (because we have to) and pass that around

felixhekhorn (commented):
325 commits so far - I'm afraid I have to take some responsibility for that ...

AleCandido:
> 325 commits so far - I'm afraid I have to take some responsibility for that ...

Commits are not always a good measurement unit, in this case I believe the more relevant one is "days". Even though the two things are of course correlated (and the conversion in this case approaches 1).

Since this PR is never-ending, and we are being unable to merge, and even to review, we should reconsider a minimal goal and way to test.
Something like:
1. it should change existing tests as little as possible (new ones are more than welcome), and of course pass them
2. it should pass a few relevant benchmarks against APFEL
3. existing configurations should be unaffected; to be tested in practice by rerunning some EKOs with this branch and main, and test for a compatible (if possible bit-per-bit equal) result

If the present status passes all these checks, I'd rather merge even if it is not perfect, and incrementally improve and refactor, otherwise we'll keep some manpower stuck on this development, and pay over and over the cost of reviewing an extensive bunch of code...

src/eko/evolution_operator/__init__.py
niclaurenti: What this function does is the following:

In the case of running `aem` we need the values of `aem` corresponding to the middle point of the `as` interval, so we need to find $a_{em}(a_s)$. This can be done for the exact evolution (you just write $da_{em}/da_s = \beta_{qed} / \beta_{qcd}$ and you solve numerically) but not for the expanded evolution (or at least I wasn't able to solve the differential equation perturbatively).

For this reason I splitted the $mu^2$ interval in steps and for every step I compute the value of $a_s$ at the borders of the interval (these are now what before was computed inside kernels as `utils.geomspace(a0, a1, 1 + ev_op_iterations)`) and the values of $a_{s,em}$ in the middle point of the $mu$-interval (this is what is going in the computation of the $\gamma$ and $\beta$ in `eko_iterate`)
I know that this is not exactly the same than computing the values of $a_{em}(\frac{a_s^k+a_s^{k+1}}{2}) with $a_s^k$ divided with `utils.geomspace(a0, a1, 1 + ev_op_iterations)` but the distance between the $a_s$ is very small and moreover it is just a different way of splitting the $a_s$ integration interval.

Regarding your questions:

I'm using `compute` and not `a` since `a` always starts from `Qref` while I wanted to evolve from either `q2_to` or `q2_from` (the one that is closer to `Qref`), given that we have already computed the $a_s$ and $a_{em}$ values in these points. In this way the numerical error should be smaller when computing a point that is close to `as1` or `as0`. Then, you are right that I never do the mathcing of `as` but it is not needed since the evolution is computed in one fixed flavor interval at the time (the matching is done when we compute `as0` and `as1`).

doc/source/theory/pQCD.rst
felixhekhorn: https://github.com/NNPDF/eko/issues/223

src/eko/evolution_operator/__init__.py
felixhekhorn: moved to #224

felixhekhorn:
> 1. it should change existing tests as little as possible (new ones are more than welcome), and of course pass them
> 
> 2. it should pass a few relevant benchmarks against APFEL
> 
> 3. existing configurations should be unaffected; to be tested in practice by rerunning some EKOs with this branch and main, and test for a compatible (if possible bit-per-bit equal) result

@niclaurenti if you confirm @AleCandido's  list- let's merge!

felixhekhorn (approved):

niclaurenti:
> > 1. it should change existing tests as little as possible (new ones are more than welcome), and of course pass them
> > 2. it should pass a few relevant benchmarks against APFEL
> > 3. existing configurations should be unaffected; to be tested in practice by rerunning some EKOs with this branch and main, and test for a compatible (if possible bit-per-bit equal) result
> 
> @niclaurenti if you confirm @AleCandido's list- let's merge!

Yes they are not affected. But let me run some pure QCD benchmarks in order to be 100% sure

PR #152, opened 2022-10-11
Use temporary working directory for output

### LINKED ISSUE #139: Work in a Temporary Folder
Instead of using an archive at runtime, that is rather impractical, let's use a temporary folder.

The archive is only generated on save, and load when an EKO is instatiated.

See https://github.com/N3PDF/eko/pull/105#discussion_r925751448

- [x] encode q2 by bytes in filenames
- [x] add contextmanager á la `open()` (to and from tar), wrapping:
     - [x] function to open, extracting the tar
     - [x] function to close, compressing to tar
- [x] add CLI for conversion (almost a wrapper of `output.legacy.load_tar()`) in `ekobox`

felixhekhorn:
Closes #139 

actually I feel this implementation in conceptually easier ...

we're running in a naming problem again ;-) since the new scheme will be: `with eko.open("bla.tar") as eko_:` both `eko` and `EKO` are taken at this point, but that is the least of our problems ...

todo:
- [x] fix `deepcopy`
- [x] fix tests


src/eko/output/struct.py
```diff
@@ -41,7 +53,6 @@ class DictLike:
 
     Some collections and scalar objects are normalized to native Python
     structures, in order to simplify the on-disk representation.
-
```
AleCandido, line 44: The extra empty line is part of the NumPy style, please, don't remove it:
https://www.sphinx-doc.org/en/master/usage/extensions/example_numpy.html

AleCandido (changes requested):

tests/conftest.py
```diff
@@ -40,6 +41,13 @@ def fake_pdf():
     return FakePDF()
 
 
+@pytest.fixture
+def default_cards():
+    t = tc.generate(0, 1.0)
+    o = oc.generate([10.0])
```
AleCandido, line 47: Use keyword arguments for all of them, it is not intuitive what is what.

AleCandido (changes requested):

tests/conftest.py
AleCandido: Even more: maybe we should make them only keyword, such that you can't use this way (prepending a `*` as first argument)

src/eko/output/struct.py
felixhekhorn: Looking there: `function_with_pep484_type_annotations` has it indeed, but `function_with_types_in_docstring` not, so I'd say there is no clear indication ... however, in the future I can do, but I'll not revert here

AleCandido: Yes, I know, but it is not only ``, there are also:
1. `module_level_function`
2. `example_generator`
3. `ExampleError` class
4. `ExampleClass` class
5. its `__init__` method
6. its `example_method`
7. its `__special__` method
8. and `_private` as well
9. as the module level docstring

I'd say the indication is not so unclear

AleCandido: In particular, since you took the burden of removing 8 of them, not a single one, please revert them as well.

If it was only for one, we could have done at any stage, but to enforce a convention we can not keep having plenty of exceptions, including in the new code.

AleCandido: Btw: they are also polluting the diff, making it more difficult to assess the status of this PR from the "Files changed" tab. Since this PR is not finishing today, and we'll have to merge and reassess its status, please keep it as clean as possible.

AleCandido:
At the moment I don't have time to go through this, but I hope to have as soon as I will have finished my current task.

However, since everything else is moving on, this PR is becoming more an obstacle than a first step, so I would also reconsider dismissing completely and restart from the present status. If I'll implement myself, I believe I'll do it.

felixhekhorn:
> At the moment I don't have time to go through this, but I hope to have as soon as I will have finished my current task.
> 
> However, since everything else is moving on, this PR is becoming more an obstacle than a first step, so I would also reconsider dismissing completely and restart from the present status. If I'll implement myself, I believe I'll do it.

No. There is nothing wrong with the code in this PR - it is exactly the agreed strategy. It would be stupid to trash my time, effort and energy (we are already short of man power).

the remaining points  from the issue #139 

> encode q2 by bytes in filenames
> add CLI for conversion (almost a wrapper of output.legacy.load_tar()) in ekobox

can be addressed in a separate PR and none of them a crucial.

AleCandido:
> No. There is nothing wrong with the code in this PR - it is exactly the agreed strategy. It would be stupid to trash my time, effort and energy (we are already short of man power).

Never said there is something wrong: I was arguing that there is not much, and it didn't take much time in the first place.
Your time, effort, and energy are relevant as everyone's else is.

At the moment, we need this, but we need #162 more (that's why someone else is already in charge of it, instead of completing this first).

If the time to retain this and reconcile with #162 is estimated to be more than writing from scratch, then it is smart (not stupid) to apply the more efficient path.
We are not perfect, and we have limited resources. So we are not even perfect in planning, and assuming we will never revert anything not to waste "efforts" is dangerous for the project itself.

AleCandido:
> the remaining points from the issue #139
> 
> > encode q2 by bytes in filenames
> > add CLI for conversion (almost a wrapper of output.legacy.load_tar()) in ekobox
> 
> can be addressed in a separate PR and none of them a crucial.

If you are rushing to merge before #162 because of possible conflicts, take into account that you might waste someone else's time.
Better to complete the content #139, at least the core part. I agree the CLI not to be crucial, we can postpone, but the q2 encoding was a relevant part, and quite a cheap one.

src/eko/__init__.py
```diff
@@ -1,14 +1,21 @@
...
 
+Please refer to our documentation for a full overview of the possibilities.
+"""
 from . import output, runner, version
 
+# export public constants
 __version__ = version.__version__
 
+# export public methods
+open = output.struct.EKO.open  # pylint: disable=redefined-builtin
```
AleCandido, line 11: Do not redefine built-in, much better to export `EKO` top-level, and use:

```py
import EKO from eko

with EKO.open() as e:
    ...
```
that is not much different from

```py
import eko

with eko.open() as e:
    ...
```
And no one will be tempted to `import open from eko`, overwriting a built-in function.
(in general, better not to silence Pylint's warnings as much as possible)

andreab1997:
> > No. There is nothing wrong with the code in this PR - it is exactly the agreed strategy. It would be stupid to trash my time, effort and energy (we are already short of man power).
> 
> Never said there is something wrong: I was arguing that there is not much, and it didn't take much time in the first place. Your time, effort, and energy are relevant as everyone's else is.
> 
> At the moment, we need this, but we need #162 more (that's why someone else is already in charge of it, instead of completing this first).
> 
> If the time to retain this and reconcile with #162 is estimated to be more than writing from scratch, then it is smart (not stupid) to apply the more efficient path. We are not perfect, and we have limited resources. So we are not even perfect in planning, and assuming we will never revert anything not to waste "efforts" is dangerous for the project itself.
> 
> P.S.: this PR in the first place is reverting something I implemented myself, if you ask @andreab1997 you will find out it is not that trivial to work with `tar` files at every step; but it was not a good idea, and I acknowledge it, so reverting is good, and I'm not complaining for wasted effort

Sorry to step in. Just to say that this PR is probably needed to address the first two points of the list I posted in #162. The other two points I can do but the last days I was working on MHOU and I did not have time to dedicate (sorry for this). So it is maybe true that we need #162 more that this PR but we need this PR in order to complete #162 (the point is that without this it is basically impossible to update the metadata file without creating another tar file each time).

src/eko/__init__.py
```diff
@@ -18,14 +25,12 @@ def run_dglap(theory_card, operators_card):
...
-            input card - see :doc:`/code/IO`
+    setup : dict
+        input card - see :doc:`/code/IO`
 
     Returns
     -------
-        output : dict
-            output dictionary - see :doc:`/code/IO`
+    output.EKO
+        output object - see :doc:`/code/IO`
```
AleCandido, line 34: `output object` is too generic, you can also leave it empty.

Please replace with something like "computed operator" - same length, but more explicit. If possible expand a bit.

src/eko/output/struct.py
```diff
@@ -329,6 +335,8 @@ class EKO:
     """Path on disk, to which this object is linked (and for which it is
     essentially an interface).
     """
+    working_dir: pathlib.Path
```
AleCandido, line 338: You can use `path`: there are not two distinct objects
- if the `EKO` is *opened*, then `path` /`wdir` is the working dir, and not the tar, since it is not associated to it (it might be loaded from a tar, as a NumPy array might be loaded from an `.npy` file, but after loading any association is lost)
- if the `EKO` is *closed*, it should not be used (on close the object should be destroyed), so again, you can dump it on a tar

So, I would decouple the two operations:
- `close()` will just clean up the folder
- `dump(path)` will save another tar (or `write_tar(path)`, here I'm not complaining about the name)

If you open with the context manager, then is the context manager that is in charge of using the same path, and it is retained in the function scope, so you don't need a class attribute for it.

src/eko/output/struct.py
```diff
@@ -650,42 +603,61 @@ def bootstrap(dirpath: os.PathLike, theory: dict, operator: dict):
...
+        path : os.PathLike
+            EKO file name
+        """
+        obj = None
+        try:
+            obj = cls.open_tar(path)
+            yield obj
+        finally:
+            obj.write_tar()
+            shutil.rmtree(obj.working_dir)
```
AleCandido, line 653: Move this to a separate `close()` function, such that is available also when the context manager is not used.

src/eko/output/struct.py
```diff
@@ -650,42 +603,61 @@ def bootstrap(dirpath: os.PathLike, theory: dict, operator: dict):
...
-    @staticmethod
-    def extract(path: os.PathLike, filename: str) -> str:
-        """Extract file from disk assets.
-
-        Note
-        ----
-        At the moment, it only support text files (since it is returning the
-        content as a string)
+    @classmethod
+    def open_tar(cls, path: os.PathLike):
```
AleCandido, line 607: Since tar is the only file format supported, I propose renaming the function:
```suggestion
    def load(cls, path: os.PathLike):
```

AleCandido (changes requested):
First batch of comments

AleCandido:
> (the point is that without this it is basically impossible to update the metadata file without creating another tar file each time).

It is the exact same of the current strategy to set an operator. If you need to create another `tar` file is just fine.

AleCandido:
>  * [x]  fix tests

Tests are failing

felixhekhorn:
Close in favour of #172

Issue #138, opened 2022-07-20
Composite Output - distributed computation


alecandido:
Follow up of #105 

- [ ] add `runcards` subfolder and move there runcards (`0.theory.yaml`, `0.operator.yaml`)
- [ ] create `recipes` subfolder and dump instructions
- [ ] create recipes' runner, dump into `parts` subfolder
- [ ] create parts' merger

UI:

- [ ] orchestrate `eko` run through CLI commands (`run` subcommand)

Further:

- [x] write runcards types (`eko.runcards`) and use them in `eko.output.struct.EKO`
- [x] rename `output` subpackage in something more appropriate (like `compute` or `orchestration`)
- [x] runcard dump broken: `Q0` and `Qref` are NumPy scalars in both theory and operators card
- [x] functions that play with `q2` should always act on existing element retrieved with a given precision (parameter of the object): if multiple object are found in the range, **raise**
  - `__getitem__`
  - `__setitem__`
  - `__contains__`
- [ ] rename `output/manipulate.py` to `output/rotation.py`
- [x] move `Rotation` from `output/struct.py` to `output/rotation.py`
- [x] move `Operator` from `output/struct.py` to `output/operator.py`
- [x] move `DictLike` from `output/struct.py` to `output/utils.py`

## Plan

Let's break all these over multiple PR:

### Recipe

- establish recipe format
  - self contained instruction to compute a part
  - do not repeat global information (like `Configs` and `Debug`) since they are always accessible for the full operator
- with dedicated in-memory struct
- provide recipe computer (from runcard-like input)

### Parts

- start splitting `operator.grid` instructions to compute parts from recipes
- provide a part struct
- benchmark with FFNS

### Operators

- compute full operators as well (merging separate parts)
- move operator struct
- benchmark with VFNS

### Runner replacement

- replace default runner, with a runner using recipes-parts-operators
- benchmark new runner against old one, over a wide range of conditions
- drop `operator.grid`

### Move to Ekobox

Since at this point the runner won't be our default way to compute an EKO, and Ekobox is sufficiently good, we can move the one-line runner to Ekobox (it will actually be there for user convenience, so makes sense to be part of the UI-oriented package).

In particular, we can also include as part of the EKO CLI, i.e.:
```sh
eko run thcard.yaml ocard.yaml [output-file]
```

## Layout

Following, the proposed layout backing the new mode:

```
runner.py
output/
  |-- legacy.py
  |-- struct.py
  |-- rotation.py
[...]

PR #159, opened 2022-11-04
Refactor runner

andreab1997:
We want to massively simplify `runnner.py`:

- In `get_output` we want to avoid the rotations: if needed they should be done afterwards. 
- In `__init__ ` we just need to skip the last part in which `targetpids`, `inputpids`, `targetgrid` and `inputgrid` are collected.
@felixhekhorn @AleCandido

### DIFF
--- a/src/eko/runner.py
+++ b/src/eko/runner.py
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""This file contains the main application class of eko."""
+"""Main application class of eko."""
 import copy
 import logging
 from typing import Optional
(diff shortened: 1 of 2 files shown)


AleCandido:
> In `get_output` we want to avoid the rotations: if needed they should be done afterwards.

In particular, in `get_output` only the basic three lines of computation and return will survive, keeping it minimal.

In any case, this runner will be completely reworked in the context of #138

andreab1997:
This should be ok now, we just need to fix the tests. Before doing that, can you confirm that this is what we want?

src/eko/runner.py
```diff
@@ -1,5 +1,5 @@
 # -*- coding: utf-8 -*-
-"""This file contains the main application class of eko."""
+"""Main application class of eko."""
```
AleCandido, line 2: Great, I appreciate :D

AleCandido:
> This should be ok now, we just need to fix the tests. Before doing that, can you confirm that this is what we want? @felixhekhorn @AleCandido

Yep, I confirm :)

felixhekhorn (approved):

AleCandido (approved):

andreab1997:
I am merging, thanks for this :)

PR #160, opened 2022-11-08
Reorganize runner

### LINKED ISSUE #114: Exterminate `from_dict`
At the moment, we've `from_dict` methods all over the place, that expect a runcard with all the elements they need, but not specifying which ones.
At LO they are recapped by `runner` (where most of these methods are used):
https://github.com/N3PDF/eko/blob/813bd656159cde5c5d176b54ca99139e2280cb44/src/eko/runner.py#L44-L55

The information that is thus passed down is very unclear, and changes in the runcard format might affect silently all these methods.

Since they are always using a small amount of parameters, I propose a close replacement, based on the idea of sections:
```py
bfd = interpolation.InterpolatorDispatcher(**operators_card["interpolation"])
```
This is somewhat a trade-off: it is not clear from the call which parameters are used (but it's clear that only one section is affected), and in any case can be determined by the callee (the signature of the function called).

### LINKED ISSUE #153: Runcards Dataclasses
There are three very good reasons to move to dedicated classes for loading runcards:

1. having more control over the structure 
    - at the moment there is a spread usage, and unclear definition in the docs
2. getting rid of dictionaries and `from_dict` methods (see also #114)
3. sanitize before dump in a unique place
    - as we know since long, YAML is tricked by NumPy scalars, making them very complex objects

They have been initialized, but nothing relevant has actually been done:
https://github.com/NNPDF/eko/blob/2f2755c9f316976eeec68d11a150915f137ba43d/src/eko/runcards.py#L1-L60

The idea is to have a first attempt, with all the sanitation we are aware, and then incrementally improve it over time.

E.g. use also `dict` in place of `output.struct.Debug` or `output.struct.Configs` for the time being - later on we can also consider moving them to the `runcards` module itself.

AleCandido:
At the moment, `runner` was more or less the core of EKO, despite us calling it "more a library than an application".

We should collect all stuffs related to runner, and improve it, also in consideration of the new runner coming from #138 

- [ ] write runcards objects, currently sketched in `runner.runcards` #153 
- [ ] the new runcard objects should be used, where now the dictionary runcard is passed
  - only the required information has to be passed, and nothing more (absolutely not the full runcard)
  - thus use the default object constructors, dropping the `from_dict` (incidentally, this will close #114)
- [ ] move `run_dglap()` out of top level, and scope in `runner` (if it is more a library, then there is no special reason the only object define top level should be the runner)
  - we can still export to the top scope for compatibility
- [ ] inherit also runcards object from `output.struct.DictLike`, since they are the objects mostly going to and from dictionary (because of serialization)
  - as a consequence, we should also move `DictLike` out of the `output` subpackage (as a rule of thumb, since subpackages are made to collect and isolate, they should not talk to each other, but only with the upper scope)

Closes #153


felixhekhorn:
Can we please drop [ade4abf](https://github.com/NNPDF/eko/pull/160/commits/ade4abff9a5a17229bf68b9fe9278bec1a1ab17d) and [d2eb93c](https://github.com/NNPDF/eko/pull/160/commits/d2eb93cf8297b02a6061821630fdb6e908053a03) from here and put them in a separate PR - they're changing everything which makes this PR unreadable (and they got nothing to do with this - in fact, the tasks in here are already highly non-trivial on their own)

AleCandido:
> Can we please drop [ade4abf](https://github.com/NNPDF/eko/pull/160/commits/ade4abff9a5a17229bf68b9fe9278bec1a1ab17d) and [d2eb93c](https://github.com/NNPDF/eko/pull/160/commits/d2eb93cf8297b02a6061821630fdb6e908053a03) from here and put them in a separate PR - they're changing everything which makes this PR unreadable (and they got nothing to do with this - in fact, the tasks in here are already highly non-trivial on their own)

Indeed, I noticed and cherry-picked to develop

AleCandido:
Deprecated in favor of #172

PR #161, opened 2022-11-08
Fix struct.detached

andreab1997:
We want `output.struct.detached` to work even if the eko does not contain `inputpids`, `targetpids`, `inputgrid` and `targetgrid`, as this will be the case from now on. @felixhekhorn @AleCandido

### DIFF
--- a/src/eko/output/struct.py
+++ b/src/eko/output/struct.py
@@ -749,19 +749,12 @@ class EKO:
 
         """
         bases = operator["rotations"]
-        for basis in ("inputgrid", "targetgrid", "inputpids", "targetpids"):
-            if basis in bases:
-                bases[f"_{basis}"] = bases[basis]
-                del bases[basis]
         bases["pids"] = np.array(br.flavor_basis_pids)
-        for k in ("xgrid", "_inputgrid", "_targetgrid"):
-            if k in operator["rotations"]:
-                if operator["rotations"][k] is None:
-                    continue
-                bases[k] = interpolation.XGrid(
-                    operator["rotations"][k],
-                    log=operator["configs"]["interpolation_is_log"],
-                )
+        if operator["rotations"]["xgrid"] is not None:
+            bases["xgrid"] = interpolation.XGrid(
+                operator["rotations"]["xgrid"],
+                log=operator["configs"]["interpolation_is_log"],
+            )
 
         return cls(
             path=path,
--- a/src/eko/output/struct.py
+++ b/src/eko/output/struct.py
@@ -750,16 +750,18 @@ class EKO:
         """
         bases = operator["rotations"]
         for basis in ("inputgrid", "targetgrid", "inputpids", "targetpids"):
-            bases[f"_{basis}"] = bases[basis]
-            del bases[basis]
+            if basis in bases:
+                bases[f"_{basis}"] = bases[basis]
+                del bases[basis]
         bases["pids"] = np.array(br.flavor_basis_pids)
         for k in ("xgrid", "_inputgrid", "_targetgrid"):
-            if operator["rotations"][k] is None:
-                continue
-            bases[k] = interpolation.XGrid(
-                operator["rotations"][k],
-                log=operator["configs"]["interpolation_is_log"],
-            )
+            if k in operator["rotations"]:
+                if operator["rotations"][k] is None:
+                    continue
+                bases[k] = interpolation.XGrid(
+                    operator["rotations"][k],
+                    log=operator["configs"]["interpolation_is_log"],
+                )
 
         return cls(
             path=path,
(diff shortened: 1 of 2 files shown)


src/eko/output/struct.py
```diff
@@ -750,16 +750,18 @@ def detached(cls, theory: dict, operator: dict, path: pathlib.Path):
...
+                del bases[basis]
         bases["pids"] = np.array(br.flavor_basis_pids)
         for k in ("xgrid", "_inputgrid", "_targetgrid"):
-            if operator["rotations"][k] is None:
-                continue
-            bases[k] = interpolation.XGrid(
-                operator["rotations"][k],
-                log=operator["configs"]["interpolation_is_log"],
-            )
+            if k in operator["rotations"]:
```
felixhekhorn, line 758: let's avoid deep nesting so just switch to `not in` and `continue`

src/eko/output/struct.py
```diff
@@ -750,16 +750,18 @@ def detached(cls, theory: dict, operator: dict, path: pathlib.Path):
         """
         bases = operator["rotations"]
         for basis in ("inputgrid", "targetgrid", "inputpids", "targetpids"):
```
AleCandido, line 752: The operator card won't contain them any longer, you can drop the whole loop

src/eko/output/struct.py
```diff
@@ -750,16 +750,18 @@ def detached(cls, theory: dict, operator: dict, path: pathlib.Path):
...
-            bases[f"_{basis}"] = bases[basis]
-            del bases[basis]
+            if basis in bases:
+                bases[f"_{basis}"] = bases[basis]
+                del bases[basis]
         bases["pids"] = np.array(br.flavor_basis_pids)
         for k in ("xgrid", "_inputgrid", "_targetgrid"):
```
AleCandido, line 757: Dropping the former loop, also this loop has to be dropped

andreab1997: This is what I did before but then the tests are really a problem. Of course fixing the tests would be the correct solution but, since at the moment we are not using at all `rotations`, I believe that in the next future we need to get rid completely of it. This will be the proper change and at that moment we can drop these loops and fix the tests accordingly. 

BTW I am saying that we are not using at all `rotations` because even if I use `x_grid_reshape` and it changes `targetgrid`, then `load()` would completely ignore `targetgrid` and would set it to `xgrid` (wrongly because instead the operator is rotated). This is the problem that I have with pineko. So now `rotations` is not even keeping track of the rotations. It is completely useless.

AleCandido: > BTW I am saying that we are not using at all `rotations` because even if I use `x_grid_reshape` and it changes `targetgrid`, then `load()` would completely ignore `targetgrid` and would set it to `xgrid`

This should not happen: we want to "ignore" `targetgrid` (i.e. not putting it at all) only in the runcard. The reason why it is failing in Pineko is different: metadata are not automatically updated, and you are not doing it specifically.
Metadata are one thing, and runcards are another. Incidentally, they both have `rotations` - but we can drop `rotations` at all from the runcard, since we'll never use it any longer.

andreab1997:
So I removed the loops as agreed yesterday and started to fix the tests. However, as you can see, this is not as trivial as removing every mention (but the ones in `legacy`) of `targetgrid`, `inputgrid`, `inputpids` and `targetpids` in `conftest.py`. I will try to understand the situation better but any suggestion will be welcome :) @AleCandido @felixhekhorn

src/eko/output/struct.py
```diff
@@ -749,19 +749,12 @@ def detached(cls, theory: dict, operator: dict, path: pathlib.Path):
...
         bases["pids"] = np.array(br.flavor_basis_pids)
-        for k in ("xgrid", "_inputgrid", "_targetgrid"):
-            if k in operator["rotations"]:
-                if operator["rotations"][k] is None:
-                    continue
-                bases[k] = interpolation.XGrid(
-                    operator["rotations"][k],
-                    log=operator["configs"]["interpolation_is_log"],
-                )
+        if operator["rotations"]["xgrid"] is not None:
```
AleCandido, line 753: Actually, I'm not sure any longer with need this `if`: `xgrid` should not be allowed to be `None`.

Isn't it? @felixhekhorn

andreab1997:
I believe that the tests that we are not passing now can only be fixed after we allow for the `metadata` file we agreed yesterday. I am now opening that PR

AleCandido:
This PR updates `EKO.detached`, but in #172 the function has been killed completely, so the PR is already outdated.

I would close without merging.

AleCandido:
Yes, this is the first of the tower, since #161 -> #162 -> #171 -> #172.

So #172 already contains everything that is here, but dropped `EKO.detached`.
In a sense, I'm proposing to "merge forward" along the tower.

andreab1997:
> This PR updates `EKO.detached`, but in #172 the function has been killed completely, so the PR is already outdated.
> 
> I would close without merging.

Yes, I agree

PR #162, opened 2022-11-10
Output metadata

andreab1997:
We want to have a file called `metadata.yaml` in which we can store the relevant metadata for the operators as the `targetgrid`, the `inputgrid`, the `targetpids` and the `inputpids`. 

Moreover we want to move the `operator_card` and the `theory_card` inside a folder `log` in which we will also store all the logs of the functions acting on the operator, in order to ensure reproducibility. @felixhekhorn @AleCandido

### DIFF
--- a/benchmarks/ekobox/benchmark_evol_pdf.py
+++ b/benchmarks/ekobox/benchmark_evol_pdf.py
@@ -4,6 +4,7 @@ import numpy as np
 import pytest
 
 from eko import basis_rotation as br
+from eko import compatibility
 from ekobox import evol_pdf as ev_p
 from ekobox import operators_card as oc
 from ekobox import theory_card as tc
@@ -39,14 +40,15 @@ def benchmark_evolve_single_member(tmp_path, cd, lhapdf_path):
             "ktThr": 1,
         },
     )
+    nt, no = compatibility.update(theory, op)
     # lhapdf import (maybe i have to dump with a x*), do plots)
     with lhapdf_path(test_pdf):
         pdf = lhapdf.mkPDF("myCT14llo_NF3", 0)
     with cd(tmp_path):
         ev_p.evolve_pdfs(
             [pdf],
-            theory,
-            op,
+            nt,
+            no,
             name="EvPDF",
             info_update={"SetDesc": "MyEvolvedPDF", "MZ": 0.2, "Debug": "Debug"},
         )
@@ -54,11 +56,11 @@ def benchmark_evolve_single_member(tmp_path, cd, lhapdf_path):
         all_blocks = (load.load_blocks_from_file("EvPDF", 0))[1]
         info = load.load_info_from_file("EvPDF")
         ev_pdf = lhapdf.mkPDF("EvPDF", 0)
-    assert info["XMin"] == op["interpolation_xgrid"][0]
+    assert info["XMin"] == no["rotations"]["xgrid"][0]
     assert info["SetDesc"] == "MyEvolvedPDF"
-    assert info["MZ"] == theory["MZ"]
+    assert info["MZ"] == nt["MZ"]
     assert info["Debug"] == "Debug"
-    xgrid = op["interpolation_xgrid"]
+    xgrid = no["rotations"]["xgrid"]
     for Q2 in [20.0, 100.0, 10000.0]:
         for x in xgrid[10:40]:
             for pid in [21, 1, -1, 2, -2, 3, -3]:
@@ -82,19 +84,20 @@ def benchmark_evolve_more_members(tmp_path, cd, lhapdf_path):
         [10, 100], update={"interpolation_xgrid": [1e-7, 0.01, 0.1, 0.2, 0.3]}
     )
     theory = tc.generate(0, 1.0)
+    nt, no = compatibility.update(theory, op)
     with lhapdf_path(test_pdf):
         pdfs = lhapdf.mkPDFs("myMSTW2008nlo90cl")
     d = tmp_path / "sub"
     d.mkdir()
     with lhapdf_path(d):
         with cd(tmp_path):
-            ev_p.evolve_pdfs(pdfs, theory, op, install=True, name="Debug")
+            ev_p.evolve_pdfs(pdfs, nt, no, install=True, name="Debug")
         # ev_pdfs
         new_pdfs = lhapdf.mkPDFs("Debug")
         new_pdf_1 = lhapdf.mkPDF("Debug", 0)
         new_pdf_2 = lhapdf.mkPDF("Debug", 1)
         info = load.load_info_from_file("Debug")
-    assert info["XMin"] == op["interpolation_xgrid"][0]
+    assert info["XMin"] == no["rotations"]["xgrid"][0]
     assert len(pdfs) == len(new_pdfs)
     for Q2 in [10, 100]:
         for x in [1e-7, 0.01, 0.1, 0.2, 0.3]:
--- a/src/eko/output/manipulate.py
+++ b/src/eko/output/manipulate.py
@@ -139,18 +139,14 @@ def flavor_reshape(
         errs = elem.error
         if targetpids is not None and inputpids is None:
             # update metadata
-            eko.update_metadata({"rotations": {"_targetpids": targetpids}})
             ops = np.einsum("ca,ajbk->cjbk", targetpids, ops)
             errs = np.einsum("ca,ajbk->cjbk", targetpids, errs)
         elif inputpids is not None and targetpids is None:
             # update metadata
-            eko.update_metadata({"rotations": {"_inputpids": inputpids}})
             ops = np.einsum("ajbk,bd->ajdk", ops, inv_inputpids)
             errs = np.einsum("ajbk,bd->ajdk", errs, inv_inputpids)
         else:
             # update metadata
-            eko.update_metadata({"rotations": {"_inputpids": inputpids}})
-            eko.update_metadata({"rotations": {"_targetpids": targetpids}})
             ops = np.einsum("ca,ajbk,bd->cjdk", targetpids, ops, inv_inputpids)
             errs = np.einsum("ca,ajbk,bd->cjdk", targetpids, errs, inv_inputpids)
         elem.operator = ops
@@ -162,8 +158,10 @@ def flavor_reshape(
     # there is no meaningful way to set them in general, after rotation
     if inputpids is not None:
         eko.rotations._inputpids = np.array([0] * len(eko.rotations.inputpids))
+        eko.update_metadata({"rotations": {"_inputpids": inputpids}})
     if targetpids is not None:
         eko.rotations._targetpids = np.array([0] * len(eko.rotations.targetpids))
+        eko.update_metadata({"rotations": {"_targetpids": targetpids}})
 
 
 def to_evol(eko: EKO, source: bool = True, target: bool = False):
--- a/src/eko/output/manipulate.py
+++ b/src/eko/output/manipulate.py
@@ -138,15 +138,12 @@ def flavor_reshape(
         ops = elem.operator
         errs = elem.error
         if targetpids is not None and inputpids is None:
-            # update metadata
             ops = np.einsum("ca,ajbk->cjbk", targetpids, ops)
             errs = np.einsum("ca,ajbk->cjbk", targetpids, errs)
         elif inputpids is not None and targetpids is None:
-            # update metadata
             ops = np.einsum("ajbk,bd->ajdk", ops, inv_inputpids)
             errs = np.einsum("ajbk,bd->ajdk", errs, inv_inputpids)
         else:
-            # update metadata
             ops = np.einsum("ca,ajbk,bd->cjdk", targetpids, ops, inv_inputpids)
             errs = np.einsum("ca,ajbk,bd->cjdk", targetpids, errs, inv_inputpids)
         elem.operator = ops
@@ -158,10 +155,10 @@ def flavor_reshape(
     # there is no meaningful way to set them in general, after rotation
     if inputpids is not None:
         eko.rotations._inputpids = np.array([0] * len(eko.rotations.inputpids))
-        eko.update_metadata({"rotations": {"_inputpids": inputpids}})
+        eko.update_metadata({"rotations": {"_inputpids": eko.rotations._inputpids}})
     if targetpids is not None:
         eko.rotations._targetpids = np.array([0] * len(eko.rotations.targetpids))
-        eko.update_metadata({"rotations": {"_targetpids": targetpids}})
+        eko.update_metadata({"rotations": {"_targetpids": eko.rotations._targetpids}})
 
 
 def to_evol(eko: EKO, source: bool = True, target: bool = False):
@@ -183,10 +180,10 @@ def to_evol(eko: EKO, source: bool = True, target: bool = False):
     flavor_reshape(eko, inputpids=inputpids, targetpids=targetpids)
     # assign pids
     if source:
-        eko.rotations._inputpids = br.evol_basis_pids
+        eko.rotations._inputpids = inputpids
         # update metadata
-        eko.update_metadata({"rotations": {"_inputpids": br.evol_basis_pids}})
+        eko.update_metadata({"rotations": {"_inputpids": inputpids}})
     if target:
-        eko.rotations._targetpids = br.evol_basis_pids
+        eko.rotations._targetpids = targetpids
         # update metadata
-        eko.update_metadata({"rotations": {"_targetpids": br.evol_basis_pids}})
+        eko.update_metadata({"rotations": {"_targetpids": targetpids}})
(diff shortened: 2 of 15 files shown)


src/eko/output/manipulate.py
```diff
@@ -134,12 +138,19 @@ def flavor_reshape(
...
             ops = np.einsum("ca,ajbk->cjbk", targetpids, ops)
             errs = np.einsum("ca,ajbk->cjbk", targetpids, errs)
         elif inputpids is not None and targetpids is None:
+            # update metadata
+            eko.update_metadata({"rotations": {"inputpids": inputpids}})
             ops = np.einsum("ajbk,bd->ajdk", ops, inv_inputpids)
             errs = np.einsum("ajbk,bd->ajdk", errs, inv_inputpids)
         else:
+            # update metadata
+            eko.update_metadata({"rotations": {"inputpids": inputpids}})
```
felixhekhorn, line 152: the call should move down to were the write is happening (else we get out of sync) .. note that we're casting to list (see [comment below](https://github.com/NNPDF/eko/pull/162/files/99ee17fdfa0cd5c3565442aee2d6323c9d339b1b..a85fd194824bc4f195a4250682dca0b18acdec3f#r1021860266) )

src/eko/output/manipulate.py
```diff
@@ -175,5 +186,9 @@ def to_evol(eko: EKO, source: bool = True, target: bool = False):
     # assign pids
     if source:
         eko.rotations._inputpids = br.evol_basis_pids
+        # update metadata
+        eko.update_metadata({"rotations": {"inputpids": inputpids}})
```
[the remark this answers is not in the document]
andreab1997: At this point I believe that in any case `rotations.inputpids` and `rotations.targetpids` should be matrices, not only for `to_evol`. The question is, do we want to store the matrix itself inside the file `metadata` or do we want to store the vector and then internally change the object (for example `rotations.targetpids`)? In the second case we can add something in `__post__init__`: Since we have both the vectors `rotations.pids` and `rotations.targetpids` as readed from `metadata` we can just compute the matrix and assign it to `rotations.targetgrid` in place of the vector. In this way, we will only store vectors, which I believe are more clear to read for an user, but then internally we have the matrices we need. Do you agree?

andreab1997:
Some of the tests we are not passing now are failing because the `metadata.yaml` file is wrongly dumped. However, it is dumped in the exact same way as the `operator_card.yaml` and indeed even the latter is not working (if you take an eko and do `eko.operator_card()` it will fail). The problem seems to be related to the fact that `yaml.dump()` cannot dump numpy arrays. Were you aware of this problem? @felixhekhorn @AleCandido

AleCandido:
> Some of the tests we are not passing now are failing because the `metadata.yaml` file is wrongly dumped. However, it is dumped in the exact same way as the `operator_card.yaml` and indeed even the latter is not working (if you take an eko and do `eko.operator_card()` it will fail). The problem seems to be related to the fact that `yaml.dump()` cannot dump numpy arrays. Were you aware of this problem? @felixhekhorn @AleCandido

Problem is not dumping, but loading: if you dump them as Numpy arrays, it will (since YAML can dump custom objects). But then when you load the "custom object" is not matching any longer (because of the scope and stuffs like this), thus it will fail.

In any case, we want basic YAML, not custom objects, so you want to convert any array in a list of lists of lists...
https://github.com/NNPDF/eko/blob/dcb40cb6d3385bfe864615905db3537a2daee661/src/eko/output/struct.py#L89-L90

andreab1997:
> > Some of the tests we are not passing now are failing because the `metadata.yaml` file is wrongly dumped. However, it is dumped in the exact same way as the `operator_card.yaml` and indeed even the latter is not working (if you take an eko and do `eko.operator_card()` it will fail). The problem seems to be related to the fact that `yaml.dump()` cannot dump numpy arrays. Were you aware of this problem? @felixhekhorn @AleCandido
> 
> Problem is not dumping, but loading: if you dump them as Numpy arrays, it will (since YAML can dump custom objects). But then when you load the "custom object" is not matching any longer (because of the scope and stuffs like this), thus it will fail.
> 
> In any case, we want basic YAML, not custom objects, so you want to convert any array in a list of lists of lists...
> 
> https://github.com/NNPDF/eko/blob/dcb40cb6d3385bfe864615905db3537a2daee661/src/eko/output/struct.py#L89-L90

Yes indeed, this is what I am doing :)

AleCandido:
> Yes indeed, this is what I am doing :)

If you use `.raw` it should automatically do it, exactly for that piece of code I quoted.

andreab1997:
The solution that I pushed is very bad but I needed something working in order to test the rest. I will push a better solution soon. However, it seems that there is no way using the `tarfile` python package to add or modify a single file inside a tar file. So, in order to do that for metadata, I believe we need to write a function that creates another tar file with the same name and with the updated metadata file. Do you have better solutions?

src/eko/output/manipulate.py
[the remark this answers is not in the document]
AleCandido: As we discussed yesterday, we want the matrix: it is uniform, also for non-standard bases, and always reliable.
The less special cases we have, the easier will be the code, and maintainability will improve.

src/eko/output/manipulate.py
AleCandido: This made sense: we want operations to be as atomic as possible (in the ACID sense)

tests/ekomark/test_apply.py
```diff
@@ -11,7 +11,7 @@ def test_apply(self, fake_legacy, fake_pdf):
         pdf_grid = apply.apply_pdf(o, fake_pdf)
         assert len(pdf_grid) == len(fake_card["Q2grid"])
         pdfs = pdf_grid[q2_out]["pdfs"]
-        assert list(pdfs.keys()) == o.rotations.targetpids
+        assert list(pdfs.keys()) == list(o.rotations.targetpids)
```
AleCandido, line 14: Just in case, you can also assert they are the same despite the iterables even without converting both to lists:
```suggestion
        assert all(x == y for x, y in zip(pdfs.keys(), o.rotations.targetpids))
```

No special advantage here (you save some memory and time, but it is definitely negligible here).

AleCandido: I take it back: it has the disadvantage of not checking all the elements in the longest iterable, in case the length is different, so you'd need to split into a separate check.
```suggestion
        assert len(pdfs.keys()) == len(o.rotations.targetpids))
        assert all(x == y for x, y in zip(pdfs.keys(), o.rotations.targetpids))
```

Either we write a function for this, to consistently use, or it is only more cumbersome (for a negligible advantage).

AleCandido: Once more (I believe I told you somewhere else), `.keys()` is not required when you iterate on it, and `list()` is iterating.

andreab1997:
So now the tests are working again. This does not mean that we can merge unfortunately. I still need to do a couple of things:

- [ ] update the metadata file after any rotation
- [ ] Read the metadata file in order to assign the correct value to `eko.rotations` when an eko is loaded. 
- [ ] Construct the `operator_card` class in order to use its method `raw` before dumping 

If you believe there is something else, please add here.

tests/ekomark/test_apply.py
andreab1997: Yes i know but for the moment I wanted to keep the changes to the tests as minimal as possible. I will drop `.keys` now. :)

andreab1997:
@AleCandido @felixhekhorn This should now be roughly what we want. I did not tick the second point of the todo list ("Read the metadata file in order to assign the correct value to eko.rotations when an eko is loaded.") yet because I am still checking if everything is correct but I believe it is already working. Then I will only need to change the name of "inputpids" to something more telling but that should be trivial. Let me know if you have suggestions.

PS: At the end (thanks to @AleCandido's help) I managed to update the metadata file without the other PR (#152)

benchmarks/ekobox/benchmark_evol_pdf.py
```diff
@@ -39,26 +40,27 @@ def benchmark_evolve_single_member(tmp_path, cd, lhapdf_path):
...
     # lhapdf import (maybe i have to dump with a x*), do plots)
     with lhapdf_path(test_pdf):
         pdf = lhapdf.mkPDF("myCT14llo_NF3", 0)
     with cd(tmp_path):
         ev_p.evolve_pdfs(
             [pdf],
-            theory,
-            op,
+            nt,
+            no,
```
AleCandido, line 51: I know they mean `new_theory_card` and `new_operator_card`, but it is a bit cryptic.
Please, favor more telling names. Trade-offs are best:

- ❌ `nt` - too short, ambiguous
- ❌ `new_theory_card_evolve_single_member` - redundant, it is already in the scope of `evolve_single_member`
- ✔️ `new_theory_card` - just perfect :)

The minimal amount of information that saves you an explicit comment is the optimal choice, according to me.

src/eko/output/manipulate.py
```diff
@@ -61,6 +61,8 @@ def xgrid_reshape(
         )
         target_rot = b.get_interpolation(targetgrid.raw)
         eko.rotations._targetgrid = targetgrid
+        # update metadata
+        eko.update_metadata({"rotations": {"_targetgrid": targetgrid}})
```
AleCandido, line 65: Are you sure you need the underscore?

I know the attribute is that one, the other being a property, but it'd look nicer without, for manual inspection.
At the end of the day, the initial underscore discriminate between the virtual property, and the actual representation, but when we serialize there is only the representation and nothing more.

src/eko/output/struct.py
```diff
@@ -91,7 +92,7 @@ def raw(self):
             elif isinstance(value, float):
                 value = float(value)
             elif isinstance(value, interpolation.XGrid):
-                value = value.dump()
+                value = value.dump()["grid"]
```
AleCandido, line 95: Uhm, this is weird: the serialization of `XGrid` should be fully managed by the class itself.

So, we should always save what `.dump()` returns, and `XGrid` itself should be able to be loaded from the whole return value of a `.dump()`.
If this is not happening, we should fix on `XGrid` side, not here.

src/eko/output/struct.py
```diff
@@ -253,7 +254,7 @@ def __post_init__(self):
             value = getattr(self, attr)
             if value is None:
                 continue
-            if isinstance(value, np.ndarray):
+            if isinstance(value, np.ndarray) or isinstance(value, list):
```
AleCandido, line 257: Most likely you need `list` because you changed above, and from serialization it will be a `dict`, but it is not a bad idea to keep.

However, you don't need two calls to `isinstance`:
```suggestion
            if isinstance(value, (np.ndarray, list)):
```
https://docs.python.org/3/library/functions.html#isinstance

src/ekobox/evol_pdf.py
```diff
@@ -111,4 +113,7 @@ def ekofileid(theory_card, operators_card):
     str
         file name
     """
-    return f"o{operators_card['hash'][:6]}_t{theory_card['hash'][:6]}.tar"
+    try:
+        return f"o{operators_card['hash'][:6]}_t{theory_card['hash'][:6]}.tar"
+    except KeyError:
+        return "o000000_t000000.tar"
```
AleCandido, line 119: Wait, where does this hard-coded string come from?

AleCandido: Maybe re-raising a more explicit error would be more appropriate...

tests/eko/test_output_struct.py
```diff
@@ -130,7 +130,9 @@ class TestEKO:
     def _default_cards(self):
         t = tc.generate(0, 1.0)
         o = oc.generate([10.0])
-        return compatibility.update(t, o)
+        nt, no = compatibility.update(t, o)
+        no["rotations"]["pids"] = no["rotations"]["targetpids"]
```
AleCandido, line 134: Shouldn't we put this in `compatibility.update`?

AleCandido (changes requested):

benchmarks/ekobox/benchmark_evol_pdf.py
andreab1997: I actually copied the name that was already there. Should I change also the others `nt`? (and similar for `no`)

src/eko/output/manipulate.py
andreab1997: The point is that at the end the metadata file should be read and used to construct a `Rotations` object. In order to do that it expects `_targetgrid` and similar.

benchmarks/ekobox/benchmark_evol_pdf.py
AleCandido: Not top priority, but I believe we should constantly improve also these details, they make it simpler for new developers and ourselves to maintain.

If you can improve something is already better than nothing, as I said it is not high-priority, don't waste too much time. But at least change this one, such that the new code has better and better style :)

src/eko/output/struct.py
andreab1997: Yes indeed, the point is that `dump()` return a dictionary like {'grid':[...], 'log':[...]} where the actual grid is inside `grid`. For the tests, sometimes the `log` part is used and sometimes it is not (and this is causing errors). So I was not sure on how to fix

src/eko/output/struct.py
andreab1997: Ok thank you :)

src/ekobox/evol_pdf.py
andreab1997: The problem is that in the new format we don't have `hash` anymore. So, even in this case, we need to choose how to fix. This was a temporary solution just to have tests working but of course we need to make a decision.

tests/eko/test_output_struct.py
andreab1997: This I don't know, I believe it is something very specific to this test. Most of the case we do not need this because `[rotations][pids]` will be already there

src/eko/output/manipulate.py
AleCandido: Indeed, I was proposing to change what `Rotations` expects, but this we can discuss together, it's easier :)

src/eko/output/struct.py
AleCandido: We should make sure that the `log` part is always used, or ignored, on the consumer side

tests/eko/test_output_struct.py
AleCandido: Ok, then maybe upgrade default cards, such that we don't need it also here :)

In the end, default cards are for the users (as everything else in `ekobox`), we don't want to generate incompatible ones :)

felixhekhorn:
Benchmarks are broken:
```
$ poe lha
Poe => python benchmarks/lha_paper_bench.py
 ──────────────────────────────────────── 
  Theories: 1 OCards: 1 PDFs: 1 ext: LHA  
 ──────────────────────────────────────── 
Computing for theory=fb3f558, ocard=7f27498 and pdf=ToyLH ...
  File "/home/felix/Physik/N3PDF/EKO/eko/benchmarks/lha_paper_bench.py", line 226, in <module>
    obj.benchmark_plain(0)
  File "/home/felix/Physik/N3PDF/EKO/eko/benchmarks/lha_paper_bench.py", line 123, in benchmark_plain
    self.run_lha(self.plain_theory(pto))
  File "/home/felix/Physik/N3PDF/EKO/eko/benchmarks/lha_paper_bench.py", line 109, in run_lha
    self.run(
  File "/home/felix/.cache/pypoetry/virtualenvs/eko-KkPVjVhh-py3.10/lib/python3.10/site-packages/banana/benchmark/runner.py", line 416, in run
    self.run_config(session, t, o, pdf_name, use_replicas)
  File "/home/felix/.cache/pypoetry/virtualenvs/eko-KkPVjVhh-py3.10/lib/python3.10/site-packages/banana/benchmark/runner.py", line 304, in run_config
    me = self.run_me(t, o, pdf)
  File "/home/felix/Physik/N3PDF/EKO/eko/src/ekomark/benchmark/runner.py", line 119, in run_me
    out = eko.run_dglap(theory, ocard)
  File "/home/felix/Physik/N3PDF/EKO/eko/src/eko/__init__.py", line 29, in run_dglap
    r = runner.Runner(theory_card, operators_card)
  File "/home/felix/Physik/N3PDF/EKO/eko/src/eko/runner.py", line 50, in __init__
    new_theory, new_operators = compatibility.update(theory_card, operators_card)
  File "/home/felix/Physik/N3PDF/EKO/eko/src/eko/compatibility.py", line 57, in update
    new_operators["rotations"]["pids"] = operators["pids"]
KeyError: 'pids'
```

felixhekhorn:
you also need to dump version informations (i.e. both program and data) to the metadata (else we get the [usual problems](https://github.com/NNPDF/pineko/issues/64#issuecomment-1339422849)) (here the additional scope we added already becomes handy) . This is also relevant for a converter (e.g. the one in #171).

AleCandido:
> you also need to dump version informations (i.e. both program and data) to the metadata

Don't mind: this I'm doing in #172

PR #170, opened 2022-11-26
LO polarised anomalous dimension

giacomomagni:
This PR is for the implementation of LO polarised anomalous dimension.


AleCandido:
Why did you put it in a new package?

Since the will share almost everything with `eko`, apart from the anomalous dimensions themselves, I would put together everything else:
```
eko
|- anomalous_dimensions/
| |- unpolarized/
| | |- ...
| |- polarized/
| | |- ...
```

giacomomagni:
> Why did you put it in a new package?
> 
 Hi @AleCandido, I think this was done by mistake. 
 This PR is just to keep track of the work done by @adrianneschauss, but for the moment it's just a draft. 
 I've already warned her about a couple of good practice tips.

AleCandido:
Ok, I guess you can take care of it: I know it was draft, but I didn't want the final product to go in that direction (let's say that I've been rather astonished when I saw 87 files changed...).

Sorry if I stepped in to early...
If you wish, I can just review the final product :)

AleCandido:
Thank you, @adrianneschauss, for solving the major issue.

About the new modules `asp1.py` and `apem1.py`:
- naming is currently inconsistent, use:
    1. `asp` and `aemp` (preferred, by me), or 
    2. `aps` and `apem`
- in any case, I would drop the `p` from the module, and just use the organization proposed [above](https://github.com/NNPDF/eko/pull/170#issuecomment-1328060169), with a further subdivision in folders between polarized and unpolarized

The idea is that just relying on a `p` is a bit confusing, and:
> Explicit is better than implicit.

However, it is definitely not urgent, please take your time :)

felixhekhorn:
> and just use the organization proposed [above](https://github.com/NNPDF/eko/pull/170#issuecomment-1328060169), with a further subdivision in folders between polarized and unpolarized

definitely :upside_down_face: remember that @t7phy is working in parallel on the time-like unpolarized part  - so you should move into `polarized/space_like/*` and @t7phy into `unpolarized/time_like/*` (later we need also to move the current content to `unpolarized/space_like`)

src/eko/evolution_operator/__init__.py
```diff
@@ -11,15 +11,14 @@
...
-from scipy import integrate
-
 from .. import anomalous_dimensions as ad
 from .. import basis_rotation as br
 from .. import interpolation, mellin
 from .. import scale_variations as sv
 from ..kernels import non_singlet as ns
 from ..kernels import singlet as s
 from ..member import OpMember
+from scipy import integrate
```
AleCandido, line 21: `pre-commit` (i.e. `isort`) should put it back in the former place, please run:
```sh
pre-commit install
```
in the local instance of your repository.

src/eko/evolution_operator/__init__.py
```diff
@@ -84,7 +83,7 @@ def path(self):
 
     @property
     def n(self):
-        """Returs the Mellin moment :math:`N`."""
+        """Returns the Mellin moment :math:`N`."""
```
AleCandido, line 86:
```suggestion
        """Return the Mellin moment :math:`N`."""
```

src/eko/evolution_operator/__init__.py
```diff
@@ -159,6 +159,8 @@ def quad_ker(
         coupling value at the process scale
     nf : int
         number of active flavors
+    p : Boolean
+        Polarised (True) or un-Polarised (False)
```
AleCandido, line 163: Unfortunately I believe that introducing a boolean is the correct way at the moment (or equally bad  to introducing a parallel dispatcher), but eventually I believe we want to get rid of dispatchers at all, at least not keep using them as they are meant now (as a single entry point for Numba).

This will be addressed by refactoring after the analogue of #123 will be implemnted.

tests/eko/test_ad.py
```diff
@@ -54,42 +55,44 @@ def test_eigensystem_gamma_singlet_projectors_EV():
 
 def test_gamma_ns():
     nf = 3
+    p = False
```
AleCandido, line 58: As already noticed in #184 these tests contains a lot of repetitions. We want to refactor limiting the amount of repetitions as you would for normal code.

E.g. here you can do it by looping over orders and outputs, without repeating very similar calls over and over.

src/eko/anomalous_dimensions/asp1.py
```diff
@@ -0,0 +1,129 @@
...
+      N : complex
+        Mellin moment
+      nf : int
+        Number of active flavors
+
+    Returns
+    -------
+      gamma_qg : complex
+        Leading-order polarised quark-gluon anomalous dimension :math:`\\gamma_{qg}^{(0)}(N)`
+    """
```
AleCandido, line 37: This is more targeted to @felixhekhorn: I propose to drop `Parameters` and `Returns` sections, since they are quite verbose and redundant.
We will replace by the specification of a common interface (or a scheme for possible interfaces) in the module/subpackage docstrings.

With "interface" I mean the io = parameters + returns

tests/eko/test_ad_asp1.py
```diff
@@ -0,0 +1,100 @@
...
+# print(l)
+# select_el= ker_s[k]
+# print (select_el)
+
+# spec = [
+#     ("is_singlet", nb.boolean),
+#     ("is_log", nb.boolean),
+#     ("logx", nb.float64),
+#     ("u", nb.float64),
+# ]
```
AleCandido, line 99: I guess this has been copied from somewhere.

We can drop here, and in case also in the place this was taken from.

tests/eko/test_ev_operator.py
```diff
@@ -41,6 +41,7 @@ def test_quad_ker(monkeypatch):
             as0=2,
             as_raw=1,
             nf=3,
+            p=False,
```
AleCandido, line 44: Since the fix is repeated, you see the pattern. If we see the pattern, let's refactor this @felixhekhorn

README.md
```diff
@@ -62,3 +62,5 @@ When using our code please cite
 - Your feedback is welcome! If you want to report a (possible) bug or want to ask for a new feature, please raise an issue: <a href="https://github.com/N3PDF/eko/issues"><img alt="GitHub issues" src="https://img.shields.io/github/issues/N3PDF/eko"/></a>
 - Please follow our [Code of Conduct](https://github.com/N3PDF/eko/blob/master/.github/CODE_OF_CONDUCT.md) and read the
   [Contribution Guidelines](https://github.com/N3PDF/eko/blob/master/.github/CONTRIBUTING.md)
+
+##Test
```
AleCandido, line 66:
```suggestion
```

No need for this section, I guess it is just a leftover :)

.gitignore
```diff
@@ -4,7 +4,7 @@ __pycache__/
 *$py.class
 
 # C extensions
-*.so
+*.so
```
AleCandido, line 7:
```suggestion
*.so
```

No need for this change, I guess it is another leftover :)

src/eko/anomalous_dimensions/aem1.py
```diff
@@ -6,6 +6,7 @@
 
 from .. import constants
 from . import as1
+from . import asp1
```
AleCandido, line 9: About this I would organize differently: a different subpackage rather than importing in the same. I would reconnect at the level of integration, in which we will switch between pol and unpol ad (either we compute the operator for one or for the other, the two are not connected).

AleCandido (changes requested):

AleCandido:
The style is rather similar to the current one of EKO, and generally good. Most possible improvements are shared with the rest of the package, apart from negotiating the structure of this new part.

src/eko/anomalous_dimensions/asp1.py
felixhekhorn: this is not so trivial, since many of the higher order functions require a non-trivial combination of S-es

AleCandido: But they were getting access to an S-cache, isn't it?

AleCandido: However, in case they are variables, we can agree on a naming scheme for them. Once they adhere to the naming scheme, they do not need to be documented one by one.

giacomomagni:
clesed in favor of #195

PR #171, opened 2022-12-03
Reorganize CLI

AleCandido:
The EKO CLI was already initialized, but completely forgotten and unused.

I will add some basic commands to generate EKOs, useful to test further developments.
- [x] conversion subcommand

Issue #169, opened 2022-11-24
Undocumented usage of banana in the tutorial


lukeronger:
I tried to use the EKO package for some tests.

However, these is an issue, which I could not understand, which is shown below:

```
    from banana.data.theories import default_card as th_card
ModuleNotFoundError: No module named 'banana.data'
```

Actually I have installed banana successfully,  with `pip3 install banana`.
Thanks a lot!

**My simple code is shown below:**
```py
import eko

from banana.data.theories import default_card as th_card
from ekomark.data.operators import default_card as op_card

# here we replace the grid with a very minimal one, to speed up the example
op_card["interpolation_xgrid"] = [1e-4, 2e-4, 4e-4, 8e-4, 1e-3, 2e-3, 4e-3, 8e-3, 1e-2, 2e-2, 4e-2, 8e-2, 1e-1, 2e-1, 4e-1, 8e-1, 1.]
# let's compute two operators: one reaching up to Q2=100 GeV^2 and a second reaching up to Q2=10000 GeV^2
op_card["Q2grid"] = [10., 100., 1000., 10000.]

evolution_operator = eko.run_dglap(th_card, op_card)

from banana import toy
pdf = toy.mkPDF("", 0)

from ekomark.apply import apply_pdf
evolved_pdfs = apply_pdf(evolution_operator, pdf)

print("key  ",  evolved_pdfs.keys())
#dict_keys([100.0, 10000.0])
print("gluon  ", evolved_pdfs[100.0]["pdfs"][21])
```

AleCandido:
Follows the reply originally sent by email.

---
 Dear Rong Wang,

I'm very pleased you tried to use our program, and I'm sorry it didn't work out of the box.

Unfortunately, we are still work in progress on some new developments, since we noticed that computing very big operators can rapidly fill your memory.
For this reason, I'd suggest to you to use an EKO version in the range 0.8.5 (the one used for the paper) - 0.10.2 (last one not affected by more recent developments).
We hope to have a new big release with new features in the next few months (and probably it will be 1.0.0).

Concerning your issue: banana is not required in general to run EKO, since it is the tool we use for internal benchmarking.
Nevertheless, EKO requires you to provide complete runcards, no default is applied. In case you're just familiarizing with it, banana has some sensible defaults to get started, even though you should not use as "recommended settings". In case you're in doubt for some parameters values, get in touch with us, and we can suggest the most appropriate to your use case.

[...]

PR #148, opened 2022-09-13
N3LO ad variations

giacomomagni:
N3LO anomalous dimension variations. 

* We will provide around 20 parametrisation for each singlet ad entry. The "best" N3LO ad will be the average of the variations `(0,0,0,0)`.
* A new key should be added to the theory card (```n3lo_ad_variation: (gg_var,gq_var, qg_var,qq_var)```)
* We will not introduce any variation in the non singlet ad.

A new version of banana is required for the test to pass.

PR #172, opened 2022-12-05
Use tempdir during execution

alecandido:
New implementation of #152, adapting some content from there.

- [x] propagate changes to the rest of the package
- [x] fix unit tests
- [x] fix unit test on py3.8
- [x] fix log-less isolated benchmarks
- [x] check still to accomplish the goal of #162
- [x] https://github.com/NNPDF/eko/pull/171#pullrequestreview-1206431139
- [x] https://github.com/NNPDF/eko/pull/171#issuecomment-1339206406
- [x] https://github.com/NNPDF/eko/pull/154 + https://github.com/NNPDF/eko/blob/20773728c10414e4b5ca4369dddd5f163e49d89d/src/eko/output/struct.py#L622-L634
- [x] import and test [old tar output](https://github.com/NNPDF/eko/blob/0b5372e8fed9a730a98d5f61d636e938da9a3eaa/tests/ekobox/v0.8.5-obf24af_t8e1305.tar)
- [x] scope `num_flavs_ref` and `num_flavs_max_as` in `Couplings` structure
- [x] prevent operations on a closed EKO
- [x] prevent modifications on a read-only EKO
- [x] more Pythonic `lambergrid` name https://github.com/NNPDF/eko/pull/172#discussion_r1051017406
- [x] test metadata presence https://github.com/NNPDF/eko/pull/172#discussion_r1048463218
- [x] remove defaults https://github.com/NNPDF/eko/pull/172#discussion_r1054262713
- [x] change `num_flavs_max_as` to a less redundant one https://github.com/NNPDF/eko/pull/172#discussion_r1051984942
- [x] update docstring https://github.com/NNPDF/eko/pull/172#discussion_r1051993427
- [x] fix tutorials https://github.com/NNPDF/eko/pull/172#pullrequestreview-1227844530
- [x] check $\alpha_s$ https://github.com/NNPDF/eko/pull/172#issuecomment-1364060491
- [ ] recover 100% coverage

### DIFF
--- a/src/eko/io/struct.py
+++ b/src/eko/io/struct.py
@@ -1006,9 +1006,9 @@ class Builder:
     def __post_init__(self):
         """Validate paths."""
         if self.access.path.suffix != ".tar":
-            raise exceptions.OutputNotTar(self.path)
+            raise exceptions.OutputNotTar(self.access.path)
         if self.access.path.exists():
-            raise exceptions.OutputExistsError(self.path)
+            raise exceptions.OutputExistsError(self.access.path)
 
     def load_cards(self, theory: TheoryCard, operator: OperatorCard):
         """Load both theory and operator card."""
--- a/tests/eko/io/test_struct.py
+++ b/tests/eko/io/test_struct.py
@@ -133,13 +133,13 @@ class TestEKO:
         # try to write to a file different from bla
         no_tar_path = tmp_path / "Blub.bla"
         no_tar_path.touch()
-        with pytest.raises(ValueError):
+        with pytest.raises(ValueError, match="Blub.bla"):
             struct.EKO.create(no_tar_path)
         # try to overwrite an existing file
         exists_path = tmp_path / "Blub.tar"
         with tarfile.open(exists_path, "w") as tar:
             tar.add(no_tar_path)
-        with pytest.raises(FileExistsError):
+        with pytest.raises(FileExistsError, match="Blub.tar"):
             struct.EKO.create(exists_path)
 
     def test_load_error(self, tmp_path):


AleCandido:
Ok, I completed the massive rework of the structure, now I have to fix all tests and polish everything.

I will add the to-do list in the original message.

AleCandido:
Ok, now this implements also the changes that were scheduled for #160.

I'm at the level of fixing tests. I'm not sure if I should target coverage in this PR though, I fear it'll become to big...

src/eko/runner/legacy.py
```diff
@@ -26,18 +25,10 @@ class Runner:
...
- 888oooo8     88888     \\\\/8/////   888
- 888    "     888`88b.      888 ///   888
- 888       o  888  `88b.    `88b //  d88'
-o888ooooood8 o888o  o888o     `Y8bood8P'
-"""
+    banner = commons.BANNER
 
     def __init__(self, theory_card: dict, operators_card: dict):
-        """Initialize runner.
+        """Initialie runner.
```
felixhekhorn, line 31:
```suggestion
        """Initialize runner.
```

AleCandido:
I still have 38 tests failing, mostly for `AttributeError`s, that are generated by the reshuffling.

At this point, features should be all available, and this PR should also close #114

AleCandido:
At this point, I decided that it is easiest to merge just this one, closing #171 #162 #161 

This branch contains all of them, and it is already big enough that it doesn't change so much, including the other ones.
Moreover, a lot of changes there are partially overwritten here.

AleCandido:
15 + 8 tests failing, I'm getting there...

AleCandido:
Since now we have open (with dir available) and closed (just a path) operators, a question arises.

When we compute an EKO, should we return an open one (and leave the responsibility to close it to a user that never opened it), or a closed one (and the user have to load it to use)?

AleCandido:
Last 3 unit tests missing, but for them I will take slightly more, since I want to briefly rework the legacy output.

In particular, I will drop `load_yaml` and `dump_yaml` (never used yaml format in production, so much better to get rid of one extra unused format) and `dump_tar`, since we will never produce new EKO in the old format.
Instead, I will make sure `load_tar` is actually working with the new struct, since this is the basic ingredient for the converter to the new format.

src/eko/runner/legacy.py
```diff
@@ -0,0 +1,124 @@
...
+        # strong coupling
+        sc = Couplings(
+            couplings_ref=np.array(new_theory.couplings.values),
+            scale_ref=new_theory.couplings.alphas.scale**2,
+            thresholds_ratios=thresholds_ratios,
+            masses=tuple(m2 / new_theory.fact_to_ren**2 for m2 in masses),
+            order=new_theory.order,
+            method=couplings_mod_ev(new_operator.configs.evolution_method.value),
+            nf_ref=new_theory.num_flavs_ref,
+            max_nf=new_theory.num_flavs_max_as,
```
felixhekhorn, line 91: why are `num_falvs_ref` and `num_flavs_max_as` not inside `new_theory.couplings`?

felixhekhorn (commented):
first batch

src/eko/io/types.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+class QuarkMassSchemes(enum.Enum):
+    """Scheme to define heavy quark masses."""
+
+    MSBAR = "msbar"
+    POLE = "pole"
+
+
+class EvolutionMethod(enum.Enum):
```
felixhekhorn, line 135: I wonder whether we should lift this stuff up (and out of `io`) since these informations also need to be passed around in eko, e.g. here https://github.com/NNPDF/eko/blob/edf78a6a32bd159f78934c75e24ccb2d5f6acec9/src/eko/evolution_operator/__init__.py#L120 and we already found out that passing strings around in numba is slowing the thing down ... so, e.g., can we have both a string mapping for users and int mapping for numba?

if we want to do this, then not in this PR

src/ekobox/cards.py
```diff
@@ -0,0 +1,109 @@
...
+from math import nan
+from typing import Any, Dict
+
+import numpy as np
+import yaml
+
+from eko import basis_rotation as br
+from eko.io import runcards
+
+Card = Dict[str, Any]
```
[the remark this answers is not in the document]
felixhekhorn: in fact it is https://github.com/NNPDF/eko/blob/edf78a6a32bd159f78934c75e24ccb2d5f6acec9/src/eko/io/types.py#L162

src/eko/io/types.py
```diff
@@ -0,0 +1,162 @@
...
+
+T = typing.TypeVar("T")
+Quantity = typing.TypeVar("Quantity", bound=typing.Union[int, float])
+
+
+def reference_running(quantity: typing.Type[Quantity]):
+    """Generate running quantities reference point classes.
+
+    The motivation for dynamic generation is provided in module docstring.
+
```
felixhekhorn, line 43: you should still provide a `Parameters` section (here and below)

src/eko/io/manipulate.py
```diff
@@ -56,33 +56,45 @@ def xgrid_reshape(
...
 
     # build new grid
     for q2, elem in eko.items():
         ops = elem.operator
         errs = elem.error
         if targetgrid is not None and inputgrid is None:
             ops = np.einsum("ij,ajbk->aibk", target_rot, ops)
-            errs = np.einsum("ij,ajbk->aibk", target_rot, errs)
+            errs = (
+                np.einsum("ij,ajbk->aibk", target_rot, errs)
```
felixhekhorn, line 80: in `ekomark` you extracted the contraction string - you can also do here

felixhekhorn (commented):
Second batch before attempting the big ones

src/eko/runner/legacy.py
```diff
@@ -0,0 +1,124 @@
...
+        self,
+        theory_card: Union[RawCard, runcards.TheoryCard],
+        operators_card: Union[RawCard, runcards.OperatorCard],
+        path: os.PathLike,
+    ):
+        """Initialize runner.
+
+        Parameters
+        ----------
+        theory_card: dict
```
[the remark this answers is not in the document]
AleCandido: Yeah, maybe with type annotations it becomes redundant, since Sphinx is already extracting those

src/eko/runner/legacy.py
AleCandido: Just because I initially started couplings with only the reference values of `alphas` and `alphaem`, later on I added reference scales (since it made sense to keep them together).
We can also move them there, since they are only used for couplings, and we can pass the full data struct to the `Couplings` object :)

tests/conftest.py
```diff
@@ -39,94 +42,90 @@ def fake_pdf():
...
+        if self.cache is not None:
+            self.cache.close()
+            self.cache.access.path.unlink()
+        self.cache = None
+
+    def get(self, update: Optional[dict] = None):
+        """Get a fake output.
+
+        To force a new output, keeping all the defaults, just pass an empty
+        dictionary::
```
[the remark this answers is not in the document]
AleCandido: Nope, you want the double one, was intentional. It is rest syntax for code blocks: https://www.sphinx-doc.org/en/master/usage/restructuredtext/basics.html#literal-blocks

tests/conftest.py
```diff
@@ -39,94 +42,90 @@ def fake_pdf():
...
+
+            factory.get({})
+
+
+        Pay attention that the EKO created is an invalid one, since no path is
```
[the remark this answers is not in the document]
AleCandido: Indeed, I make such that it is actually a working one, it was stupid not to, unless we are actually mocking.

tests/eko/io/test_init.py
```diff
@@ -0,0 +1,285 @@
...
+        np.testing.assert_almost_equal(o2.xgrid.raw, fake_card["interpolation_xgrid"])
+
+
+class TestManipulate:
+    def test_xgrid_reshape(self, eko_factory: EKOFactory):
+        # create object
+        xg = interpolation.XGrid(np.geomspace(1e-5, 1.0, 21))
+        o1 = eko_factory.get()
+        o1.xgrid = xg
+        o1.rotations._targetgrid = xg
```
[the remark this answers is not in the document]
AleCandido: Yes, as I explained to you, we should never use those directly. But this I decided after writing this, and forgot to update...

src/eko/evolution_operator/__init__.py
```diff
@@ -277,9 +277,9 @@ def n_pools(self):
         return max(os.cpu_count() + n_pools, 1)
 
     @property
-    def fact_to_ren(self):
+    def xif(self):
```
felixhekhorn, line 280: this is $\xi_F^2$!

tests/eko/io/test_init.py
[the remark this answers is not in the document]
AleCandido: In particular, they are already following from the fallback, so these are completely unneeded.

tests/eko/io/test_init.py
```diff
@@ -0,0 +1,285 @@
...
+        o1.rotations._inputgrid = xg
+        o1._operators = {
+            10: eko.io.Operator.from_dict(
+                dict(
+                    operator=eko_identity([1, 2, len(xg), 2, len(xg)])[0],
```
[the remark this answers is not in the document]
AleCandido: It is, and this file contains the only tests still failing (if you look around, there are still a couple of occurrences of `fake_legacy`).

In particular, the current way to do this properly is to set the actual grids you want to use in the `eko_factory` runcards, before `get`ting it. The operators you still need to fix afterwards, since they are randomly generated.

tests/eko/io/test_init.py
```diff
@@ -0,0 +1,285 @@
...
+                    error=np.zeros((2, len(xg), 2, len(xg))),
+                )
+            )
+        }
+        xgp = interpolation.XGrid(np.geomspace(1e-5, 1.0, 11))
+        # only target
+        ot = copy.deepcopy(o1)
```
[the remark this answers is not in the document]
AleCandido: I missed it: we should always use `.deepcopy()` by default, unless there is a specific reason for sharing the folder.
That might be the case for tests, but I don't see it right now, so I'll keep going with the default.

src/eko/msbar_masses.py
```diff
@@ -339,26 +332,28 @@ def evolve(
...
+        list of |MSbar| masses squared
+
     """
     # TODO: sketch in the docs how the MSbar computation works with a figure.
     nfa_ref = theory_card["nfref"]
 
     q2_ref = np.power(theory_card["Qref"], 2)
     masses = np.concatenate((np.zeros(nfa_ref - 3), np.full(6 - nfa_ref, np.inf)))
-    fact_to_ren = theory_card["fact_to_ren_scale_ratio"] ** 2
+    xif = theory_card["fact_to_ren_scale_ratio"] ** 2
```
felixhekhorn, line 356: again: you can just call it `xif2`

tests/eko/io/test_init.py
```diff
@@ -0,0 +1,285 @@
...
+        # error
+        fpath = tmp_path / "fail.tar"
+        o1.deepcopy(fpath)
+        with pytest.raises(ValueError):
+            with EKO.append(fpath) as of:
+                manipulate.flavor_reshape(of)
+
+    def test_to_evol(self, eko_factory: EKOFactory, tmp_path):
+        xgrid = interpolation.XGrid([0.5, 1.0])
+        interpolation_polynomial_degree = 1
```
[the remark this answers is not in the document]
AleCandido: Agreed, I have been lazy and didn't check one by one, I believe.

src/eko/evolution_operator/__init__.py
AleCandido: Fixed in bd17d85

src/eko/msbar_masses.py
AleCandido: Fixed in 6dbebad4fc7d94a6fa67df91aa7e2ffaebc8c58e

src/ekobox/cards.py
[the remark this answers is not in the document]
AleCandido: I was aware I had a leftover somewhere, thank you for having found it :)

src/eko/io/types.py
AleCandido: > I wonder whether we should lift this stuff up (and out of `io`) since these informations also need to be passed around in eko, e.g. here

Yes, this I considered as well. Actually, the first instance of the `types.py` file appeared top-level, not inside `eko.io`, but then I moved in since it was mostly related to/used in runcards.
Type hints are extremely useful, I'm decorating even tests because they are giving me autocompletion (and it saves me the time of doing manually the lookup for attributes and methods), so maybe we want really to have eko-wide types.

AleCandido: > and we already found out that passing strings around in numba is slowing the thing down ... so, e.g., can we have both a string mapping for users and int mapping for numba?

I guess it's feasible, even though I'm not currently sure as the best way to implement it (a tuple value? some built-in? dataclass value?). However, we can reconsider later on.

src/ekobox/cards.py
```diff
@@ -0,0 +1,109 @@
...
+        scvar_method=None,
+        inversion_method=None,
+        n_integration_cores=0,
+    ),
+    debug=dict(
+        skip_singlet=False,
+        skip_non_singlet=False,
+    ),
+    rotations=dict(
+        xgrid=np.geomspace(1e-7, 1.0, 50).tolist(),
```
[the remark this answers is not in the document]
AleCandido: Maybe I would change the call to `make_lambert_grid` to `(min, max, npoints)`, to make it familiar to NumPy users (since it is the way `{lin,log,geom}space` work).

src/ekobox/cards.py
```diff
@@ -0,0 +1,109 @@
...
+        pids=list(br.flavor_basis_pids),
+    ),
+)
+
+
+class example:
```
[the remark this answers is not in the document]
AleCandido: I considered `staticmethod`s, but then I'd not be able to access each other. However, for the time being they are not doing it, so I can also revert to static.

About the scope: yes, it is a simple way to create it, better than making a module on its own just for the purpose, and as modules is useful for imports:
```py
from ekobox.cards import example

t = example.theory()
o = example.operator()
```

Moreover, having them as functions ensures the copy, and this is comfortable to avoid undesired effects :)

felixhekhorn (commented):
I had a look to everything and with one read it seems fine ... depending on the urgency we can postpone
- documentation is still scarse
- tutorials are of course completely broken

to be definitely addressed before merging:
- recover LHA benchmark

AleCandido:
Tests are failing on py3.8 for a not very clear reason, I need further investigation, but passing on py3.9 and py3.10.
The motivations are actually only two:
```
ValueError: negative dimensions are not allowed
TypeError: 'numpy.float64' object cannot be interpreted as an integer
```
All coming from the same place: deserialization. But it is mystery why it fails only on that Python version, it would be useful to have py3.8 instance to debug, maybe I will do inside a container.

I will try to debug first isolated benchmarks, they seems to be easier to solve, and failing also on recent versions :)

src/eko/io/types.py
AleCandido: There are a lot of parameters/returns section missing, but they are not the priority. Moreover, it is better to have docstrings, especially for complex functions, but if the function is simple enough, with telling names for parameters and type hints, the parameters section is not that fundamental (i.e. I'd rather spend my time on something else).

src/eko/io/types.py
AleCandido: Now a dedicated issue #181, we're not going to face properly in this PR

src/ekobox/cli/runcards.py
```diff
@@ -0,0 +1,39 @@
...
+
+
+@command.group("runcards")
+def subcommand():
+    """Manage EKO runcards."""
+
+
+@subcommand.command("example")
+@option_dest
+def sub_example(destination: pathlib.Path):
```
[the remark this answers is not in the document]
AleCandido: This one I like as it is, but we can rediscuss later on. CLI in general is a nice extra in this PR, but not the main focus.

src/eko/runner/legacy.py
AleCandido: Now an item on the original comment https://github.com/NNPDF/eko/pull/172#issue-1477202955, I will do.
I close here, just because the information is available in a more visible place.

src/ekobox/cards.py
```diff
@@ -0,0 +1,109 @@
...
+        inversion_method=None,
+        n_integration_cores=0,
+    ),
+    debug=dict(
+        skip_singlet=False,
+        skip_non_singlet=False,
+    ),
+    rotations=dict(
+        xgrid=np.geomspace(1e-7, 1.0, 50).tolist(),
+        pids=list(br.flavor_basis_pids),
```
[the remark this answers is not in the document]
AleCandido: Ok, might be weird, but it is not wrong. This is a problem with the runcard: if you do not want `pids` to be in the `OperatorCard`, please open a separate PR based on this one.
I would say that it should be easier to fix, but it might propagate in a few places. Since it is weird but not completely wrong (most of the time you just use the example one, that has the correct basis), then I'd rather not consider blocking for this PR.

Can you open the further PR yourself? @felixhekhorn

src/ekobox/cards.py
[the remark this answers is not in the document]
AleCandido: As @felixhekhorn explained to me, `lambert_grid` is somehow different from `{lin,log,geom}space`, since it has a fixed point 1, so `max` should not be changed (other than for special debugging purpose), so it makes to have it the last one. Moreover, often you do not want even to move `min`, but just `npoints`, so the arguments are in the correct order.

The only NumPy style modification I will do is then in the function name: instead of `make_lambert_grid`, I will transition to the more Pythonic `lambertgrid` (I would have liked `lambertspace`, that would make sense, but since it is already different in the signature, let's reflect this difference also in the name).

AleCandido:
Now also *most* of the isolated benchmarks are passing, apart from MSbar masses ones, but for some weird reason the CI is providing no logs:
https://github.com/NNPDF/eko/actions/runs/3715495494/jobs/6300731885#step:8:19

It starts being extremely annoying...

AleCandido:
Looks like the type hint of a NumPy array, i.e. [`npt.NDArray`](https://numpy.org/doc/stable/reference/typing.html#numpy.typing.NDArray), was not a subclass of [`np.ndarray`](https://numpy.org/doc/stable/reference/generated/numpy.ndarray.html#numpy.ndarray), in some weird sense, since instead it was already in its MRO:
```py
>>> issubclass(npt.NDArray, np.ndarray)
False
>>> np.ndarray in npt.NDArray.__mro__
True
```

However, it seems like this weird behavior is not present any longer after py3.8, and now we have a fix for py3.8 as well:
https://github.com/NNPDF/eko/blob/45231a62a2a9821a76d1dff9bdc2e3ee00522f6a/src/eko/io/dictlike.py#L123

Baseline: types are getting more and more reliable, but (especially in older versions) there are still some rough edges...

AleCandido:
And now the mystery: I ran in the very same container of isolated benchmarks on my computer, and this is what I get
```
============================================== test session starts ==============================================
platform linux -- Python 3.10.8, pytest-7.1.3, pluggy-1.0.0
rootdir: /home/ciao/eko, configfile: pyproject.toml
plugins: cov-3.0.0, env-0.6.2
collected 19 items / 6 skipped

benchmarks/eko/benchmark_ad.py ..                                                                         [ 10%]
benchmarks/eko/benchmark_evol_to_unity.py .                                                               [ 15%]
benchmarks/eko/benchmark_msbar_evolution.py ...                                                           [ 31%]
benchmarks/eko/benchmark_strong_coupling.py .............                                                 [100%]

---------- coverage: platform linux, python 3.10.8-final-0 -----------
Coverage HTML written to dir htmlcov
Coverage XML written to file coverage.xml


=================================== 19 passed, 6 skipped in 120.01s (0:02:00) ===================================
```

benchmarks/ekobox/benchmark_evol_pdf.py
```diff
@@ -78,25 +69,28 @@ def benchmark_evolve_single_member(tmp_path, cd, lhapdf_path):
...
-    )
-    theory = cards.generate_theory(0, 1.0)
-    nt, no = compatibility.update(theory, op)
+def benchmark_evolve_more_members(
+    tmp_path, cd, lhapdf_path, theory_card: TheoryCard, operator_card: OperatorCard
+):
+    theory = theory_card
+    theory.order = (1, 0)
+    op = operator_card
+    operator_card.mu0 = 1.0
```
felixhekhorn, line 78:
```suggestion
    op.mu0 = 1.0
```
I guess? (and similar below)

src/eko/couplings.py
```diff
@@ -1,6 +1,10 @@
-r"""Contains the QCD beta function coefficients and the handling of the running coupling :math:`\alpha_s`.
+r"""Manage running (and non) couplings.
```
felixhekhorn, line 1:
```suggestion
r"""Manage running (and fixed) couplings.
```
do you mean like this? I'm confused about the language

src/eko/couplings.py
```diff
@@ -357,46 +377,50 @@ class Couplings:
...
         # Sanity checks
-        if couplings_ref[0] <= 0:
-            raise ValueError(f"alpha_s_ref has to be positive - got {couplings_ref[0]}")
-        if couplings_ref[1] <= 0:
-            raise ValueError(
-                f"alpha_em_ref has to be positive - got {couplings_ref[1]}"
-            )
-        if scale_ref <= 0:
-            raise ValueError(f"scale_ref has to be positive - got {scale_ref}")
+        def positive(name, var):
```
felixhekhorn, line 388:
```suggestion
        def assert_positive(name, var):
```
with according changes below

src/eko/couplings.py
```diff
@@ -357,46 +377,50 @@ class Couplings:
...
+        positive("alpha_s_ref", couplings.alphas.value)
+        positive("alpha_em_ref", couplings.alphaem.value)
+        positive("scale_ref", couplings.alphas.scale)
         if order[0] not in [0, 1, 2, 3, 4]:
             raise NotImplementedError("a_s beyond N3LO is not implemented")
         if order[1] not in [0, 1, 2]:
             raise NotImplementedError("a_em beyond NLO is not implemented")
         self.order = tuple(order)
+        method = method.value
         if method not in ["expanded", "exact"]:
```
felixhekhorn, line 401: Can this still fail if we have an enum already?

src/eko/io/exceptions.py
```diff
@@ -24,3 +24,35 @@ class OperatorLocationError(ValueError, OutputError):
     def __init__(self, path: os.PathLike):
         self.path = path
         super().__init__(f"Path '{path}' not in operators folder")
+
+
+class ReadOnlyOperator(RuntimeError, OutputError):
+    """It is not possible to write on a read-only operator.
+
+    In particular, the behavior would be deceitful, since writing is possible
```
felixhekhorn, line 32: This should go into a comment rather than a docstring - it's only relevant for developers

src/eko/io/exceptions.py
```diff
@@ -24,3 +24,35 @@ class OperatorLocationError(ValueError, OutputError):
...
+    But eventually, no writing will happen on a persistent archive, so any
+    modification is lost after exiting the program.
+
+    """
+
+
+class ClosedOperator(RuntimeError, OutputError):
+    """It is not possible to write on nor to read from a closed operator.
+
+    This is milder issue than :cls:`ReadOnlyOperator`, since in this case not
```
felixhekhorn, line 43: see above

src/eko/io/manipulate.py
```diff
@@ -186,20 +225,25 @@ def to_evol(eko: EKO, source: bool = True, target: bool = False):
...
+        rotate on the input tensor
+    target :
+        rotate on the output tensor
 
     """
     # rotate
     inputpids = br.rotate_flavor_to_evolution if source else None
     targetpids = br.rotate_flavor_to_evolution if target else None
-    flavor_reshape(eko, inputpids=inputpids, targetpids=targetpids)
+    # prevent metadata update, since flavor_reshape ha not enough information
```
felixhekhorn, line 239:
```suggestion
    # prevent metadata update, since flavor_reshape has not enough information
```

src/eko/io/types.py
```diff
@@ -63,6 +65,14 @@ class CouplingsRef(DictLike):
...
     alphas: FloatRef
     alphaem: FloatRef
+    num_flavs_ref: Optional[FlavorsNumber]
+    r"""Number of active flavors at strong coupling reference scale.
+
+    I.e. :math:`n_{f,\text{ref}}(\mu^2_{\text{ref}})`, formerly called
+    ``nfref``.
+
+    """
+    num_flavs_max_as: FlavorsNumber
```
felixhekhorn, line 75:
```suggestion
    max_num_flavs: FlavorsNumber
```
`as` has become redundant and word order would be more English

src/eko/io/types.py
```diff
@@ -63,6 +65,14 @@ class CouplingsRef(DictLike):
 
     alphas: FloatRef
     alphaem: FloatRef
+    num_flavs_ref: Optional[FlavorsNumber]
```
felixhekhorn, line 68: inside you now call it "scale" - maybe change also here?

src/eko/io/struct.py
```diff
@@ -338,6 +342,52 @@ class AccessConfigs:
...
+        """
+        if msg is None:
+            msg = ""
+
+        self.assert_open()
+        if self.readonly:
+            raise exceptions.ReadOnlyOperator(msg)
+
+    @property
+    def read(self):
```
felixhekhorn, line 379: why do we need this alias?

src/eko/io/struct.py
```diff
@@ -338,6 +342,52 @@ class AccessConfigs:
...
+        exceptions.ReadOnlyOperator
+            if operators has been declared read-only
+
+        """
+        if msg is None:
+            msg = ""
+
+        self.assert_open()
+        if self.readonly:
+            raise exceptions.ReadOnlyOperator(msg)
```
felixhekhorn, line 376: why do you need a custom message here?

src/eko/msbar_masses.py
```diff
@@ -332,16 +344,27 @@ def evolve(m2_ref, q2m_ref, strong_coupling, xif2, q2_to, nf_ref=None, nf_to=Non
...
     r"""Compute the |MSbar| masses.
 
     Computation is performed solving the equation :math:`m_{\bar{MS}}(\mu) =
     \mu` for each heavy quark and consistent boundary contitions.
 
     Parameters
     ----------
-    theory_card : dict
-        theory run card
+    nf_ref :
```
felixhekhorn, line 362: this is out of sync

tests/eko/test_msbar_masses.py
```diff
@@ -1,128 +1,150 @@
...
-    "HQ": "MSBAR",
-    "Qmc": 2.1,
-    "Qmb": 4.1,
-    "Qmt": 174.9,
-    "order": (3, 0),
-    "ModEv": "TRN",
-    "ModSV": None,
-}
+@pytest.fixture()
+def theory_card(theory_card: TheoryCard):
```
felixhekhorn, line 16: are you're sure you want to shadow here?

src/eko/interpolation.py
```diff
@@ -243,48 +244,50 @@ def evaluate_x(x, area_list):
...
     """
     x = np.log(x)
     return evaluate_x(x, area_list)
 
 
 class BasisFunction:
-    """
-    Object containing a list of areas for a given polynomial number
-    defined by (xmin-xmax) and containing a list of coefficients.
+    """Represent an element of polynomial basis.
```
felixhekhorn, line 270:
```suggestion
    """Represent an element of the polynomial basis.
```

src/eko/interpolation.py
```diff
@@ -243,48 +244,50 @@ def evaluate_x(x, area_list):
...
+
+    It contains a list of areas for a given polynomial number defined by
+    (xmin-xmax) and containing a list of coefficients.
```
felixhekhorn, line 273:
```suggestion
    (xmin-xmax) which in turn contain a list of coefficients.
```

src/eko/interpolation.py
```diff
@@ -243,48 +244,50 @@ def evaluate_x(x, area_list):
...
 
-    Upon construction will generate all areas and generate and compile
-    a function to evaluate in N (or x) the interpolator
+    Upon construction will generate all areas and generate and compile a
```
felixhekhorn, line 275:
```suggestion
    Upon construction it will generate all areas and generate and compile a
```

src/eko/interpolation.py
```diff
@@ -243,48 +244,50 @@ def evaluate_x(x, area_list):
...
+    xgrid : numpy.ndarray
+        Grid in x-space from which the interpolators are constructed
+    poly_number : int
+        number of polynomial
+    list_of_blocks: list(tuple(int, int))
+        list of tuples with the (kmin, kmax) values for each area
+    mode_log: bool
+        use logarithmic interpolation?
+    mode_N: bool
+        if true compiles the function on N, otherwise compiles x
```
felixhekhorn, line 289:
```suggestion
        if true compiles the function on N, otherwise compiles on x
```

felixhekhorn:
(Just to say that the rebase messed up the review since github completely lost track ... I hope I started again from where I left ...)

felixhekhorn:
Even after https://github.com/NNPDF/eko/pull/172/commits/714a1560cb86aeae93a566ecb9b668e5b4772fc8 LHA benchmarks fail with
```
  File "/home/felix/Physik/N3PDF/EKO/eko/benchmarks/lha_paper_bench.py", line 226, in <module>
    obj.benchmark_plain(0)
  File "/home/felix/Physik/N3PDF/EKO/eko/benchmarks/lha_paper_bench.py", line 123, in benchmark_plain
    self.run_lha(self.plain_theory(pto))
  File "/home/felix/Physik/N3PDF/EKO/eko/benchmarks/lha_paper_bench.py", line 109, in run_lha
    self.run(
  File "/home/felix/.cache/pypoetry/virtualenvs/eko-KkPVjVhh-py3.10/lib/python3.10/site-packages/banana/benchmark/runner.py", line 416, in run
    self.run_config(session, t, o, pdf_name, use_replicas)
  File "/home/felix/.cache/pypoetry/virtualenvs/eko-KkPVjVhh-py3.10/lib/python3.10/site-packages/banana/benchmark/runner.py", line 313, in run_config
    log_record = self.insert_log(session, t, o, pdf, me, ext)
  File "/home/felix/.cache/pypoetry/virtualenvs/eko-KkPVjVhh-py3.10/lib/python3.10/site-packages/banana/benchmark/runner.py", line 342, in insert_log
    log_record = self.log(t, o, pdf, me, ext)
  File "/home/felix/Physik/N3PDF/EKO/eko/src/ekomark/benchmark/runner.py", line 184, in log
    pdf_grid = apply.apply_pdf_flavor(
  File "/home/felix/Physik/N3PDF/EKO/eko/src/ekomark/apply.py", line 64, in apply_pdf_flavor
    pdfs = np.zeros((len(eko.rotations.inputpids), len(eko.rotations.inputgrid)))
AttributeError: 'NoneType' object has no attribute 'rotations'
```

src/eko/couplings.py
AleCandido: I applied "minimal" changes, as much as possible. In principle, you can pass whatever at runtime, the check about the type is only in the static analyzer, so in a sense the two tests are decoupled. But if you pass the static analyzer, in principle should be enough.

src/eko/io/exceptions.py
AleCandido: Nope, the error itself is relevant for whoever is getting the error. If the user will never get the error, so much the better, but in case, better if he can access an explanation as well. If he won't understand it, well, we loose nothing.

src/eko/couplings.py
AleCandido: Yes: it was running and non-running (i.e. fixed). Suggestion accepted

AleCandido:
> Even after [714a156](https://github.com/NNPDF/eko/commit/714a1560cb86aeae93a566ecb9b668e5b4772fc8) LHA benchmarks fail with

The problem now is to fix MSbar masses: updating the test to use the new theory card something got messed up. The moment unit tests are back (only MSbar masses affected), I will propagate to iso-bench (only MSbar masses missing), and finally I will tackle the actual benchmarks.

src/eko/io/types.py
AleCandido: Impossible, "scale" is the reference scale for running quantities, so it is always a member of a `FloatRef` or an `IntRef`.

src/ekobox/cli/runcards.py
```diff
@@ -0,0 +1,39 @@
...
+from .. import cards
+from . import library as lib
+from .base import command
+
+_logger = logging.getLogger(__name__)
+
+DESTINATION = pathlib.Path.cwd().absolute() / "runcards"
+"""Default destination for generated files"""
+
+option_dest = lib.destination(DESTINATION)
```
giacomomagni, line 14: isn't this a bit verbose, just for an additional option?

src/ekomark/apply.py
```diff
@@ -32,7 +35,11 @@ def apply_pdf(eko, lhapdf_like, targetgrid=None, rotate_to_evolution_basis=False
     return apply_pdf_flavor(eko, lhapdf_like, targetgrid)
 
 
-def apply_pdf_flavor(eko, lhapdf_like, targetgrid=None, flavor_rotation=None):
+CONTRACTION = "ajbk,bk"
+#  CONTRACTION = "jakb,bk"
```
giacomomagni, line 39:
```suggestion
```

giacomomagni (commented):
A small reminder that also tutorials need to be update probably.

AleCandido:
@niclaurenti there is a huge copy-pasting of tests in https://github.com/NNPDF/eko/blob/tempdir/tests/eko/test_couplings.py

Essentially, all the `_QED` and `_mix` ones are the same code, but for `order`. Please, take the time to refactor it, because it only makes it harder to maintain (I had to repeat the same changes over and over every time that something changes in couplings).

src/eko/msbar_masses.py
```diff
@@ -328,52 +334,71 @@ def evolve(
             )
             matching = 1.0
             for pto in range(1, strong_coupling.order[0]):
-                for l in range(pto + 1):
+                for logpow in range(pto + 1):
```
giacomomagni, line 337:
```suggestion
                for logpow in range(1, pto + 1):
```
This should fix the `0**0` issue without changing the thing, since all the `m_coeffs[0, :]` are 0

AleCandido: This is not what I observe, since if L is 0, the only contributions to `matching` are coming from the `logpow=0` term...

giacomomagni: yes, true. I was wrong `pto=0` is already excluded, we do the same for `a_s`. But then yes this is what we wanted...
```suggestion
                # 0**0=1, from NNLO there is a matching also in this case
                for logpow in range(pto + 1):
```

src/eko/io/runcards.py
```diff
@@ -0,0 +1,373 @@
...
+    # TODO: drop legacy compatibility, only linear scales in runcards, such
+    # that we will always avoid taking square roots, and we are consistent with
+    # the other scales
+    _mugrid: Optional[npt.NDArray] = None
+    _mu2grid: Optional[npt.NDArray] = None
+    """Array of final scales."""
+
+    # optional
+    eko_version: str = vmod.__version__
+    """Perturbative order of the QCD and QED couplings."""
```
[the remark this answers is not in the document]
felixhekhorn: too much copy paste I guess :upside_down_face:

AleCandido:
Tests are back, if I manage to get also isolated benchmarks (feasible by today), then we are left with proper benchmark. If you want to have a look that might speed up things a little bit...

src/eko/io/runcards.py
```diff
@@ -0,0 +1,373 @@
...
+    """"""
+    intrinsic_flavors: IntrinsicFlavors
+    """"""
+    quark_masses: HeavyQuarkMasses
+    """"""
+    quark_masses_scheme: QuarkMassSchemes
+    """Scheme used to specify heavy quark masses."""
+    matching: MatchingScales
+    """"""
+    # TODO: drop this
```
felixhekhorn, line 59: I guess the `TODO` can be dropped - since we need the setting ... instead, we need a docstring

src/eko/io/runcards.py
```diff
@@ -0,0 +1,373 @@
...
+    """
+    ev_op_iterations: int
+    """Number of intervals in which to break the global path."""
+    interpolation_polynomial_degree: int
+    """Degree of elements of the intepolation polynomial basis."""
+    interpolation_is_log: bool
+    r"""Whether to use polynomials in :math:`\log(x)`.
+    If `false`, polynomials are in :math:`x`.
+    """
+    scvar_method: Optional[ScaleVariationsMethod]
```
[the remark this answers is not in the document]
felixhekhorn: why? we always said there are no defaults in eko ...

src/eko/io/runcards.py
AleCandido: Yes, instead of the dropping we might need a better name. If you have sensible proposals better, otherwise we will keep going with it.

src/eko/io/types.py
```diff
@@ -0,0 +1,174 @@
...
+IntrinsicFlavors = typing.List[FlavorIndex]
+
+
+@dataclasses.dataclass
+class CouplingsRef(DictLike):
+    """Reference values for coupling constants."""
+
+    alphas: FloatRef
+    alphaem: FloatRef
+    num_flavs_ref: Optional[FlavorsNumber] = None
```
felixhekhorn, line 68: do we really want a default here? (and below?)

src/eko/io/runcards.py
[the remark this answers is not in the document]
AleCandido: Yes, sorry. I just violated that constraint myself in https://github.com/NNPDF/eko/commit/d980b0df617f3054645376f7d3f2b8780529ea4c#diff-501cd4589f93df16bdcb78d30c90b967d14c3807a459eded2fd14990bc6ad879

I will fix later on...

src/eko/msbar_masses.py
```diff
@@ -328,52 +334,71 @@ def evolve(
...
+        list of |MSbar| masses squared
+
     """
     # TODO: sketch in the docs how the MSbar computation works with a figure.
-    nfa_ref = theory_card["nfref"]
-
-    q2_ref = np.power(theory_card["Qref"], 2)
-    masses = np.concatenate((np.zeros(nfa_ref - 3), np.full(6 - nfa_ref, np.inf)))
-    fact_to_ren = theory_card["fact_to_ren_scale_ratio"] ** 2
+    mu2_ref = couplings.alphas.scale**2
```
felixhekhorn, line 376: oookay, so the statement "inside eko all scales are squared" is apparently slightly more convoluted: "inside eko" really means "inside the respective function bodies and classes", but not inside the runcard objects (such as `couplings.alphas.scale`) which are now passed around. To be explicit: "inside eko" does __not__ mean "the arguments of functions" - do you confirm @AleCandido ?

tests/eko/test_msbar_masses.py
```diff
@@ -5,124 +5,144 @@
...
-    "ModEv": "TRN",
-    "ModSV": None,
-}
-
-
-class TestMsbarMasses:
-    def test_compute_msbar_mass(self):
         # Test solution of msbar(m) = m
-        for method in ["EXA", "EXP"]:
+        #  __import__("pdb").set_trace()
```
felixhekhorn, line 34:
```suggestion
```

src/eko/io/runcards.py
felixhekhorn: not for the moment, let's keep it for now

src/eko/io/types.py
```diff
@@ -0,0 +1,174 @@
+"""Common type definitions, only used for static analysis.
```
felixhekhorn, line 1: VS code is rightly complaining about you shadowing python [`types`](https://docs.python.org/3/library/types.html?highlight=types#module-types)

src/eko/msbar_masses.py
AleCandido: The difference from inside and outside is really IO: if it is user input is outside, and runcards represent the user input (since their content comes directly from deserialization).
However, we can make properties for squared scales :)

src/eko/io/types.py
AleCandido: Already declared, there is an open point in the OP. It was just an easy fix.

src/eko/msbar_masses.py
```diff
@@ -1,10 +1,20 @@
-r"""
-This module contains the |RGE| for the |MSbar| masses
-"""
+r"""|RGE| for the |MSbar| masses."""
+from math import nan
```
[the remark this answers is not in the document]
AleCandido: Why not `math.nan`? In case of doubt, I prefer the simpler one, and `math` is standard library.

However, NumPy scalars are also problematic for serialization, so the alternative would be `float("nan")`, but it is lengthier to type many times.

src/ekobox/cli/runcards.py
AleCandido: this is copied from `nnusf`: this option is actually happening to be useful several times (not yet in EKO)

src/eko/io/types.py
```diff
@@ -0,0 +1,174 @@
...
+
+Unfortunately, Python has still some discrepancies between runtime classes and
```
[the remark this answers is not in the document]
AleCandido: This module is very internal, so yes, it is meant for developers...

src/eko/msbar_masses.py
AleCandido: I will fix. However, unfortunately, plenty of docstrings are at this point, so we will need a docstrings review campaign right before or right after merging

src/eko/msbar_masses.py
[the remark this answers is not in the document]
giacomomagni: fine, you just had numpy imported here

AleCandido:
MSbar isolated benchmarks fixed, I'm now left with a regression in strong coupling tests: they are running but getting wrong values. I expected that one input is not passed as before (like linear vs squared energies)

tests/eko/test_msbar_masses.py
AleCandido: Yes, this fixture is overwriting the other one for this module. Indeed it is just an update, and it depends from the broader.

src/eko/io/struct.py
AleCandido: Sometimes you might want to provide a specific reason why it is read-only, if the opening happened inside EKO (there is a single use case somewhere, you can check it).

tests/eko/test_msbar_masses.py
felixhekhorn: then you should explicitly silence pylint

src/eko/io/struct.py
AleCandido: Because it pairs with `write`, that is a more complex condition then the bare attributes.

src/eko/io/types.py
felixhekhorn: Mmm? but it refers exactly to that "scale" - it is the number of flavors at the scale ... if you wish `ref_scale`? (a more drastic option would be to move even inside `FloatRef` - not sure we want that ...)

src/eko/couplings.py
```diff
@@ -1,6 +1,5 @@
-r"""Manage running (and non) couplings.
+r"""Manage running (and fixed) couplings.
 
-Contains the QCD beta function coefficients and the handling of the running
 coupling :math:`\alpha_s`.
```
felixhekhorn, line 3: I guess @giacomomagni was more after the idea :upside_down_face: 
```suggestion
Manage QCD coupling :math:`\alpha_s` and QED coupling :math:`\alpha`.
We provide an interface to access them simultaneously and provide several
strategies to solve the associated |RGE|.
```

src/eko/io/legacy.py
```diff
@@ -0,0 +1,184 @@
...
+
+
+def load_arrays(dir: pathlib.Path) -> dict:
+    """Load arrays from compressed dumps."""
+    arrays = {}
+    for fp in dir.glob(f"*{ARRAY_SUFFIX}"):
+        with lz4.frame.open(fp, "rb") as fd:
+            # static analyzer can not guarantee the content to be bytes
+            content = fd.read()
+            if isinstance(content, str):
```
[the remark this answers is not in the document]
AleCandido: It will never occur, since we are writing bytes, but `lz4.frame.open().read()` can return both, according to the string you put in the second argument of `open`. In this case is `"rb"` so it will be forever bytes.

However, a string literal is a value, and PyRight (same for mypy) cannot rely on values (that only exist at runtime) for the static analysis (`"rb"` is hardcoded, but in order to discriminate the behavior according to that value it should execute `open`, so it is only possible at runtime).
Thus is not able to guarantee that it will be bytes, and I'm telling it through `isinstance`.

Essentially, I want to enforce more Type Hints, and make better use of the editor static checks (in practice, it prevents the red underlining in VS code).

src/eko/io/legacy.py
```diff
@@ -0,0 +1,184 @@
...
+        if name in plural(OPERATOR):
+            op5 = ar
+        elif name in plural(ERROR):
+            err5 = ar
+        else:
+            warnings.warn(f"Unrecognized array loaded, '{name}'")
+
+    if op5 is None:
+        raise RuntimeError("Operator not found")
+    if err5 is None:
```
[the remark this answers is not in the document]
AleCandido: I want to accept error-less operators, to compress their size.

src/eko/io/runcards.py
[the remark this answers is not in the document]
AleCandido: Too much copy-paste, I agree. Fixed in d199660013a75747d32faa15278086b02bf8fed9

src/eko/couplings.py
AleCandido: Done in 6321b6e503f4cb145c8e9f3df3c7386657404f5d

tests/eko/test_msbar_masses.py
AleCandido: We are not pylinting tests:
https://github.com/NNPDF/eko/blob/e3466fae35fd245c7e1a99158d8582f590276555/pyproject.toml#L107-L108

src/eko/io/types.py
AleCandido: The scale related to `num_flavs_ref` is the scale of `alphas`, that is already present: indeed both `FloatRef` and `IntRef` are a 2-tuple (aka a couple), whose first element is either a `float` or an `int`, and the second one is just a `float`, which is actually the scale.

So the scale of `num_flavs_ref` is `alphas.scale`, and as you can see it is already named `scale`.

benchmarks/ekobox/benchmark_evol_pdf.py
```diff
@@ -76,7 +76,7 @@ def benchmark_evolve_more_members(
     theory = theory_card
     theory.order = (1, 0)
     op = operator_card
-    operator_card.mu0 = 1.0
+    op.mu0 = 1.0
     operator_card.mu2grid = [10, 100]
     operator_card.rotations.xgrid = XGrid([1e-7, 0.01, 0.1, 0.2, 0.3])
```
felixhekhorn, line 81:
```suggestion
    op.mu2grid = [10, 100]
    op.rotations.xgrid = XGrid([1e-7, 0.01, 0.1, 0.2, 0.3])
```

src/eko/io/types.py
AleCandido: True, but it is a reflection module to mess up with Python classes creation (not type hints, since that is `typing` module... I know, it is rather confusing).

Do we really care?

AleCandido: In the end, we can always import scoped, like:
```py
import eko.io.types
```
or once lifted:
```py
import eko.types
```

It is not the name of a top-level package, so it should be fine. Don't you think so?

felixhekhorn: I'd say we can also rename and so appease VS code (i.e. have only relevant curly lines around) ... `common_types`? `ctypes`? `input_types`?

AleCandido: [`ctypes`](https://docs.python.org/3/library/ctypes.html?highlight=ctypes#module-ctypes)

felixhekhorn (changes requested):
I attempted to fix the tutorial, but the simple application fails with `OutputExistsError: /tmp/tmpghgoalu4` - there is something wrong with the automatic folder allocation ...

AleCandido:
> I attempted to fix the tutorial, but the simple application fails with `OutputExistsError: /tmp/tmpghgoalu4` - there is something wrong with the automatic folder allocation ...

I will try to have a look once the points in the OP will be completed.

tests/eko/io/test_init.py
```diff
@@ -0,0 +1,285 @@
...
+        path_copy = tmp_path / "eko_copy.tar"
+        o1.deepcopy(path_copy)
+        with EKO.append(path_copy) as o2:
+            manipulate.xgrid_reshape(
+                o2, interpolation.XGrid([0.1, 1.0]), interpolation.XGrid([0.1, 1.0])
+            )
+            manipulate.flavor_reshape(o2, inputpids=np.array([[1, -1], [1, 1]]))
+        # reload
+        with EKO.read(path_copy) as o3:
+            chk_keys(o1.raw, o3.raw)
```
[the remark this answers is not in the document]
AleCandido: Done in 758349692dca7d17e9b3f7fb4514c101ad9ac329

src/eko/io/struct.py
```diff
@@ -0,0 +1,1081 @@
...
+    def read(cls, path: os.PathLike):
+        """Read the content of an EKO.
+
+        Type-safe alias for::
+
+            EKO.open(... , "r")
+
+        """
+        eko = cls.open(path, "r")
+        assert isinstance(eko, EKO)
```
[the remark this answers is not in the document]
AleCandido: Yes, they have to, but I'm passing this information to the static analyzer, see https://github.com/NNPDF/eko/pull/172#discussion_r1055483647

src/eko/io/struct.py
```diff
@@ -0,0 +1,1081 @@
...
+        without_err = op.error is None
+        oppath = self.paths.opnewpath(mu2, compress=compress, without_err=without_err)
+
+        with open(oppath, "wb") as fd:
+            without_err2 = op.save(fd, compress)
+        assert without_err == without_err2
+
+        self._operators[mu2] = op
+
+    def __delitem__(self, q2: float):
```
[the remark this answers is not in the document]
AleCandido: The property not, but the operator yes.

You might be right, but this is more intuitive on another side, since:
```py
del eko[10.0]
```
is more intuitively destroying an operator rather than:
```py
eko[10.0] = None
```

I understand it is confusing with respect to usual dictionary syntax, but it is somehow the analogue of `pathlib` usage of `__truediv__`: it is not a division, but a concatenation. Intuitively it is closer to `+`, but in that context the symbol makes sense.

src/eko/io/runcards.py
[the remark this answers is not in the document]
felixhekhorn: also here https://github.com/NNPDF/eko/blob/7a4d17232964fbfdd6b48a27cfda9e984d2cca7a/src/eko/couplings.py#L384-L385

felixhekhorn:
with https://github.com/NNPDF/eko/pull/172/commits/d2df0114981bc39f2b4e0d204444d6f73238b6ee the $a_s$ tutorial runs - but the result is crazy: `The value of alpha_s at Q^2=100 GeV^2 is:  26027.470163094906`

felixhekhorn:
I attempted to solve https://github.com/NNPDF/eko/issues/169 here by using the new `from ekobox.cards import example` - however this leaves us with a new problem: the usage of the new objects is completely undocumented since https://eko.readthedocs.io/en/latest/code/IO.html refers to the NNPDF scheme (which is still needed somewhere since https://docs.nnpdf.science/theory/theoryparamsdefinitions.html is incomplete for our purpose)

AleCandido:
> I attempted to solve https://github.com/NNPDF/eko/issues/169 here by using the new `from ekobox.cards import example` - however this leaves us with a new problem: the usage of the new objects is completely undocumented since https://eko.readthedocs.io/en/latest/code/IO.html refers to the NNPDF scheme (which is still needed somewhere since https://docs.nnpdf.science/theory/theoryparamsdefinitions.html is incomplete for our purpose)

It is documented by docstrings and type hints. 
Thus, we should simply rewrite the IO section removing most of the content, and adding the relevant pointers. 

Actually we can also use autodoc manually to make runcards objects docstrings and dictlike module docstring to appear in the IO section directly.

src/eko/io/types.py
AleCandido: Done in e888119543e20d3d574444426729426ac27ba959

felixhekhorn:
> I knew this was a complicated one, the constructor is spread all over tests, see https://github.com/NNPDF/eko/pull/187

maybe annoying, but not complicated ... instead this one https://github.com/NNPDF/eko/blob/5c3c792d62a3e4eccd9d87dc5515e2fb13e330b4/src/eko/io/types.py#L69
I leave to you @AleCandido - because it is complicated ... actually the thing can be either `int` or `None`; both is fine, but either has to be explicitly chosen - I think you have to make a new type for this?

AleCandido:
> I leave to you @AleCandido - because it is complicated ... actually the thing can be either `int` or `None`; both is fine, but either has to be explicitly chosen - I think you have to make a new type for this?

`FlavorsNumber` is an alias of `int`, and `Optional[X]` is the old shortcut for `Union[X, None]` (now it is simpler to write `X | None`, since they added the type union operator in py3.10).
So, the type hint is exactly the one you want. If you want to drop the default, you just have to remove the ` = None` part :)

AleCandido:
Just missing coverage, that I would postpone to a later PR.

I'm ready to merge, and we are just 3 commits away from the 200 target :)

felixhekhorn:
> Just missing coverage, that I would postpone to a later PR.

Fine - then let's do it (of course you have the pleasure of the big green button :upside_down_face: ). However, afterwards, we have to actually put some effort to cover the remaining 140 lines ...

> I'm ready to merge, and we are just 3 commits away from the 200 target :)

and with https://github.com/NNPDF/eko/pull/172/commits/d05ca845b0a2994d95d5a97b366651676f1e220d there are only two left ... (the improved test fails indeed without the accompanying fix)

felixhekhorn:
The moment we merge
- @giacomomagni has to take care of #148 (mostly orthogonal, but still non-zero impact)
- @niclaurenti has to take care of #135 (mostly orthogonal, but still non-zero impact)
- @andreab1997 should try to get pineko back
- among the three of us @AleCandido @giacomomagni to fix #170 for the meeting on Thursday

andreab1997:
> The moment we merge
> 
> * @giacomomagni has to take care of [N3LO ad variations #148](https://github.com/NNPDF/eko/pull/148) (mostly orthogonal, but still non-zero impact)
> * @niclaurenti has to take care of [Implement QED exact solution #135](https://github.com/NNPDF/eko/pull/135) (mostly orthogonal, but still non-zero impact)
> * @andreab1997 should try to get pineko back
> * among the three of us @AleCandido @giacomomagni to fix [LO polarised anomalous dimension #170](https://github.com/NNPDF/eko/pull/170) for the meeting on Thursday

Yes, I will do that. BTW remember to revert all the `alpha_raw` stuff here (as we agreed, but we could also open another PR of course).

src/eko/io/struct.py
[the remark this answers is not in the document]
AleCandido: In any case, usually you are not supposed to use this directly, there are better user facing methods (like the `.items()` iterator or the `.operator()` context manager).

AleCandido:
I would just wait for @giacomomagni confirmation, and then merge.



> > *
> 
> Yes, I will do that. BTW remember to revert all the `alpha_raw` stuff here (as we agreed, but we could also open another PR of course).

@andreab1997 open a dedicated issue for the time being, and right after merging we will start the (hopefully small) PR

giacomomagni (approved):
Okay green light from my side. 
We might need to updated something minor also in Yadism?

AleCandido:
> We might need to updated something minor also in Yadism?

Eventually, yes. 
For the time being, yadism is just running with an older version of EKO (essentially for interpolation, that is almost unchanged), so you might need to environments for grids and FkTables computation, but it would work.

AleCandido:
> > 
> 
> If I recall correctly we have already introduced `eko=0.11` in yadism [see here](https://github.com/NNPDF/yadism/pull/157). I can take care of it.

You're right, and thanks for the offering. I would suggest to pretend it is working, and as soon as anyone find an issue, we will get back to you (or any of us) and ask to make it compatible.

felixhekhorn:
just to say that for sure we need to do some more adjustments for eko v0.12 e.g. [here](https://github.com/NNPDF/yadism/blob/7956826bb89d60aa768cb752002cd155453a2ca5/src/yadmark/benchmark/runner.py#L49)

PR #174, opened 2022-12-08
Include new a_gg^{3} results

giacomomagni:
This PR is for:
* [x] including the new calculations on `a_gg^{3}` from https://arxiv.org/pdf/2211.05462v1.pdf, which is parametised
* [x] update the non complete part of `a_Hg^{3}` using all the moments provided here: https://arxiv.org/pdf/0904.3563.pdf (eq 8.50-8.54) and the small-x limit given by https://arxiv.org/pdf/1205.5727.pdf (eq 3.47)
* [x] Remove Non singlet harmonics cache in singlet like quantities (`A_qq`), in favor of the full `eta=(-1)^n` to be continued with `+-1`


tests/eko/test_matching_n3lo.py
```diff
@@ -184,13 +181,8 @@ def test_Blumlein_3():
...
                 aS3[2, 1], ref_val_Hq[L][idx], rtol=2e-5, atol=2e-6
             )
-
-            # here we have a different convention for (-1)^N,
-            # for even values qqNS is analytically continued
-            # as non-singlet. The accuracy is worst for large N
-            # due to the approximations of F functions.
             np.testing.assert_allclose(
-                aS3[1, 1], ref_val_qqNS[L][idx] + ref_val_qqPS[L][idx], rtol=8e-4
+                aS3[1, 1], ref_val_qqNS[L][idx] + ref_val_qqPS[L][idx], rtol=3e-3
```
felixhekhorn, line 185: the absolute size of `rtol` is still okay, I guess - but do you know how the factor of ~4 happened? (I mean you add ~400 lines so we should get better, no? (I know numerics now is more complicated, but then I wonder whether it was worth it ...))

giacomomagni: so this change is due to discrepancies with large-N moments (the test is checking `N=100`) if needed I can specify 2 different values...
I decide to keep the analytical structure of `A_qq` (up to weigth=5 harmonics) in order to fix the `eta=(-1)^n` business which before was not correct.
For compleness this [plot](https://user-images.githubusercontent.com/39065935/207275362-b0d422ff-06f7-4bc0-b0a7-fcd1d2308fa2.png) shows the relative difference as function of N for the parametrised part in case of even moments.

felixhekhorn: maybe two different `rtol` is a good idea ... concerning the plot: but then shouldn't we match with ~0.01e-2=1e-4? also are you sure you have the correct large N behaviour? the plot looks pretty much like a log (i.e. your approximation has the coefficient of ln(n) slightly off?)

giacomomagni: okay, I added a couple of more functions! Now it's even better than before :)

felixhekhorn: :tada: - re accuracy: okay, understood!

felixhekhorn: or maybe not :upside_down_face: - now test are failing ...

felixhekhorn (approved):

PR #179, opened 2022-12-09
Cache harmonics sums for the as2 anomalous dimensions

niclaurenti:
Dear @felixhekhorn @AleCandido @giacomomagni I'm opening this PR since in the anomalous dimensions at order as2 there are some harmonics sums that are used also in the QED sector. Then, as @felixhekhorn was suggesting we can use the ones that were cached in the computation of the QED anomalous dimensions. I didn't want to include these changes in the QED PR since it is already big enough.


felixhekhorn:
why not upgrade the existing `compute_cache`? it computes already a list of e.g. weight 3 sums? and as said we want to have the minimal number possible ...

giacomomagni:
> why not upgrade the existing `compute_cache`? it computes already a list of e.g. weight 3 sums? and as said we want to have the minimal number possible ...

To use that you would need to rotate some harmonics, for instance get rid of `S[m, (n-1)/2]` in favor or `S[-m,n]`, but I'd definitely do that. Note that also `g3` only enters in the anomalous dimensions through `S[-2,1,n]` and using the series definition you can simplify `g3(n+2)`

tests/eko/evolution_operator/test_init.py
```diff
@@ -28,183 +28,40 @@ def test_quad_ker(monkeypatch):
...
+        ((1, 0), br.non_singlet_pids_map["ns+"], 0, "", 0.0, 0.0),
+        ((3, 1), br.non_singlet_pids_map["ns+u"], 0, "", 0.0, 0.0),
+        ((1, 0), 100, 100, "", 0.123, 1.0),
+        ((1, 0), 100, 21, "", 0.0, 0.0),
+        ((1, 1), 100, 100, "iterate-exact", 0.123, 1.0),
+        ((1, 1), 100, 21, "iterate-exact", 0.123, 0.0),
+        ((1, 1), 10200, 10200, "iterate-exact", 0.123, 1.0),
+        ((1, 1), 10200, 10204, "iterate-exact", 0.123, 0.0),
+    ]
+    for p in params:
```
felixhekhorn, line 41: maybe better `for order,mode0,mode1,bla,bla in params`?

niclaurenti:
@felixhekhorn @AleCandido @giacomomagni Should I close this PR given that this will be implemented in #202 ?

felixhekhorn:
You can close it now if you're not actively using this branch or later when #202 is fully in place ...

niclaurenti:
> You can close it now if you're not actively using this branch or later when #202 is fully in place ...

Ok let's close it when #202 is completed

PR #189, opened 2023-01-09
Introduce Rust extension

### LINKED ISSUE #185: Split anomalous dimensions into a Rust crate
The motivations for this issue are mainly:
1. we are using inconsistent tooling in the theory pineline to achieve performant computations: here we use Numba, while PineAPPL is implemented in Rust for the same reason
2. Numba has some limitations in terms of supported features, from this point of view Rust is much more flexible (it is a whole language with its own ecosystem, while Numba is bound to be a "subset" forever)
3. Numba seems to be especially slow at compiling N3LO terms
4. It is much more difficult to have bindings for anomalous dimensions for other languages, that might be relevant since we have original pieces

I'm currently trying to outline a suitable layout for the new Rust module, I'm doing some attempts in this repo:
https://github.com/AleCandido/atuin

I already asked for `maturin` authors support, but a minimal skeleton (no further crates for bindings) is almost working:
https://github.com/PyO3/maturin/issues/1372

AleCandido:
Closes #185

The working layout reference is https://github.com/AleCandido/atuin/tree/main/full


felixhekhorn:
> Unfortunately, I'd like to do it incrementally, but I'm not sure I can: is there a way to use a compiled extension callable inside a Numba decorated function?

Mmm ... reading [here](https://numba.readthedocs.io/en/stable/reference/pysupported.html#cffi) I didn't understand whether `numba` can *read* CFFI calls, because [here](https://numba.readthedocs.io/en/stable/user/cfunc.html) I think they do the other way round: calling numba from C ... [this section on numpy](https://numpy.org/devdocs/reference/random/examples/numba_cffi.html) might also be interesting ...

AleCandido:
> Mmm ... reading [here](https://numba.readthedocs.io/en/stable/reference/pysupported.html#cffi) I didn't understand whether `numba` can _read_ CFFI calls, because [here](https://numba.readthedocs.io/en/stable/user/cfunc.html) I think they do the other way round: calling numba from C ... [this section on numpy](https://numpy.org/devdocs/reference/random/examples/numba_cffi.html) might also be interesting ...

All the resources are relevant, especially second and third.
We have the further problem that we are not directly binding to C ourselves. But I would expect it to be complicate anyhow (though worth having a look, not soon).

felixhekhorn:
Not using numba on `quad_ker` would cost us big time

t(numba)/t(raw) ~ 0.005 for 100 times calling quad_ker for 100 points

<details>
<summary>check script</summary>

```py
import timeit
import sys

import numba as nb
import numpy as np

from eko.evolution_operator import quad_ker

quad_ker_nb = nb.njit(cache=True)(quad_ker)

args = dict(order=(1, 0),
                mode0=br.non_singlet_pids_map["ns+"],
                mode1=0,
                method="",
                is_log=True,
                logx=0.1,
                areas=np.array([[1.0, 2.0, 3.0], [4.0, 5.0, 6.0]]),
                as_list=np.array([2.0, 1.0]),
                mu2_from=1.0,
                mu2_to=2.0,
                a_half=np.array([[1.5, 0.01]]),
                alphaem_running=False,
                nf=3,
                L=0,
                ev_op_iterations=1,
                ev_op_max_order=(1, 0),
                sv_mode=1,
                is_threshold=False,
                is_polarized=False,
                is_time_like=False,)

print("Compile numba")
quad_ker_nb(.6,**args)

us = np.linspace(.5, 1., 100)

def a(): return np.sum([quad_ker(u,**args) for u in us])
def b(): return np.sum([quad_ker_nb(u,**args) for u in us])

num = int(float(sys.argv[1]))

bb = timeit.timeit(b,number=num)
print("numba",bb)
aa = timeit.timeit(a,number=num)
print("raw",aa)
print("numba/raw",bb/aa)
```

</details>

together with this eko hack:

```diff
diff --git a/src/eko/evolution_operator/__init__.py b/src/eko/evolution_operator/__init__.py
index 335dfa4f..4e99f225 100644
--- a/src/eko/evolution_operator/__init__.py
+++ b/src/eko/evolution_operator/__init__.py
@@ -184,7 +184,7 @@ class QuadKerBase:
         return self.path.prefactor * pj * self.path.jac


-@nb.njit(cache=True)
+#@nb.njit(cache=True)
 def quad_ker(
     u,
     order,
```

AleCandido:
Is it possible that the class is making everything worse? Or is it just a Python call overhead?

To get more info, we could try to make a function with similar features of `quad_ker`, and see how much Numba gains.
But it is clear that just commenting the decorator is not an option, even if it had been the easiest of course (if it were working properly).

AleCandido:
Actually @felixhekhorn, we should make the test with NNLO: if we lose a lot of time on LO we don't care so much, since it is already fast enough. The big deal is how much we lose with already expensive computation: if it's negligible overhead, because the rest is much slower, for me it would be fine to spoil LO for a while.

AleCandido:
https://github.com/konchunas/pyrs

felixhekhorn:
https://github.com/cathay4t/librabc

crates/ekuad/src/lib.rs
```diff
@@ -5,14 +5,22 @@ use ekore;
 #[no_mangle]
 pub unsafe extern "C" fn quad_ker(x: f64, extra: *mut c_void) -> f64 {
     let ex = *(extra as *mut Extra);
-    ekore::ciao(x * ex.slope, ex.shift)
+    //ekore::ciao(x * ex.slope, ex.shift)
+    (ex.py)(x * ex.slope, ex.shift)
```
AleCandido, line 9: Perfect, thanks @felixhekhorn 

Here also something like `ekore::ciao` had a role, because it was showing that it could make use of functions defined in siblings' crates.

However, the composition is trivial, once they are all available functions, so I would not bother complicating the example, but at this point rather aim for the actual thing.

.pre-commit-config.yaml
```diff
@@ -37,6 +37,24 @@ repos:
         args: ["--add-ignore=D107,D105"]
         additional_dependencies:
           - toml
+  - repo: local
+    hooks:
+    - id: fmt-eko
+      name: fmt-eko
+      description: Format eko files with cargo fmt.
+      entry: cargo fmt --manifest-path crates/eko/Cargo.toml --
```
AleCandido, line 45: Unfortunately, this has a drawback: there is no `cargo` installed on the `pre-commit.ci`

felixhekhorn: that is correct - however, I don't believe that we can ever run rust there, so the best we can do is to skip this in the CI and rely on the developers to run ... (note that I copied this from an existing hook, so they do nothing more - but in fact less that's why I needed to copy in the first place)

AleCandido: How would you skip in the CI in practice?

felixhekhorn: https://github.com/NNPDF/eko/pull/189/commits/02b772f0979390a1aa048da89bf5f8ba840334bd

crates/ekore/Cargo.toml
```diff
@@ -9,3 +9,4 @@ license = "GPL-3.0-or-later"
 
 [dependencies]
 num = "0.4.1"
+bibliothek = "0.1.0"
```
AleCandido, line 12: I would make this a dev-dep, do you believe it would be a problem?
https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#development-dependencies

AleCandido: Nono, I also prefer the
```
:cite:`bla2023`
```
option (even though it looks more like ReST than MD).

My point was mainly about making it a development dependency, because you do not need to compile the actual library, but only for the docs.
I'm not sure if it would fully work (e.g. on crates.io), but it should be sensible, so it might. I'd just try.

AleCandido: > all in all I have to say, I'm disappointed by Rust - not having a built-in citation system is a no go for academics; we want to and have to honour other people ...

Fair enough :)

AleCandido:
> @AleCandido @giacomomagni this is now at the same stage as #246 - i.e. LO FFNS unpol. LHA benchmark is back working

Thank you very much for all the work done!

> Let me repeat the open points that are also found on the other side:
> 
>     * deployment: create the correct libraries for the different OS/arch and deploy them to PyPI - here we have maturin, so it should be easier (we should be able to copy pineappl), but remains to be proven ... @AleCandido can you help with this?

This I can do: essentially is the exact same of PineAPPLpy, since I would only distribute EKO, and not compiled `ekore` (yet, at least). This means we will rely on `maturin`.

>     * unit test: just need to be ported

Indeed, you can keep going with the stand-alone modules in `crates/ekore/tests/`, so we'd only need a translation. You can check if the equivalent of `conftest.py` is available, but I think so (even though fixtures might be absent, so you may have to use regular functions, but I'm not that expert).

>     * documentation: I need to find out how to run locally and then start finding out how to tweak and improve (like how to display formulae)

The `ekore` internal documentation can be compiled with `cargo doc` (you can add `--open` to directly show them in the browser). No need to deploy them, crates.io is self-compiling the docs (it's like if RTD were a part of PyPI, automatically running on deployment).

>     * find a way to translate N3LO stuff (up to N2LO, I think, I could do by hand, but not beyond ...)

This we can discuss together, but I'd be happy to delegate to @giacomomagni, if he agrees to do it.

> we should also add a bunch of poe commands for ease of use ... e.g.
> 
>     * what is the correct compile command? I currently run `pip install -e crates/eko/` as is done in your `try.sh` @AleCandido

It is a `pyproject.toml` package, so `pip` is aware and you can compile that way. However, we could also try to add a path dependency in Poetry. We should check how it's being recast when compiling with `poetry build`: if it resolves the version from the `pyproject.toml` of the other project, we won't need anything else.

>     * docs + unit tests

This is simple enough, just choose suitable names :)
 
> about speed: we achieve a similar level here, maybe even another 10% faster: ~8s

Great!

> @AleCandido as discussed the impact on core eko is in a single commit: [731832c](https://github.com/NNPDF/eko/commit/731832cb860dfb86df2d71f77e8bc56f305c4012) - the moment we loose the green ticks 🙃

As agreed, I will proceed splitting that single commit out in a new long-standing branch (to be merged at the end of the transition), and merge the rest in `master` as soon as it is approved (I will review tomorrow or by the end of the week, and will leave the final approval to @giacomomagni: since it's going to be a major change and design choice, a fresh look would be worth).

crates/ekore/Cargo.toml
felixhekhorn: > even though it looks more like ReST than MD

it *is* the copied rst and thus not working :upside_down_face: the actual proposal is
```rust
/// Implements Eq. (3.4) of [Moch:2004pa][cite Moch:2004pa].
///
/// [cite Moch:2004pa]: ./../../../../bib/index.html#moch2004pa
```
plus the `bib.rs`

felixhekhorn: But I don't like all this non-sense:
grafik
the ratio relevant information/non-sense is 2/27

AleCandido: Then we could try patching `bibliothek` to use functions rather than structs. A function has no implementation-like add-on, so it should not be cluttered, e.g. [`gamma_gg`](file:///media/alessandro/moneybin/Projects/NNPDF/eko/crates/target/doc/ekore/anomalous_dimensions/unpolarized/spacelike/as1/fn.gamma_gg.html)

crates/ekore/Cargo.toml
```diff
@@ -9,4 +9,6 @@ license = "GPL-3.0-or-later"
 
 [dependencies]
 num = "0.4.1"
-bibliothek = "0.1.0"
+bibliothek = { version = "0.1.0" }
+# bibliothek = { version = "0.2.0", path = "../bibliothek" }
```
AleCandido, line 13: Have you tried?

giacomomagni:
LGTM. 
I can take care of translating the ad expression to rust.
Do we plan to complete this update before or after the new 4.0 papers?

AleCandido:
> LGTM. I can take care of translating the ad expression to rust. Do we plan to complete this update before or after the new 4.0 papers?

Most likely the papers will be written with the old version of EKO. In practice, it should make no difference: we're changing a lot of the internals, but the result should be the same bit-by-bit.
Since the papers only care about the fits, and the fits only care about (the FK tables that only care about) the resulting operators, it makes no difference. There is no hurry to finish (it is not a priority), but as soon as we are ready I'd start using this.

crates/ekore/Cargo.toml
felixhekhorn: I asked the author for a new tag ...

crates/ekore/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+name = "ekore"
+version = "0.1.0"
+edition = "2021"
+description = "EKO expressions"
+license = "GPL-3.0-or-later"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
+
+[lib]
+doctest = false
```
felixhekhorn, line 11: @AleCandido I found no other way to make `cargo test` run ... the original error is
```
   Doc-tests ekore
error: proc macro panicked
  --> /home/felix/Physik/N3PDF/EKO/eko/crates/ekore/src/lib.rs:13:5
   |
13 |     bibliothek::include_bibtex!("refs.bib");
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: message: called `Result::unwrap()` on an `Err` value: Os { code: 2, kind: NotFound, message: "No such file or directory" }
```
the problem seems that `cargo doc` and `cargo test` run in different directories ... any idea?

AleCandido: I would try to specify the absolute path with [`env!`](https://doc.rust-lang.org/std/macro.env.html) and [`CARGO_MANIFEST_DIR`](https://doc.rust-lang.org/cargo/reference/environment-variables.html#environment-variables-cargo-sets-for-build-scripts) environment variable.

AleCandido: I.e. if you invoke:
```rs
env!(`CARGO_MANIFEST_DIR`)
```
it should be replaced at build time with the directory containing the `Cargo.toml` of the package being built, and then you concatenate with `"/refs.bib"` and it should work (I hope - I haven't tested `cargo doc` and `cargo test`, but you will find out immediately).

AleCandido: To anticipate you some discoveries about strings:
```rs
env!("CARGO_MANIFEST_DIR").to_owned() + "/refs.bib"
```
in case you never concatenated strings before.

felixhekhorn:
```
error: proc macro panicked
  --> ekore/src/lib.rs:12:5
   |
12 |     bibliothek::include_bibtex!((env!("CARGO_MANIFEST_DIR").to_owned() + "/refs.bib"));
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = help: message: expected path to .bib file as string literal as first parameter: InvalidToken { expected: StringLit, actual: Group, span: Two(#0 bytes(335..388)) }

```

felixhekhorn:
> As agreed, I will proceed splitting that single commit out in a new long-standing branch (to be merged at the end of the transition), and merge the rest in `master` as soon as it is approved (I will review tomorrow or by the end of the week, and will leave the final approval to @giacomomagni: since it's going to be a major change and design choice, a fresh look would be worth).

I think I prefer rather than another branch just a single `.patch` file (in the end that is all we should need) which can be applied whenever we need to work on this project. To activate the rust compilation just type
```sh
patch -p1 <src/eko/evolution_operator/__init__.py.patch
```
(instead to generate said path file:
```sh
git diff --merge-base master src/eko/evolution_operator/__init__.py > src/eko/evolution_operator/__init__.py.patch
```
)

crates/ekore/Cargo.toml
AleCandido: Ok, the reason of this is that it is most likely using something like `fmt!()` (or one of its relatives), that actually requires a string literal.

In general, macros (including `macro_rules`, those that look like functions just ending in `!`) are expanded at compiled time (of course), but they might leave the arguments there as they are for some other functions. Instead, `fmt!` manipulates the literal, so it has to be a literal.

AleCandido: At this point, I would give up on `bibliothek` completely, and just inline the result of its evaluation (i.e. let's maintain a `.rs` file instead of `.bib` file).

crates/ekore/src/harmonics/polygamma.rs
```diff
@@ -0,0 +1,344 @@
...
+#[cfg(test)]
+mod tests {
+    use crate::harmonics::polygamma::cern_polygamma;
+    use float_cmp::assert_approx_eq;
+    use num::complex::Complex;
+
+    #[test]
+    fn fortran() {
+        const ZS: [Complex<f64>; 9] = [
+            Complex::<f64> { re: 1.0, im: 0. },
```
AleCandido, line 131: I know that now it's a bit too late, but it could have been slightly helpful to define a `complex!` macro, to use like this:
```rs
complex![1.0, 0.]
```
rather than the whole struct every time...

crates/try.sh
AleCandido, line 1: This one we can drop at this point, isn't it?

crates/integrate.py
AleCandido, line 1: Even this one should be dropped, we should not need the example any longer. Or am I wrong?

crates/ekore/src/harmonics/cache.rs
```diff
@@ -0,0 +1,48 @@
+//! Cache harmonic sums for given Mellin N.
+
+use num::complex::Complex;
+use std::collections::HashMap;
```
AleCandido, line 4: In case you care about the performances of this object, some (i.e. a relevant amount of) people are using this library instead of standard library
https://crates.io/crates/hashbrown

AleCandido (commented):
Some simple comments

crates/try.sh
felixhekhorn: 4221c9a3

crates/integrate.py
felixhekhorn: 4221c9a3

crates/ekore/src/harmonics/cache.rs
felixhekhorn: 1fa54f3c

crates/ekore/Cargo.toml
felixhekhorn: let's give the author a chance https://codeberg.org/luleyleo/bibliothek/issues/5 - else we do as you say ...

crates/ekore/src/harmonics/polygamma.rs
felixhekhorn: done in [408b5be](https://github.com/NNPDF/eko/pull/189/commits/408b5bededed5c8b27f6c75525207e1010bf6501) (although in practice it is used only inside tests, so I'm not sure about the correct scope)

crates/ekore/Cargo.toml
felixhekhorn: 941d64f8

felixhekhorn:
[63eb191](https://github.com/NNPDF/eko/pull/189/commits/63eb191ffa5130e88a8bd205165b3283df09a274) adds a quick and dirty bibtex -> fake-crate parser, which does it job for the moment. We can and will update it the moment we need it.

If you agree @AleCandido @giacomomagni this can be merged. (just as a reminder: this does not touch the Python program in any way and if you want to use this new feature you have to apply the provided patch)

src/eko/evolution_operator/quad_ker.py
```diff
@@ -0,0 +1,367 @@
...
+#             mu2_to,
+#         )
+#         if sv_mode == sv.Modes.expanded and not is_threshold:
+#             ker = (
+#                 sv.expanded.non_singlet_variation_qed(
+#                     gamma_ns, as_list[-1], a_half[-1][1], alphaem_running, order, nf, L
+#                 )
+#                 * ker
+#             )
+#     return ker
```
AleCandido, line 367: Do we need all this commented code?

pyproject.toml
```diff
@@ -1,6 +1,20 @@
 [build-system]
-requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning"]
-build-backend = "poetry.core.masonry.api"
+requires = ["maturin>=0.14.8,<0.15", "poetry-dynamic-versioning"]
+build-backend = "maturin"
```
AleCandido, line 3: This one might be the most delicate change, since it is actually affecting how we build the package.

However, we should just test once making a pseudo-tag locally (that is a proper tag, but avoiding pushing it) and building with `pip wheel` (that should be what we should use in the workflow, but we are not).
https://github.com/NNPDF/workflows/blob/818e8d74406df7c51992ea3b9377441a401a7087/.github/workflows/python-poetry-pypi.yml#L66-L69

So, updating the workflow with `pip wheel` (that should keep working for Poetry as well) is a requirement for merging this.

crates/make_bib.py
AleCandido, line 1: Thanks for this.

We don't care too much of the details, but it is simple that can be maintained easily.

crates/eko/pyproject.toml
```diff
@@ -0,0 +1,16 @@
...
+requires-python = ">=3.8"
+classifiers = [
+  "Programming Language :: Rust",
+  "Programming Language :: Python :: Implementation :: CPython",
+  "Programming Language :: Python :: Implementation :: PyPy",
+]
+dependencies = ["cffi"]
+
+[tool.maturin]
+bindings = "cffi"
```
AleCandido, line 16: Did we decide for CFFI eventually?

AleCandido:
Btw, I won't be able to approve, since I opened the PR...

pyproject.toml
felixhekhorn: you're right - this is a delicate change and most likely we will break the package (even without touching the code); I suggest we do another `.patch` file

crates/eko/pyproject.toml
AleCandido: Ok, let's say fine for the time being. I wanted to use PyO3, because we'd avoid a further dependency, and I believe it was possible to make it work with Numba. But it is a time-consuming task, it's not the best timing.

pyproject.toml
felixhekhorn: But then I don't understand? how can we deal with the maturin backend without having maturin in game?

felixhekhorn: without rust we don't need and want maturin as backend, no? (just to be clear: backends means the one who actually builds the wheel (creating the folder blabla))

AleCandido: Without Rust we don't need `maturin`. But I see no reason why we don't want it, since it could work for both.

giacomomagni (approved):

giacomomagni:
@felixhekhorn can we merge here, so we can go on with it ?

felixhekhorn:
> @felixhekhorn can we merge here, so we can go on with it ?

the [packaging issue](https://github.com/NNPDF/eko/pull/189#discussion_r1410943979) with maturin needs to be resolved first ... I'm still tempted by postponing the problem by creating another `.patch` file as I'm not as familiar with Python packaging as @AleCandido :innocent: ... as far as I understood we would also need to touch the workflows which would ripple through the whole NNPDF packages ...

AleCandido:
> the [packaging issue](https://github.com/NNPDF/eko/pull/189#discussion_r1410943979) with maturin needs to be resolved first

Correct

> as far as I understood we would also need to touch the workflows which would ripple through the whole NNPDF packages ...

Almost correct: the current situation is no good, because the dependencies are specified under the Poetry-specific sections, but the build backend is coming from Maturin, with the effect of forgetting EKO dependencies.

However, the part I was worried about might be fake, since `poetry build` is only the frontend, and if `maturin` is specified as backend that one will be used (not `poetry-core`, that indeed has to be specified separately, even when using the `poetry` frontend to build).

So, you might not need any change to the workflow at all for deploying, but I believe you'd need to duplicate the dependencies if you want to use the testing workflow in EKO (because it will install the environment with Poetry, which is not aware of the deps if they are not specified in its sections, that of course are not read by Maturin, which is reading the standard ones...).

> I'm still tempted by postponing the problem by creating another `.patch`

Of course, that is an option, with the benefit of diverging less from `main` as time passes. But the content of this PR is mostly to everything in `main`, since it mostly happens in `crates/` and what is outside that folder is actually mostly the packaging part (or the part in Python, already contained in a patch).
Thus, the benefit is mostly limited to close a PR. Which could also be enough, if you wish.

felixhekhorn:
I opted for the lazy option of adding another `.patch` file and postponing the problem ( any help is appreciated :innocent: ) this way we can still close the PR within almost a year :upside_down_face: 

status:
- LO unpolarized QCD LHA is back working
- we gained some speed
- opt-in for the new feature by executing `rustify.sh` (which will patch 3 files)
- the patches will need to be updated in the future if the interface changes (as is happening in #335 )
- the new feature is only working locally, i.e. not in workflows or PyPI - this has to be resolved in a future PR

if you agree @AleCandido and @giacomomagni let's merge and push for the rest of ekore

felixhekhorn (approved):

pyproject.toml
felixhekhorn: postponed to #340

felixhekhorn:
an attempt using PyO3 is https://github.com/NNPDF/eko/commit/357cb91ecf5b7ffed6ce905c6fb1e4a8e36f0135

Issue #163, opened 2022-11-15
LHA benchmark bot


felixhekhorn:
I'd like to have a "LHA benchmark bot" that runs upon adding a label (e.g. `run-lha-benchmark`) and that runs all LHA benchmarks, since we need to conserve them. There is no need to run them on every commit, but I'd like to run at the end of each PR.

The idea is, of course, stolen from nnpdf.

For the moment I asked people [expli](https://github.com/NNPDF/eko/pull/148#issuecomment-1315393778) - [citly](https://github.com/NNPDF/eko/pull/135#issuecomment-1315396705), but this could be automized ...

AleCandido:
All the benchmarks run can be automated with very little effort. The only problem was how to evaluate the output (when it should raise an error/warning).

I don't know what you mean by "bot", but if you mean "like fitbot" that is only a workflow, like ours, simply triggered by a label event. We can do it as well, or even do something better.
Still, the "end of a PR" is not an event, just because it is ill-defined. Instead of the label, you can trigger by the button coming with `workflow_dispatch`, it is just the same, and the straight-forward way to get a button (no magic label to know, it is the officially documented button).

felixhekhorn:
- Indeed something like the "fitbot" ...
- also a button is just fine, and yes, exactly because it is ill-defined I'd like to keep it manual
- and the output can for now be just the usual print by ekomark that has to be validated by a user

felixhekhorn:
a workflow can expose some assets after he has run, right?

AleCandido:
If you acknowledge the impossibility to fail, the alternative is to make a report, like vp ones: you need to compare to something, otherwise is difficult to evaluate, but I believe you can compare against PR base.

So, you need a workflow that:
- trigger on `workflow_dispatch`
- launch two jobs:
  1. checkout PR branch, and run benchmark runners
  2. checkout PR base, and run benchmark runners (this one can be cached with `actions/cache`)
- each job upload the results as an artifact
- a third job is fired when the other two are completed, download the artifacts, produce a comparison, and reupload it as an artifact
- after that the job (or fourth one) post a message on the PR with the link to the artifact for the comparison

AleCandido:
Artifacts for the branch benchmark result can have a small retention period (couple of days), while the other two you keep for a longer time (the base for caching purpose, the report for checking). But you still don't need both of them after PR merging, and if PR continues for long, it is worth rerunning (at least you will have to rebase)

AleCandido:
[...]

Issue #155, opened 2022-10-19
Tutorial problems


cschwan:
I'm trying to follow the [first tutorial for EKO](https://eko.readthedocs.io/en/latest/overview/tutorials/dglap.html), but I ran into the following problem:

```
In [2]: from banana.data.theories import default_card as th_card
---------------------------------------------------------------------------
ModuleNotFoundError                       Traceback (most recent call last)
Cell In [2], line 1
----> 1 from banana.data.theories import default_card as th_card

ModuleNotFoundError: No module named 'banana.data'
```

Note that I also ran `pip install banana` without which it fails trivially.

felixhekhorn:
Conjectured solution: we need: `pip install eko[mark]` instead of the `pip install eko` written [here](https://eko.readthedocs.io/en/latest/overview/tutorials/index.html)

> Note that I also ran pip install banana without which it fails trivially.

_our_ `banana` module is actually inside the `banana-hep` package - which gets installed by ekomark: https://github.com/NNPDF/eko/blob/01efee1d27d215e895dd055debff8905af3d548c/pyproject.toml#L70

cschwan:
Installing ekomark doesn't work on my computer:

```bash
$ pip install ekomark
ERROR: Could not find a version that satisfies the requirement ekomark
ERROR: No matching distribution found for ekomark
```

felixhekhorn:
yes, sorry I was unclear again ... the command you have to run is: `pip install eko[mark]`

this will (among other stuff) install the `ekomark` _module_ and the `banana-hep` _package_ (containing the `banana` _module_)

cschwan:
Oh I see - I didn't realize that square brackets are allowed as a Python package name - my mistake! That solves the problem described above for me.

AleCandido:
Indeed, the proper way to update the docs should be:
```sh
pip install 'eko[mark]'
```
since some shells (e.g. zsh) interpret the square brackets...

PR #190, opened 2023-01-09
Drop `develop`

AleCandido:
In order to simplify the workflow for the newcomers, I propose to finally drop `develop` branch, together with [Git flow](https://nvie.com/posts/a-successful-git-branching-model/) altogether (i.e. stop prefixing branches with `feature/` and similar, and complex merge strategies).

In practice, not much will change:
- I might need to update some workflows (but maybe not)
- `master` won't be synced any longer with latest tag: tags are tags, and each correspond to a version, but there is no branch following them
- `master` will be the new `develop`, i.e. the main branch we are working on (as it usually is)

Feedback from people already working on this repo is appreciated.


AleCandido:
All the commits here are listed by chance: we can merge them as soon as we are ready, with the old workflow, independently of this PR.

The PR will be merged only if and when we will have decided to drop `develop` branch.

felixhekhorn:
I wonder how to enforce the idea of #163 - i.e. run benchmarks (in a first stage LHA, later even more) before releasing ... else I guess it is fine ...

giacomomagni:
I'm happy with this. Do we plan to have a working release once this PR is merged? 

> I wonder how to enforce the idea of #163 - i.e. run benchmarks (in a first stage LHA, later even more) before releasing ... else I guess it is fine ...

That would be really great!

AleCandido:
Just to tell also here what I already wrote to @felixhekhorn in private: I perfectly know how to do it
- use the same container of isobench
- run the benchmark runners (without firing any error in any case)
- upload the resulting db as an artifact
- (to improve) we add a function to identify pathologies, and one to compare with old db for regressions: no one will raise errors, since we have no reliable way to claim an error, but they are useful to skim the output

However, at the moment we're pretty busy with more pressing tasks, for this reason I proposed to postpone this further.
We already have a dedicated issue since some time #155

AleCandido:
I mistakenly removed draft mode and currently unable to restore. There is no meaning behind it

felixhekhorn:
> run benchmarks (in a first stage LHA, later even more) before releasing

here I was more worried about the formal requirement of having some sort of green tick than about the actual implementation

AleCandido:
> here I was more worried about the formal requirement of having some sort of green tick than about the actual implementation

If you don't take care of the implementation, you will always have a green tick or always a red cross, regardless of what is happening inside.
The only part you can skip is the db upload. But if something is so broken to make benchmarks crashing, it is extremely likely it would make also unit tests or isobenchs (i.e. integration tests...) fail before, by crashing.

PR #191, opened 2023-01-09
Fix scale variations expanded

### LINKED ISSUE #188: Remove alpha_raw from the game since it is not needed and wrong
We introduced `alpha_raw` because we thought that was necessary to correctly compute factorization scale variations. Now we realized that it is not and we should compute the scale variation kernels using `alpha` evaluated at the factorization scale (as we used to do in the past). So we need to revert and remove completely `alpha_raw` from the game. 
@AleCandido @felixhekhorn

andreab1997:
Here I just want to do two things:

- [x] Put a global minus sign to the sv kernels (as was done in #151 that will be closed) 
- [x] Remove `alpha_raw` from the game since we decided to write everything as an expansion in `alpha(xi_F * Q)` (Close #188)

@felixhekhorn @AleCandido @giacomomagni


andreab1997:
@giacomomagni I believe you worked on the test that is currently failing in the other PR. Can you remind me what your conclusions were?

giacomomagni:
The 

> @giacomomagni I believe you worked on the test that is currently failing in the other PR. Can you remind me what your conclusions were? (In order to fix here)

the idea of the test was to quantify the difference between expanded and exponential (B vs A) in the case of liarized and truncated solution. So to fix it we should find where the minus sign come from. If I recall correctly your analysis shows it was due to a shift `k -> -k `, right?

andreab1997:
> The
> 
> > @giacomomagni I believe you worked on the test that is currently failing in the other PR. Can you remind me what your conclusions were? (In order to fix here)
> 
> the idea of the test was to quantify the difference between expanded and exponential (B vs A) in the case of liarized and truncated solution. So to fix it we should find where the minus sign come from. If I recall correctly your analysis shows it was due to a shift `k -> -k `, right?

Yes indeed. At a certain point I had a doubt about the relative sign between the term `gamma_0*gamma_0` and `gamma_0*beta_0` but now I am convinced that is correct as it is (if we keep expanding everything in terms of `alpha_s(xi_F*Q)`). So yes, the shift should just be `k _> -k`

felixhekhorn:
> If I recall correctly your analysis shows it was due to a shift k -> -k , right?

ehm, if you mean by k what is called L here:
https://github.com/NNPDF/eko/blob/20008591d5a413b270f4d9a5cf890a4d5608e09f/src/eko/scale_variations/expanded.py#L54
no - this is only true at LO SV = NLO AD, but not in the case of the snippet above

andreab1997:
> > If I recall correctly your analysis shows it was due to a shift k -> -k , right?
> 
> ehm, if you mean by k what is called L here:
> 
> https://github.com/NNPDF/eko/blob/20008591d5a413b270f4d9a5cf890a4d5608e09f/src/eko/scale_variations/expanded.py#L54
> 
> 
> no - this is only true at LO SV = NLO AD, but not in the case of the snippet above

Sorry, my fault. With `K` I mean the sv kernel not the log. I should have been clearer.

AleCandido:
@andreab1997 since we finally sorted out the discussion about $\alpha_s$, can you add a summary to the docs, and (if relevant) redirect to Pineko docs for a deeper discussion (where the whole picture of scale variations and their implementation, ingredients and schemes, will be described).

andreab1997:
Now the test should be ok. It was enough to remove the term that we know is generated by the change of `alpha` from the diff to make it pass (I left a comment but if you want I can remove it). I would say that this is the final proof that the implementation (at least up to NNLO) is correct. Let me know if you agree with this. I will soon work also to check N3LO (now that everything is clear should be simpler). I will also update the docs before merging this. @giacomomagni @felixhekhorn @AleCandido

giacomomagni:
> Now the test should be ok. It was enough to remove the term that we know is generated by the change of `alpha` from the diff to make it pass (I left a comment but if you want I can remove it). I would say that this is the final proof that the implementation (at least up to NNLO) is correct. Let me know if you agree with this. I will soon work also to check N3LO (now that everything is clear should be simpler). I will also update the docs before merging this. @giacomomagni @felixhekhorn @AleCandido

If the logic is the same you would need to subtract: `- a1**3 * b0**2 * g[0] * k**3`, from the full difference A-B, right?

andreab1997:
> If the logic is the same at N3LO you would need to subtract: `- a1**3 * b0**2 * g[0] * k**3`, from the full difference A-B, right?

Not exactly, because at N3LO you have a lot other terms appearing. In particular, all the `alpha^2` terms get a contribution at N3LO (while at NNLO only the `alpha` terms get a contribution from the change of alpha, which is only one term).

AleCandido (approved):

giacomomagni:
To remember what has to be done: 
* [ ] fix names and conventions as [here](https://github.com/NNPDF/eko/blob/5b6c3fecf87706c227e1d9094d9e7fc9b6f24052/src/eko/evolution_operator/__init__.py#L289)
* [ ] implement the actual tests of differences A-B  or similar we have derived in Mathematica, at least at NLO

andreab1997:
Can we merge this now?

src/eko/evolution_operator/__init__.py
```diff
@@ -291,7 +286,7 @@ def grid_size(self):
         """Return the grid size."""
         return self.int_disp.xgrid.size
 
-    def mur2_shift(self, q2):
+    def scheme_exponentiated_shift(self, q2):
```
AleCandido, line 289: Since this function is in a file not dealing with scale variations, which "scheme" you are referring to is a bit ambiguous.

I would suggest a bit more explicit alternative:
```suggestion
    def sv_exponentiated_shift(self, q2):
```

The other option is to put this function in the `sv` module, instead of using `self` you can just have:
```py
def scheme_exponentiated_shift(mu2, xif2, mode):
```
(a good chance to rename this `q2` to `mu2`, and once in the `sv` module there is no ambiguity any longer about which `mode` you are referring to)

Of course also:
```py
def exponentiated_shift(mu2, xif2, mode):
```
works perfectly: it is crystal clear what `exponentiated` means in this context :)

tests/eko/scale_variations/test_expanded.py
```diff
@@ -43,13 +43,12 @@ def test_singlet_sv_dispacher():
...
-    and not in the size of the `fact_to_ren` itself.
-    However in our implementation the exponentiated mode depends on
-    the actual value of a0 and a1, since the evolution integral in :math:`\alpha_s`
-    is evaluated. Thus this test ratio :math:`(ker_A - ker_B)/ker_{unv}`
-    still contains a dependency on `fact_to_ren`.
+    We test that the quantity :math:`(ker_A - ker_B)/ker_{unv}`
+
+    Note this is NOT the real difference between scheme expanded
+    and exponentiated since here we don't take into account the
+    shifts in path length and :math:`\alpha_s` values.
```
felixhekhorn, line 50: Why don't we also shift the alpha values, as we should? we know how to do it (it's written in the notebook)

tests/eko/scale_variations/test_expanded.py
```diff
@@ -64,20 +63,50 @@ def scheme_diff(g, k, pto, is_singlet):
         the accuracy of singlet quantities is slightly worse.
         """
         if pto[0] >= 2:
```
felixhekhorn, line 65: do we really want a `>=` here? the difference strictly depends on the order - also I don't believe we want the `diff +=`, but instead `diff =`

tests/eko/scale_variations/test_expanded.py
```diff
@@ -64,20 +63,50 @@ def scheme_diff(g, k, pto, is_singlet):
...
-            diff = g[0] * k * a0
+            diff = g[0] * k * a0 - 2 * a1 * k * g[0]
         if pto[0] >= 3:
             b0 = beta_qcd_as2(nf)
             g02 = g[0] @ g[0] if is_singlet else g[0] ** 2
-            diff += a0**2 * g[1] * k - k**2 * (
-                1 / 2 * a0**2 * b0 * g[0] + a1 * a0 * g02 - 1 / 2 * a0**2 * g02
+            diff += (
```
felixhekhorn, line 70: this is the `+=` that I meant above

tests/eko/scale_variations/test_expanded.py
```diff
@@ -64,20 +63,50 @@ def scheme_diff(g, k, pto, is_singlet):
         the accuracy of singlet quantities is slightly worse.
         """
         if pto[0] >= 2:
-            diff = g[0] * k * a0
+            diff = g[0] * k * a0 - 2 * a1 * k * g[0]
```
felixhekhorn, line 66: This is not higher order (~a with PTO=2) and not what we found in the notebook: $L \left(\text{a0}^2 \beta [0]-(\text{a0}-a[t])^2 \gamma [0]\right) (\text{b1} \gamma [0]-\gamma [1])$

felixhekhorn (changes requested):

tests/eko/scale_variations/test_expanded.py
giacomomagni: This is just laziness. As mentioned above the tests will not test the real HO difference A-B, since we have verified that is too complex.

tests/eko/scale_variations/test_expanded.py
giacomomagni: As the test is build, this is correct with `+=`.

tests/eko/scale_variations/test_expanded.py
giacomomagni: Indeed this is not HO because the test is not doing the full A vs B, so it's not what we have derived in that matematica NB.

andreab1997:
In any case I would merge this. We already checked the implemented expression for scale_varaitions with the notebook. Writing a sensible test that uses the results we derived can take a lot of time so I would open an issue for that but I would not block this PR for the test (since it is useful for several projects now). Do you agree? @giacomomagni @felixhekhorn @AleCandido

andreab1997:
Note that the issue about this test has been opened (#201 )

src/eko/scale_variations/expanded.py
```diff
@@ -52,11 +52,12 @@ def variation_as2(gamma, L, beta0, g0e2):
         variation at |NNLO|
     """
     return -gamma[1] * L + 1.0 / 2.0 * (beta0 * gamma[0] + g0e2) * L**2
+    # changing alpha will change the sign of beta*gamma0
```
AleCandido, line 55: This comment should be expanded (like this is only understandable if you know all the background, which we will forget ourselves in a while), and possibly included in the docstring with a suitable reference.

After all, it was the main source of problems.

AleCandido (commented):
I just have one comment about documentation, for the rest it looks ok.

However, I believe docs to be crucial here, since the problem is a sophisticated one, mainly about physics, more than code.

You don't have to document all in EKO, if the documentation is Pineko just add a reference to that location. But someone investigating EKO should be able (at least through redirection) where the choices made stem from

AleCandido:
> In any case I would merge this.

@andreab1997 I only had a tiny comment about documentation, and for the rest is fine. But you should reply to @felixhekhorn comments before merging.

andreab1997:
> > In any case I would merge this.
> 
> @andreab1997 I only had a tiny comment about documentation, and for the rest is fine. But you should reply to @felixhekhorn comments before merging.

Ok thank you! About @felixhekhorn comments I believe @giacomomagni answered but, in any case, they are all about the test and we agreed (I believe) to open an issue for that (I already did) because for sure the test as it is is not optimal and we will need to change.

AleCandido:
> > 
> 
> Ok thank you! About @felixhekhorn comments I believe @giacomomagni answered but, in any case, they are all about the test and we agreed (I believe) to open an issue for that (I already did) because for sure the test as it is is not optimal and we will need to change.

If they are addressed (or delayed on purpose), people participating or who is taking the decision should resolve them. If it is just a matter of clicking a button it takes no time, but it is an explicit action to say we already took a decision.

andreab1997:
> 

Yes I was waiting for @felixhekhorn answer before closing the conversations :)

felixhekhorn (approved):
As long as we do the correct test (maybe we can rename the current version - since it is not doing what is promised) at some point (the sooner the better), we can even merge to move forward

andreab1997:
So, Can I merge this?

PR #194, opened 2023-01-11
Reorganize package content

felixhekhorn:
The idea is to split the hard math away from the other administrative stuff (such as how to join operators etc.). This would be a small step towards #185 and eventually #189.
- This might also facilitate the implementation of #170 that would only act on this new side @adrianneschauss . Same would be true for FF (in the first stage) @t7phy .
- also #148 would mostly fall into the new side
- but instead #135 should fall more towards the other, but in the current state actually does some further fixes on the math side


AleCandido:
This will also simplify #185, eventually.
In theory, the scope of that PR would become to move `ekore` and only it to Rust (in practice we might need several further adjustments because of `numba`)

AleCandido:
@felixhekhorn do you actually want a review now, or are you just attracting our attention on the topic?

felixhekhorn:
no, not yet - I want to attempt the move into `unpolarized/space_like/*` first ... I'll mark this PR

AleCandido:
Maybe we should make a poll about how to sort external dimensions, currently:

1. anom-dim vs matching
2. polarization
3. time vs space
4. pto
5. flavor

I'm positive about keeping 4 and 5 in their current position, encoding them in modules and functions, but I'm quite unopinionated on the first three.

giacomomagni (dismissed):
Test are passing, so these changes should be enough :)

felixhekhorn:
- the most content of this PR is just a moving of files and the associated renames
- for the OME I had to alter a bit the structure to split the hard math away (which I called OME - I hope that is correct); @giacomomagni could you have a look please?
- we can decide to lift the bunch of `raise NotImplementedError` into the calling functions inside `eko` to save us some docstrings ...

For review: please @AleCandido @niclaurenti @giacomomagni have a look now
- @enocera please feel free to have a look and leave a comment or review
- @adrianneschauss #195 should go smooth on top of this since I took https://github.com/NNPDF/eko/pull/194/commits/bb6af22675ab83c28f79aa1d9853dd9a63ffb375 from there
- @t7phy I think you can also build on top of this and start overwriting `ekore.anomalous_dimensions.unpolarized.space_like.gamma_ns` and the singlet counter part

niclaurenti (approved):

src/eko/beta.py
```diff
@@ -7,7 +7,7 @@
 import numba as nb
 
 from . import constants
-from .harmonics.constants import zeta3
+from ekore.harmonics.constants import zeta3
```
AleCandido, line 10: This one we said to merge into the global `constants` module, isn't it?

src/eko/evolution_operator/__init__.py
```diff
@@ -182,7 +192,16 @@ def quad_ker(
...
+        if is_polarized:
+            if is_time_like:
+                gamma_singlet = ad_pt.gamma_singlet(order, ker_base.n, nf)
+            else:
+                gamma_singlet = ad_ps.gamma_singlet(order, ker_base.n, nf)
+        else:
+            if is_time_like:
+                gamma_singlet = ad_ut.gamma_singlet(order, ker_base.n, nf)
+            else:
+                gamma_singlet = ad_us.gamma_singlet(order, ker_base.n, nf)
```
AleCandido, line 204: Here I'd be tempted to go for dynamic imports, to avoid the incredibly long (and repeated) branching. More or less as we are doing for yadism.

What do you think? @felixhekhorn

felixhekhorn: I didn't try, but I'm quite convinced numba doesn't like dynamic imports since it needs to know ahead of time which functions are inside and need a compilation

AleCandido: You're definitely right, I keep forgetting the point where Numba kicks in...

src/eko/beta.py
felixhekhorn: I suggest to postpone to #202

PR #196, opened 2023-01-17
Use filesystem safe b64

felixhekhorn:
As @andreab1997 found out [`base64.b64encode`](https://docs.python.org/3/library/base64.html?highlight=base64#base64.b64encode) is not safe for usage with filesystem operations due to the presence of `/` in the default base64 alphabet, but [`base64.urlsafe_b64encode`](https://docs.python.org/3/library/base64.html?highlight=base64#base64.urlsafe_b64encode) is


andreab1997 (approved):
Ok, is this fixing both the "+" and the "/"?

AleCandido (approved):

PR #198, opened 2023-01-19
split AHq and AHg by Log

giacomomagni:
In order to reuse splitting functions in the approx FONLL we need to split `A_Hq` and `A_Hg` by logs. 
@andreab1997 I believe this it.

### DIFF
--- a/src/eko/matching_conditions/as3/aHg.py
+++ b/src/eko/matching_conditions/as3/aHg.py
@@ -35,6 +35,39 @@ def A_Hg(n, sx, nf, L):
         Incomplete part of the |OME|  :math:`a_{Hg}^{S,(3)}(N)`.
 
     """
+    l0 = a_Hg_l0(n, sx, nf)
+    l1 = a_Hg_l1(n, sx, nf)
+    l2 = a_Hg_l2(n, sx, nf)
+    l3 = a_Hg_l3(n, sx, nf)
+    return l0 + l1 * L + l2 * L**2 + l3 * L**3
+
+
+@nb.njit(cache=True)
+def a_Hg_l0(n, sx, nf):
+    r"""Part of :math:`A_{Hg}^{S,(3)}(N)` proportional to :math:`L^0`.
+
+    The part `a_Hg3(n, sx, nf)` is not fully known, but is parametrized.
+
+    Parameters
+    ----------
+    n : complex
+        Mellin moment
+    sx : list
+        harmonic sums cache
+    nf : int
+        number of active flavor below the threshold
+
+    Returns
+    -------
+    complex
+        :math:`A_{Hg}^{S,(3)}(N)|_{L^0}`
+
+    See Also
+    --------
+    a_Hg3: eko.matching_conditions.as3.aHg_param.a_Hg3
+        Incomplete part of the |OME|  :math:`a_{Hg}^{S,(3)}(N)`.
+
+    """
     S1, _ = sx[0]
     S2, Sm2 = sx[1]
     S3, S21, _, Sm21, _, Sm3 = sx[2]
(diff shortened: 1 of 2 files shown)


andreab1997 (approved):

AleCandido:
> In order to reuse splitting functions in the approx FONLL we need to split `A_Hq` and `A_Hg` by logs.
> @andreab1997 I believe this it.

Implementing FONLL at N3LO this way is prohibitive. In particular, the main problem is not even "FONLL-F" (better called "FONLL-(3,3)") but "FONLL-D" and "FONLL-E" (respectively (3,1) and (3,2)).
@felixhekhorn implemented himself FONLL-B in yadism, and it took a lot of time.

Moreover, in yadism we need the already-convoluted x-space expressions, while in EKO all expressions are supposed to be N-space.

I'm not sure this effort is worth, since it is hard to use anyhow, and it will be superseded by https://github.com/NNPDF/yadism/issues/166

andreab1997:
> > In order to reuse splitting functions in the approx FONLL we need to split `A_Hq` and `A_Hg` by logs.
> > @andreab1997 I believe this it.
> 
> Implementing FONLL at N3LO this way is prohibitive. In particular, the main problem is not even "FONLL-F" (better called "FONLL-(3,3)") but "FONLL-D" and "FONLL-E" (respectively (3,1) and (3,2)). @felixhekhorn implemented himself FONLL-B in yadism, and it took a lot of time.
> 
> Moreover, in yadism we need the already-convoluted x-space expressions, while in EKO all expressions are supposed to be N-space.
> 
> I'm not sure this effort is worth, since it is hard to use anyhow, and it will be superseded by [NNPDF/yadism#166](https://github.com/NNPDF/yadism/issues/166)

Actually we are using another code to compute N3LO FONLL and I believe this is the reason for this PR.

AleCandido:
> Actually we are using another code to compute N3LO FONLL and I believe this is the reason for this PR.

If the other code is able to provide N3LO FONLL, then it has to be able to compute N3LO massive and N3LO asymptotic.

We need those as building blocks, but then we won't use FONLL values directly, because they are not flexible enough to reconstruct all the options currently available, that instead can be constructed from the individual observables.

andreab1997:
> > Actually we are using another code to compute N3LO FONLL and I believe this is the reason for this PR.
> 
> If the other code is able to provide N3LO FONLL, then it has to be able to compute N3LO massive and N3LO asymptotic.
> 
> We need those as building blocks, but then we won't use FONLL values directly, because they are not flexible enough to reconstruct all the options currently available, that instead can be constructed from the individual observables.

The code that we are using is the code of my master thesis where, as you probably know, I have shown that FONLL can be obtained without the asymptotic functions but with other functions that can be defined as combinations of massless and massive functions (which are called `tilde` in the code). So we actually don't need N3LO asymptotics (and we don't have them).

AleCandido:
Moreover, as agreed many times with @felixhekhorn, there should be no mention of FONLL inside EKO itself: EKO is only about evolution, while FONLL is a scheme to treat heavy quark mass effects in observables.

If you need this branch for the time being it is perfectly fine, but please do not merge it in other branches that will be eventually merged in master.

FONLL is being heavily reworked right now by me and @felixhekhorn, if you are implementing something, maybe we should discuss how to do it together.

AleCandido:
> The code that we are using is the code of my master thesis where, as you probably know, I have shown that FONLL can be obtained without the asymptotic functions but with other functions that can be defined as combinations of massless and massive functions (which are called `tilde` in the code). So we actually don't need N3LO asymptotics (and we don't have them).

If you have the massive, you also have the asymptotics: the seconds are the logarithmic part of the first, no more no less.

However, if you prefer to have other functions, we should use them at all orders, not only at N3LO, otherwise we will have an inconsistent code, that is hard to maintain.

andreab1997:
> > The code that we are using is the code of my master thesis where, as you probably know, I have shown that FONLL can be obtained without the asymptotic functions but with other functions that can be defined as combinations of massless and massive functions (which are called `tilde` in the code). So we actually don't need N3LO asymptotics (and we don't have them).
> 
> If you have the massive, you also have the asymptotics: the seconds are the logarithmic part of the first, no more no less.
> 
> However, if you prefer to have other functions, we should use them at all orders, not only at N3LO, otherwise we will have an inconsistent code, that is hard to maintain.

Yes indeed and this is what we have in my code. We always use `tilde` functions

giacomomagni:
Hi @AleCandido thanks for your comments. Let me clarify some points: 

* No one is mentioning FONLL inside the EKO code. It's just written in the description of this PR to explain why I needed this. If you prefer I could have written:  "being able to retain the logs dependency after doing the Mellin inversion of  `A_Hq, A_Hg` ". 
* We are **not** going to implement FONLL in Yadism for the moment. As was said at the last phone call we will produce only k-factors (most likely).
* Most likely the thing we are producing for the N3LO will be just approximate and probably not fully consistent with the exact thing. I hope this will not be the case, but as for the other N3LO ingredients we do what we can. (this comment is totally unrelated to the changes, but refers to the discussion above)

Given this, merging this commit will make @andreab1997 and myself life easier and we will appreciate, otherwise we can survive in any case :)

felixhekhorn (approved):

Issue #181, opened 2022-12-16
Introduce types in the whole package


AleCandido:
At the moment, type hints has been mainly introduced in `eko.io`, and inconsistently here and there in other places.

We should consistently define and use types, as a starting point we should better distribute types that are common to the whole package, lifting a huge part of `eko.io.types` to `eko.types`.

This will be also helpful for #178 

First discussed in https://github.com/NNPDF/eko/pull/172#discussion_r1048512758

AleCandido:
Despite being user-facing, and somehow part of the interface, types are optional in Python. I'll consider them as a non-breaking change, at least for the time being.

felixhekhorn:
Out of the suggested tools here https://mypy.readthedocs.io/en/stable/existing_code.html#automate-annotation-of-legacy-code only [MonkeyType](https://monkeytype.readthedocs.io/en/latest/index.html) seems a realistic option

AleCandido:
Do not use any automated option: types are useful to add information, while, if the information is already there, you could leave it untyped. This is perfectly fine for Python (to have types only somewhere), since `Any` is implicitly added.

I don't remember what's the default (I could quickly look up), but Mypy for sure has an option to (dis)allow `Any`s.

So, let's keep doing it incrementally. If you want to add a bunch of them manually, feel free to do it. But even in that case, please break it in multiple PRs.

PR #199, opened 2023-01-19
Check matching

felixhekhorn:
these are (mostly) some plotting scripts to check matching away from threshold is working in eko - in especially redo fig. 6 from the eko paper for PDFs.

While doing this I did indeed find a bug: https://github.com/NNPDF/eko/commit/d671cc0dfa3ed71b2497e126422d96da674dd043 introduced with #172 in `develop`  (not in v0.10 which I was using for the "quick" plot @AleCandido - I have to redo my current version of the plots). I still have to fix all the tests.


tests/eko/test_couplings.py
```diff
@@ -257,8 +254,7 @@ def benchmark_expanded_n3lo(self):
         m2c = 2
         m2b = 25
         m2t = 30625
-        threshold_list = [m2c, m2b, m2t]
-        threshs = MatchingScales.from_dict(dict(zip("cbt", threshold_list)))
+        threshold_list = [m2c / masses[0], m2b / masses[1], m2t / masses[2]]
```
felixhekhorn, line 257: @giacomomagni I wonder if this was another bug ... note that in the old implementation `threshs` were absolute values (i.e. [GeV^2]) where as we always expect ratios (look at the argument name) - the bug fixed in this PR here is just that the ratio of the arguments should be between squared values

felixhekhorn: Note that I lowered the `rtol` in the test by almost an order of magnitude, though I wonder whether those two things are correlated ...

AleCandido:
@felixhekhorn instead of manually squaring all of those instances, it wouldn't just be easier to add a `squared` properties to `MatchingScales`?

felixhekhorn:
> @felixhekhorn instead of manually squaring all of those instances, wouldn't just be easier to add a `squared` properties to `MatchingScales`?

This had the least impact on the `eko` module (since the linear already got computed). We can reconsider this in #181. I would try to close this asap to not block an eko release.

giacomomagni (dismissed):

tests/eko/test_couplings.py
AleCandido: The run is between 90 GeV and 100 GeV. Simply it was not crossing any threshold before, and it is not doing it even now (moreover, since the ratio was squared, it is running in both cases in 5 flavors).

src/eko/msbar_masses.py
```diff
@@ -350,7 +350,7 @@ def compute(
     couplings: CouplingsRef,
     order: Order,
     evmeth: CouplingEvolutionMethod,
-    matching: MatchingScales,
+    matching,
```
AleCandido, line 353:
```suggestion
    matching: List[float],
```
Just import `List` from the `typing` module.

AleCandido:
I liked more to have the type, but maybe we should solve later on in a more involved way: we can consider to have inheritance from a couple of markers like `EnergyScale` and `SquaredEnergyScale` (unless we can parametrize type with numbers, then we can use `EnergyScale[1]` and `EnergyScale[2]`) in order to make our code safer and more explicit.

I'd really like to clean up consistently the usage of linear and squared scales. But we will do in a separate PR.

src/eko/msbar_masses.py
felixhekhorn: done in https://github.com/NNPDF/eko/pull/199/commits/9e8604a0ee47c701e09068db52e1eb71e2861591

AleCandido (approved):

Issue #143, opened 2022-08-12
Removing deprecated list of list in harmonic cache


giacomomagni:
Explicit call of a function that has a list of list as input doesn't work with `numba`:

````
>>> import eko
>>> sx = eko.harmonics.compute_cache(3, 5,True)
>>> eko.anomalous_dimensions.as4.ggg.gamma_gg_nf0(3,sx)
/project/theorie/gmagni/miniconda3/envs/n3lo_fit/lib/python3.10/site-packages/numba/core/ir_utils.py:2152: NumbaPendingDeprecationWarning: 
Encountered the use of a type that is scheduled for deprecation: type 'reflected list' found for argument 'sx' of function 'gamma_gg_nf0'.

For more information visit https://numba.readthedocs.io/en/stable/reference/deprecation.html#deprecation-of-reflection-for-list-and-set-types

File "../eko/src/eko/anomalous_dimensions/as4/ggg.py", line 202:
@nb.njit(cache=True)
def gamma_gg_nf0(n, sx):
^

  warnings.warn(NumbaPendingDeprecationWarning(msg, loc=loc))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/project/theorie/gmagni/miniconda3/envs/n3lo_fit/lib/python3.10/site-packages/numba/core/dispatcher.py", line 487, in _compile_for_args
    raise e
  File "/project/theorie/gmagni/miniconda3/envs/n3lo_fit/lib/python3.10/site-packages/numba/core/dispatcher.py", line 420, in _compile_for_args
    return_val = self.compile(tuple(argtypes))
  File "/project/theorie/gmagni/miniconda3/envs/n3lo_fit/lib/python3.10/site-packages/numba/core/dispatcher.py", line 965, in compile
    cres = self._compiler.compile(args, return_type)
  File "/project/theorie/gmagni/miniconda3/envs/n3lo_fit/lib/python3.10/site-packages/numba/core/dispatcher.py", line 125, in compile
    status, retval = self._compile_cached(args, return_type)
  File "/project/theorie/gmagni/miniconda3/envs/n3lo_fit/lib/python3.10/site-packages/numba/core/dispatcher.py", line 139, in _compile_cached
    retval = self._compile_core(args, return_type)
  File "/project/theorie/gmagni/miniconda3/envs/n3lo_fit/lib/python3.10/site-packages/numba/core/dispatcher.py", line 152, in _compile_core
[...]

PR #200, opened 2023-01-20
Time-like Anomalous dimensions

t7phy:
Added the LO Anomalous dimensions as starters, hence the PR is currently set to draft

PR #202, opened 2023-01-26
Caching harmonic sums

felixhekhorn:
Here is my proposal for a central harmonic sums cache. Main concepts:
- we drag one cache array around which is not worse than dragging the `sx` array around (and better than the two in #179 ); furthermore this would solve #143 , since we resort to a plain list
- we fill this array slowly whenever we request a new harmonic sum
- and so the moment we compute a new harmonic sum, we register it in the cache
- the central task for everybody that needs a harmonic sum is to register a "key" (in `cache.py`) (Note that the order does not matter, since we would never use the number of the key, but only its name)
- when computing the actual harmonic sum we even could go recursive and have e.g. $g_3$ and $S_{-2,1}$ simultaneously around (we can either go recursive or just do a straight return there, á là `if key == Sm21: return get(g3)+n+zeta`, or we inline the relation where the thing is needed)

Please tell me what you think!

The single commit here is just to lay out the idea at LO and by no means complete ...

cc @enocera 

TODO
- [x] move harmonic constants to global constants


giacomomagni:
Hi @felixhekhorn thanks for this idea, looks good to me! I have two minor comments:
* We should add another argument `is_singlet` or `eta` need for more complex harmonics
* We should define a naming convention for `S(1,N/2), S(1, (N-1)/2) ...` and similar

AleCandido:
@felixhekhorn are you sure that Numba has no issue with globals?

felixhekhorn:
`S1` = $S_1(N)$
`S1h` = $S(1,N/2)$
`S1mh` = $S(1, (N-1)/2)$
`Sm1` = $S(-1,N)$

t7phy:
Hi,
- I have added all the harmonic sums to the cache that are currently implemented in the w files. I have however not yet added the harmonic sums with N/2 and (N-1)/2 as I was not sure up to which order they need to be added. (let me know which ones are needed and I could update them)
- I have also added the harmonic constants to eko.constants module and updated the corresponding imports.
- I also added the bool parameter is_singlet to the cache function to allow consistency with the Smx functions. 
- 
Perhaps have a look and if its alright, we could merge this also and use the new cache in the splitting functions.

niclaurenti:
> Hi,
> 
> * I have added all the harmonic sums to the cache that are currently implemented in the w files. I have however not yet added the harmonic sums with N/2 and (N-1)/2 as I was not sure up to which order they need to be added. (let me know which ones are needed and I could update them)
> * I have also added the harmonic constants to eko.constants module and updated the corresponding imports.
> * I also added the bool parameter is_singlet to the cache function to allow consistency with the Smx functions.
> 
> Perhaps have a look and if its alright, we could merge this also and use the new cache in the splitting functions.

Hi @t7phy ,
we need $S(1,N/2)$, $S(2,N/2)$, $S(3,N/2)$, $S(1,(N-1)/2)$, $S(2,(N-1)/2)$, $S(3,(N-1)/2)$, $g3(N)$ for the anomalous dimensions at `as2` and $S(1,N/2)$, $S(2,N/2)$, $S(3,N/2)$, $S(1,(N+1)/2)$, $S(2,(N+1)/2)$, $S(3,(N+1)/2)$, $g3(N)$,  $g3(N+2)$ for the QED sector.
However, observe that these harmonic sums are not all independent. For example `Sm1" depends on `S1` and `S1h` in the singlet case, while on `S1` and `S1mh` for the non singlet.
So what we should do is to reuse the computation of the harmonics that we have to compute the ones that we need.

src/ekore/harmonics/cache.py
```diff
@@ -0,0 +1,122 @@
...
+    elif key == S2:
+        s = w2.S2(n)
+    elif key == S3:
+        s = w3.S3(n)
+    elif key == S4:
+        s = w4.S4(n)
+    elif key == S5:
+        s = w5.S5(n)
+    elif key == Sm1:
+        s = w1.Sm1(n, get(S1, cache, n), is_singlet)
```
niclaurenti, line 85: For example in this case, since `Sm1` depends on `S1` and one between `S1h` and `S1mh` (depending whether `is_singlet` is `True` or not) I would pass the whole cache, so that the function will check if the harmonic needed is  present, and if not, it will compute it.

t7phy: So if you see in the Sm1 parameters, cache is indeed being using for S1 to use an already computed S1 sum, however the Sm1 function has no input parameter option for the other two. That would require changing the wX files I think.

niclaurenti: Yes, what I'm saying is that I would change the input parameter of `Sm1`(and the same for `Sm2`, and `Sm3`) from `S1` to `cache` and then inside the function `w1.Sm1` I would call
```
if is_singlet:
    return c.get(c.S1h, cache, N, is_singlet) - S1
return c.get(c.S1mh, cache, N, is_singlet) - S1
```

niclaurenti: Obvioulsy you have to call
```
S1 = c.get(c.S1, cache, N, is_singlet)
```
at the beginning of the function

felixhekhorn: It is a matter of design I would say where to compute the "helper s-es", e.g. `S1` for `Sm1` ... but we can decide to delegate it below (as suggested by @niclaurenti ) - especially in view of the arguments of `Sm22`

src/ekore/harmonics/cache.py
```diff
@@ -117,6 +130,30 @@ def get(key: int, cache: npt.ArrayLike,
...
+        s = w3.s3((n-1)/2)
+    elif key == S1ph:
+        s = w1.S1((n+1)/2)
+    elif key == S2ph:
+        s = w2.S2((n+1)/2)
+    elif key == S3ph:
+        s = w3.S3((n+1)/2)
+    elif key == g3:
+        s = mellin_g3(n, get(S1, cache, n))
+    elif key == S1p2:
```
niclaurenti, line 153: $S1(N+2)$ can be obtained from $S1(N)$ with the recursive relations, i.e.
```
S1p2 = polygamma.recursive_harmonic_sum(cache[S1], n, 2, 1)
```
Moreover, once that you have g3p2 you can obtain for free `Sm21` using that
```
g3p2 = -1 / eta * (zeta2 * sm1_np1 - 5.0 / 8 * zeta3 + zeta2 * log2 - sm21_np1)
``` 
with
```
s1_np1 = polygamma.recursive_harmonic_sum(s1, n, 1, 1)
sm1_np1 = sm1 - eta / (n + 1)
sm21_np1 = sm21 - eta / (n + 1) ** 2 * s1_np1
```
and `eta` is the symmetry factor (i.e. 1 for singlet and -1 for non-singlet)

t7phy: I have made all the changes except the Sm21. As far as I can tell, it uses g3p1 (https://github.com/NNPDF/eko/blob/master/src/ekore/harmonics/w3.py#L138), however the comment (https://github.com/NNPDF/eko/blob/master/src/ekore/harmonics/w3.py#L135) is unclear to me. If it is instead using g3p2 (due to N-1 implementation), then perhaps it is inconsistent with (https://github.com/NNPDF/eko/blob/master/src/ekore/anomalous_dimensions/unpolarized/space_like/as1aem1.py#L263).
As far as I can tell, there are no longer any repeated harmonic sum calculations (for the ones that are implemented in the cache)

niclaurenti: Yes I know that it uses $g_3(N+1)$ and not $g_3(N+2)$ but the two are related through recursive relations (similar for the ones relating $S_1(N)$ with $S_1(N+1)$ but slightly more complicated). The equation implemented in https://github.com/NNPDF/eko/blob/master/src/ekore/harmonics/w3.py#L138 should be equivalent to the one in this comment (if my computation is correct).
The point here is that it is better to reuse the value of $g_3(N+2)$, that we already have (actually this is true only if QED is switched on), instead of recomputing $g_3(N+1)$.

src/ekore/harmonics/w1.py
```diff
@@ -3,6 +3,7 @@
 import numba as nb
 import numpy as np
 
+from . import cache as c
```
felixhekhorn, line 6: did you ever run this code? I'm afraid (in contrast to what I said earlier) we can not do this, because this look pretty much like a cyclic import

t7phy: Ah, you are right. I missed this too. So here we will have to let the computation be repeated. I will revert the w file changes.

felixhekhorn: No, if we repeat the computation we loose the purpose of this PR - the solution is to compute all auxillary harmonic sums in `get` _first_ and them pass them down to, e.g. `Sm1`

t7phy: In the latest commit, I have removed the cyclic import while ensuring that the half harmonic sums are only used from the cache and not recomputed.

niclaurenti:
The relations that you need are:
$g_3(N+1)=-S_1(N)/N^2 - g_3(N) +\zeta_2/N$
$g_3(N+2)=S_1(N)/N^2 +g_3(N) - \zeta_2/N +\zeta_2/(N+1)-S_1(N)/(N+1)^2-1/(N+1)^3$

t7phy:
> The relations that you need are: g3(N+1)=−S1(N)/N2−g3(N)+ζ2/N g3(N+2)=S1(N)/N2+g3(N)−ζ2/N+ζ2/(N+1)−S1(N)/(N+1)2−1/(N+1)3

Hey, 
I have added g3p1 and g3p2 to the g_functions and cache files such that they reuse the computed s1 and g3.

niclaurenti:
> > The relations that you need are: g3(N+1)=−S1(N)/N2−g3(N)+ζ2/N g3(N+2)=S1(N)/N2+g3(N)−ζ2/N+ζ2/(N+1)−S1(N)/(N+1)2−1/(N+1)3
> 
> Hey, I have added g3p1 and g3p2 to the g_functions and cache files such that they reuse the computed s1 and g3.

Perfect, now what you have to do is to pass g3p1 to the function that computes Sm21, so that g3p1 is not recomputed inside Sm21

niclaurenti:
@t7phy Moreover now you should pass the cache to all the others anomalous dimensions in the same way that felix did for the as1

felixhekhorn:
let me stress once more what I already said on #200 

> in any case I would suggest you strongly to first do the LO benchmark before going up in order

as you can see [here](https://github.com/NNPDF/eko/actions/runs/4099496319/jobs/7069461334) you already have a significant number of typos ...

enocera:
> let me stress once more what I already said on #200
> 
> > in any case I would suggest you strongly to first do the LO benchmark before going up in order
> 
> as you can see [here](https://github.com/NNPDF/eko/actions/runs/4099496319/jobs/7069461334) you already have a significant number of typos ...

@t7phy I agree with @felixhekhorn , as I also mentioned the other day: let's do the benchmark at LO.

doc/source/overview/features.rst
```diff
@@ -1,7 +1,12 @@
 Features
 ========
 
+<<<<<<< HEAD
```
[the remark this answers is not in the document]
t7phy: I have been using pre commit for the time_like branch but not here because pydocs requires imperative statements in docstrings(for instance, it needs to be Compute xyz instead of Computes xyz), however a lot of the previous functions (such as in as2, as3, etc.) do not implement it so I just skip pre commit for this branch for the time being. Will get to it soon enough ;)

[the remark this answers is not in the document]
felixhekhorn: I know - but for new code you should, of course, do it and then you are only allowed to skip pre-commit in the very last step, i.e. when no one, but pydocstyle is complaining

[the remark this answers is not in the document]
t7phy: right, noted!

t7phy:
@felixhekhorn @niclaurenti @giacomomagni @AleCandido  I have modified the complete space-like implementation to be compatible with the new harmonic cache (ADs and OMEs). Please have a look.

In principal it should be fine to merge in of itself, however as all the tests were based on old cache system, they are failing. I have updated many of them but there is just too many for me to update on my own.

src/eko/evolution_operator/operator_matrix_element.py
```diff
@@ -21,22 +21,25 @@
...
+        a_s : float
+            strong coupling, needed only for the exact inverse
+        backward_method : ["exact", "expanded" or ""]
+            empty or method for inverting the matching condition (exact or expanded)
 
+        Returns
+        -------
+        ome : numpy.ndarray
+            matching operator matrix
+    <<<<<<< HEAD:src/eko/evolution_operator/operator_matrix_element.py
```
giacomomagni, line 39: here there is an unsolved conflict.

src/ekore/anomalous_dimensions/unpolarized/space_like/as2.py
```diff
@@ -10,12 +10,13 @@
...
-from .... import harmonics
-from ....harmonics.constants import log2, zeta2, zeta3
+from eko.constants import log2, zeta2, zeta3
+
+from ....harmonics import cache as c
 
 
 @nb.njit(cache=True)
-def gamma_nsm(n, nf, sx):
+def gamma_nsm(n, nf, cache, is_singlet):
```
giacomomagni, line 19: I think `is_singlet` should not appear as argument in any function except for `nsp` which can be used as singlet and as non singlet. 
What I mean is that calling this function with `is_singlet=True` is wrong, so we should not have this possibility. 
Viceversa hold for `gg, gq, qg, ps ...` 
@niclaurenti is the expert for the mixed QED QCD one.

AleCandido:
What is actually missing from this PR? I.e. why is this still draft?

Please, put checkboxes in the OP for all the items, and tick those that you believe to be completed, even if they can be improved (everything can always be improved).
In general, we should try to merge smaller PR, to reduce the amount of changes and improve the review process.

src/ekore/anomalous_dimensions/unpolarized/space_like/as2.py
niclaurenti: I agree with @giacomomagni . In this case you don't need to pass `is_singlet` because you already know that we are not in the singlet sector. It is better to call `get` with `is_singlet=False` (however notice  that `is_singlet` is useless for the first harmonic sums so consider using a default argument like `is_singlet=None`)

giacomomagni:
Given that #148 has been merged, it's a good time to take care of this.
@t7phy if it's okay for you I can work a bit on this ?

t7phy:
> Given that #148 has been merged, it's a good time to take care of this.
> @t7phy if it's okay for you I can work a bit on this ? 

Yes, please go ahead.

giacomomagni:
okay there are endless confilcts in basically all files... maybe it's easier to start from scratch?

niclaurenti:
> Given that #148 and QED have been merged, it's a good time to take care of this.
> @t7phy if it's okay for you I can work a bit on this ?

If you need help with the harmonic sums needed in the QED sector I can help

giacomomagni:
> If you need help with the harmonic sums needed in the QED sector I can help

I was trying to merge master here again, but everything (almost all the files in `ekore` diplaly conflicts), we have changed
Timelike, QED and N3LO in the meantime + all the refactrs... 
I'd say we can start again in a new branch. The two main ideas were: 

1. We move all the constants in to `eko.constants`.
2. We introduce the harmonic cache as done here.

Was there anything else to save from here @felixhekhorn  @t7phy ?

t7phy:
> > If you need help with the harmonic sums needed in the QED sector I can help
> 
> I was trying to merge master here again, but everything (almost all the files in `ekore` display conflicts), we have changed
> Time like, QED and N3LO in the meantime + all the refactors... 
> I'd say we can start again in a new branch. The two main ideas were: 
> 
> 1. We move all the constants in to `eko.constants`.
> 2. We introduce the harmonic cache as done here.
> 
> Was there anything else to save from here @felixhekhorn  @t7phy ? 

Yes I agree, I think starting in new branch is simpler. The two files are the only main changes to copy. Ofcourse, you could also copy all the space like ADs and OMEs as I had already added the new harmonic sums to them.

felixhekhorn:
> I was trying to merge master here again, but everything (almost all the files in `ekore` display conflicts), we have changed Time like, QED and N3LO in the meantime + all the refactors... I'd say we can start again in a new branch. The two main ideas were:
> 
>     1. We move all the constants in to `eko.constants`.
> 
>     2. We introduce the harmonic cache as done here.
> 
> 
> Was there anything else to save from here @felixhekhorn @t7phy ?

I agree, a clean restart might be better ... (solving the `is_singlet` along the way)

giacomomagni:
closed in favor of #252 and new PR

PR #207, opened 2023-01-31
Implement QED truncated solution

niclaurenti:
TODO:
- [x] Implement `"iterate-expanded"` solution for non-singlet
- [x] Implement `"truncated"` solution for non-singlet
- [ ] Implement `"ordered-truncated"` solution for non-singlet
- [ ] Compute and implement a truncated solution for 4x4 and 2x2 sectors


src/eko/kernels/singlet.py
```diff
@@ -359,7 +359,7 @@ def eko_iterate(gamma_singlet, a1, a0, beta_vec, order, ev_op_iterations):
 
 
 @nb.njit(cache=True)
-def r_vec(gamma_singlet, beta, ev_op_max_order, order, is_exact):
+def r_vec(gamma_singlet, beta, ev_op_max_order, order, is_exact, dim=2):
```
AleCandido, line 362: Not sure that Numba appreciates default for arguments:
https://numba.pydata.org/numba-doc/dev/reference/types.html?highlight=optional#optional-types

Just drop the default, and plug it where it was originally called.

src/eko/kernels/singlet.py
```diff
@@ -454,7 +454,7 @@ def u_vec(r, ev_op_max_order):
 
 
 @nb.njit(cache=True)
-def sum_u(uvec, a):
+def sum_u(uvec, a, dim=2):
```
AleCandido, line 457: Same of above

src/eko/kernels/valence_qed.py
```diff
@@ -101,5 +52,37 @@ def dispatcher(
             singlet EKO
     """
     if method in ["iterate-exact", "iterate-expanded"]:
-        return eko_iterate(gamma_valence, a1, a0, aem_list, nf, order, ev_op_iterations)
-    raise NotImplementedError('Only "iterate-exact" is implemented with QED')
+        return eko_iterate(
+            gamma_valence, a1, a0, aem_list, nf, order, ev_op_iterations, dim=2
+        )
+    if method == "perturbative-exact":
+        return eko_perturbative(
```
AleCandido, line 59: Why is this perturbative, if exact is selected? Moreover: is the total valence sector any different from the singlet one in QED? Apart for the dimension (that could be passed by argument)

AleCandido (changes requested):
Still need to review `singlet_qed.py`

src/eko/kernels/valence_qed.py
niclaurenti: If you see singlet.py the methods `perturbative-exact` and `perturbative-expanded` both use `eko_perturbative` but one with `is_exact=True` and the other with `False`.

AleCandido: Sorry, my bad :)

felixhekhorn:
This is not possible by analytical means (see Appendix of [NNPDF4.0QED](https://inspirehep.net/literature/2747770)), and doing it numerically is far more complicated

PR #210, opened 2023-02-09
bye bye harmonics weight 5

giacomomagni:
To facilitate #202  and delete unused stuff, we move the harmonics weight 5 in an independent package provided in extras.
Note that only `S5` and `Sm5` will remain in eko (since they are actually used).


t7phy:
> To facilitate #202 and delete unused stuff, we move the harmonics weight 5 in an independent package provided in extras. Note that only `S5` and `Sm5` will remain in eko (since they are actually used).

But what is the purpose of separating it? This would break the cache in #202

giacomomagni:
> But what is the purpose of separating it? This would break the cache in #202

These functions are never used inside any expression implemented in `ekore` so far (I hope you won't need them as well. Note that `S5` and `Sm5` will remain in `ekore`), 
so they are actually a leftover of an old implementation of some `N3LO` matching condition which is now given parametrised.  Keeping them in `ekore` is just slowing dow the compilation and maintenance.

extras/harmonics_w5/pyproject.toml
```diff
@@ -0,0 +1,102 @@
...
+[tool.poetry.group.dev]
+optional = true
+
+[tool.poetry.group.dev.dependencies]
+pdbpp = "^0.10.3"
+ipython = "^8.0"
+# benchmark
+asv = "^0.4.2"
+virtualenv = "^20.13.2"
+devtools = "^0.10.0"
```
AleCandido, line 50: I appreciate the idea of keeping a package, and even tests are meaningful (you wrote them, that's useful work, no need to discard). But maybe these are too much...

extras/harmonics_w5/pyproject.toml
```diff
@@ -0,0 +1,102 @@
...
+output-format = "colorized"
+[tool.pylint.format]
+# Maximum number of characters on a single line.
+max-line-length = 100
+[tool.pylint.design]
+# Maximum number of arguments for function / method
+max-args = 10
+
+[tool.pydocstyle]
+convention = "numpy"
```
AleCandido, line 102: I wonder if also this part is a bit excessive...
In the end, we are not going to run any CI, nor frequently run PyLint here.

If anything, I would keep just the `test` and `lint` tasks (it is useful not to have to remember how to run tests/check the code, especially when you don't work on it frequently).

AleCandido (approved):
Definitely fine. I added a couple of suggested simplifications to the package, but you can even skip and merge. As you wish.

t7phy (approved):

Issue #205, opened 2023-01-27
Read-only runcards objects


AleCandido:
Runcards objects should mirror exactly the user input.

E.g. `Rotations` in `OperatorCards` contains a lot of members that should never be set by the user, and are only modified later on. This should definitely fit somewhere else.

Let's review this objects as soon as possible. They should also not be loaded from dictionaries containing leading `_` keys.

@andreab1997 provide your feedback about the new runcards.

PR #217, opened 2023-02-23
Create recipes

AleCandido:
- [ ] create recipe files in a new EKO
    - *new EKO*: empty `recipes/`, `parts/`, and `operators/` folders
- [ ] double check the information in the recipe to be sufficient for computation

### DIFF
--- a/src/eko/couplings.py
+++ b/src/eko/couplings.py
@@ -9,7 +9,7 @@ See :doc:`pQCD ingredients </theory/pQCD>`.
 """
 import logging
 import warnings
-from typing import List
+from typing import Iterable, List
 
 import numba as nb
 import numpy as np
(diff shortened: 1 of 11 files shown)


AleCandido:
In commit 85975ad8f156df21edc81d26fce29e69449cf202 a further breaking change is introduced.

There is no hurry to merge this PR, for as long as there is not the full alternative runner. In any case, even the full alternative runner will deprecate, but not immediately drop, the old one, so this will still be compatible.

Since I had to play with these quantities to use them in the new runner as well, I decided to start improving (or just factorizing) a bit the content of the current one.
In light of #205 I tried to improve the existing classes.

However, if the changes are welcome, I can keep going fixing the tests/benchmarks/docs (in principle should simplify the current definitions, also clarifying the role of linear and squared scales).
Otherwise, it is sufficient to drop that commit. @felixhekhorn let me know.

AleCandido:
This is now rebased on top of #220, and I will restart work on this one as soon as #220 will be close to final.

PR #220, opened 2023-02-28
Reorganize runcards and metadata

AleCandido:
https://github.com/NNPDF/eko/pull/217#issuecomment-1445440835

Given that I had to rediscover the hot water in #217, I decided that I really needed to solve first #205, i.e. to stabilize the runcards layout.

I knew already from @andreab1997 feedback that there are some suboptimal components (e.g. presence of `Rotations` in `OperatorCard`, or the doubling of `_mugrid`/`_mu2grid`, none of which should be dumped with a prepended underscore), and I know by myself that some other objects are cumbersome (e.g. the quarks props require a dict with keys `"cbt"`, or the `value`/`scale` quantities), or potentially wrongly located (I moved `EvolutionMethod` to the `OperatorCard`, along the other properties specifying it, but I never asked everyone else).
Since we have to start soon computing expensive EKOs, it is better to solve this immediately.

So, please, @andreab1997, @giacomomagni, and @felixhekhorn provide feedback on the current status of runcards, such that I can make a proposal for the final layout (of course also @niclaurenti and @t7phy are invited to comment, but I expect the played much less with these runcards).

- [x] to replace some representations, hiding the underscored objects (potentially assigning default values, and adding some post validations) and
- [x] to drop the Rotations from the OperatorCard
- [x] update `mu2grid` to host tuples with flavor
- [x] implement `flavored_mu2grid()`
- ~~check/fix actual benchmarks~~
  - [x] check, they are broken
  - but already broken in `master`, they will be fixed there
- [x] check `Generic` based types are working, drop superseded *dynamic generation* explanation https://github.com/NNPDF/eko/pull/220#discussion_r1133769831


src/eko/io/types.py
```diff
@@ -33,29 +33,56 @@
 # Scale functions
 # ---------------
 
+T = TypeVar("T")
```
felixhekhorn, line 36: This is duplicating line 10

src/eko/quantities/heavy_quarks.py
```diff
@@ -1,62 +1,73 @@
...
-            """Allow access by index.
+    @c.setter
+    def c(self, value: T):
+        self[0] = value
 
-            Consequently it allows iteration and containing check.
+    @property
+    def b(self) -> T:
+        """Bottom quark."""
+        return self[0]
```
felixhekhorn, line 35:
```suggestion
        return self[1]
```

src/eko/io/types.py
AleCandido: Thanks, I'll drop :D

src/eko/quantities/heavy_quarks.py
AleCandido: Too much copy-pasting...

AleCandido:
@felixhekhorn as usual, tests are broken in surprisingly complicate ways, so I'm a bit stuck fixing them...
(in particular, as usual, the most complicated ones are `msbar_masses`, it is kind of a curse...)

AleCandido:
Now tests are passing thanks @felixhekhorn (after 3 trials of running CI, thanks @codecov...).

felixhekhorn:
> after 3 trials of running CI

If I understood correctly the blog post you linked the other day there is nothing we can do (except move away from CodeCov) and trying to re-run makes the situation only worse ... so if we fail, we can check whether we fail in pytest or on codecov and simply ignore the latter (and hope for the next commit)

AleCandido:
> If I understood correctly the blog post you linked the other day there is nothing we can do (except move away from CodeCov) and trying to re-run makes the situation only worse ... so if we fail, we can check whether we fail in pytest or on codecov and simply ignore the latter (and hope for the next commit)

I tried to update the action version, but as you said in principle nothing should change...

src/eko/io/types.py
```diff
@@ -1,142 +1,87 @@
...
+class ReferenceRunning(list, Generic[T]):
+    """Running quantities reference point.
 
-def heavy_quark(quarkattr):
-    """Generate heavy quark properties container classes.
+    To simplify serialization, the class is just storing the content as a list,
+    but:
 
-    The motivation for dynamic generation is provided in module docstring.
+    - it is constructed with a ``Running(T, Scale)`` signature
```
felixhekhorn, line 43: The default constructor is `ReferenceRunning([value, scale])` (see msbar tests)

felixhekhorn: else, of course, you can use `typed`

AleCandido:
In c5f0d19 I actually dropped more than I wanted to do in a single commit, but it was worth to fix tests once and for all.
So:
- `pids` are now stuck at the flavor basis, since is the only one supported for computation
- `_mugrid` and `_mu2grid` are not any longer present, maintaining them was more expensive than beneficial; from now on, all scales outside EKO should be linear
- `Rotations` has been finally removed from the `OperatorCard`, while still present in the `Metadata`

There is another breaking change I'd like to operate in this PR (other than the new `mugrid` with flavor numbers): at the moment, the operator works internally with squared scales.
This is perfectly fine, since we know inside EKO all scales should be squared.
But there is just one weird point: accessors of an `EKO` object are also using squared scales, so the user should type `eko[mu2]`, when he requested a `mugrid`.

While this is going in the correct direction, i.e. you decide a scale, and then you square it (that is more predictable than a square root), it's still unpleasant...
Would you mind if I start converting also them for the end user? Internally, it will still use squared scales, and I can add some "squared methods" for the internal usage (but of course there is only one `__getitem__`, since type info does not survive on values).

@felixhekhorn @giacomomagni @niclaurenti @andreab1997

AleCandido:
Just to point out once more: our unit tests are not unit at all... and for this reason some specific changes are a real pain...

AleCandido:
@felixhekhorn you'll be happy: upgrading `mu2grid` had an effect (often indirect) in many other places, so a bunch of files had to be changed.

On top of that, since I was already looking at most of the occurrences, I decided to drop `Q2grid` everywhere (compatibly with legacy mode).
The two things together resulted in a lot of files to be changed. But most of the changes should be trivial, so, despite the number, it should be rather simple to review, even though not the easiest task of course (but we knew from the beginning that this was going to be a breaking PR...).

AleCandido:
There are only two new items that have been generated by 6b830a910110bbde3756c9dc8317e5eb99ff2c75:

- check/fix actual benchmarks: I'd like to use `mugrid`, consistently with the policy of having linear scales outside EKO, but this might have been done inconsistently, and we do not have benchmarks running in the CI, so the green tick might be misleading
- implement `flavored_mu2grid()`: this is almost a requirement for the former one, since in most cases we do not want to specify `mu_nf` in benchmarks, since there is no choice for LHA, APFEL, PEGASUS, ... but from now on the choice is available for EKO! Thus, we need to split the number of flavors computation outside the runner\*, and make it available for others

\* actually, the runner will do nothing less because of this, but only "more", since the paths computed by the atlas should take into account the final number of flavors, instead of inferring it

src/eko/io/runcards.py
```diff
@@ -376,3 +415,24 @@ def update(theory: Union[RawCard, TheoryCard], operator: Union[RawCard, Operator
...
+    There is no one-to-one relation between number of running flavors and final
+    scales, unless matchings are all applied. But this is a custom choice,
+    since it is possible to have PDFs in different |FNS| at the same scales.
+
+    .. todo::
+
+        To be actually implemented
+
+    """
+    return [(np.sqrt(mu2), 0) for mu2 in mu2grid]
```
felixhekhorn, line 438:
```suggestion
    return [(np.sqrt(mu2), nan) for mu2 in mu2grid]
```
as for other quantities (e.g. `nfref`) let's use `nan` to autodetect

src/eko/io/runcards.py
```diff
@@ -5,30 +5,37 @@
...
 import numpy as np
 import numpy.typing as npt
 
 from .. import basis_rotation as br
-from .. import interpolation
+from .. import interpolation, msbar_masses
 from .. import version as vmod
+from ..couplings import couplings_mod_ev
+from ..quantities.heavy_quarks import (
+    FLAVORS,
```
felixhekhorn, line 20: I think we should scope here: `from bla import FLAVORS as HFLAVORS` (or even more explicit if you wish) (like wise everywhere else)

src/eko/io/legacy.py
```diff
@@ -98,27 +101,29 @@ class PseudoOperator(DictLike):
...
+        mu2list = old.get("Q2grid")
+        if mu2list is None:
+            mu2list = old["mu2grid"]
+        mu2grid = np.array(mu2list)
 
         xgrid = XGrid(old["interpolation_xgrid"])
-        pids = old.get("pids", np.array(br.flavor_basis_pids))
 
-        rotations = Rotations(xgrid=xgrid, pids=pids)
+        rotations = Rotations(xgrid=xgrid)
```
felixhekhorn, line 118: drop if no longer used (which I think is and will be the case)

src/eko/io/runcards.py
```diff
@@ -82,6 +111,10 @@ class Configs(DictLike):
     """
     ev_op_iterations: int
     """Number of intervals in which to break the global path."""
+    scvar_method: Optional[ScaleVariationsMethod]
+    """"""
```
felixhekhorn, line 115:
```suggestion
    """Scale variation method"""
```

src/eko/io/runcards.py
```diff
@@ -176,32 +208,49 @@ def targetgrid(self) -> interpolation.XGrid:
...
+        return d
+
 
 @dataclass
 class OperatorCard(DictLike):
     """Operator Card info."""
 
     mu0: float
     """Initial scale."""
+    mugrid: List[Target]
```
felixhekhorn, line 245:
```suggestion
    mugrid: List[Target]
    """Array of final scales."""
```

src/eko/io/runcards.py
```diff
@@ -176,32 +208,49 @@ def targetgrid(self) -> interpolation.XGrid:
...
+    xgrid: interpolation.XGrid
+    """Momentum fraction internal grid."""
```
felixhekhorn, line 247:
```suggestion
    """Internal momentum fraction grid."""
```

src/eko/quantities/__init__.py
```diff
@@ -0,0 +1,33 @@
...
+subclassing ``DictLike``, indeed::
+
+    inspect.isclass(RunningReference)       # False
+    issubclass(RunningReference, DictLike)  # raise an error, since
+                                            # RunningReference is not a class
+
+Essentially classes can be used for type hints, but types are not all classes,
+especially when they involve generics.
+
+For this reason I prefer the less elegant dynamic generation, that seems to
```
felixhekhorn, line 30: is this true for this subpackage?

src/eko/io/runcards.py
AleCandido: It is just a placeholder, there is written right above `To be actually implemented`, and this is the first item missing in the OP

felixhekhorn (commented):
a few comments

src/eko/io/runcards.py
AleCandido: In this case, if you prefer, I can just `from ..quantities import heavy_quarks as hq`.

src/eko/io/runcards.py
AleCandido: This level of documentation is coming together with the type...

src/eko/quantities/__init__.py
AleCandido: It was, it has been replaced by the `Generic` style, so I dropped dynamic generation.

However, I was aware that this comment is around (but good that you reminded me, because in the meanwhile I forgot), but I postponed the removal after confirmation that the new `Generic` based types were working as expected, with no further surprise.

src/eko/io/runcards.py
felixhekhorn: Fine, but `mu` is not really telling, so let's just add that single sentence ...

AleCandido: I'm not explicitly against that sentence, and for sure in favor of more telling names. If you have anything better than `mugrid` I'd be glad to adopt, even though I believe `mugrid` to be actually meaningful in this case.

My main point is that the content of your sentence is not much in the name, but rather in the type:
- *Array of* $\iff$ `List[]`
- *final scales* $\iff$ `Target`

Moreover, *Array* is wrong, since it is not an array, and *final scales* is wrong as well, since they are supposed to be pairs `(scale, nflavs)`, while the information in the type is much more accurate.

They are not especially wrong, they are only approximate. And since they are less accurate then type itself, nor adding more information, I don't see much the point.

src/eko/io/runcards.py
AleCandido: Implemented in 7673b78

src/eko/io/runcards.py
AleCandido: Done in d0451659042bb7729173526573625997fcdd007b

src/eko/io/legacy.py
AleCandido: It was already changed in c5f0d1972c669797eae2d7dfb585d981b10e22d3
https://github.com/NNPDF/eko/commit/c5f0d1972c669797eae2d7dfb585d981b10e22d3#diff-3cb1f56e5e97e4f210cfa50e04bf1a2c30e49aeceb02c0f1e368738d4f4c4546L113-R112

Since the commit is from 10th of March, and your comment from 13th, I believe you just reviewed an old version...

AleCandido: Done in 39bc095c4c0720bd5c3f3c9ad6a5cfac3254e40c

src/eko/io/runcards.py
AleCandido: Done in 3a380883cfe68afb46ba1dcf25baf5611860e129

src/eko/io/runcards.py
AleCandido: Done in 3a380883cfe68afb46ba1dcf25baf5611860e129

src/eko/quantities/__init__.py
AleCandido: Done in 747956376a6503747da8edcd6a37da5265228b06 and 9fc14bafe7ee0e48ac2a043b2fd6926cddb6f018

AleCandido:
@felixhekhorn the PR should be ready, but for testing (and possibly fixing) actual benchmarks.

Your review, and potentially also your help (to speed up), would be appreciated :bow:

AleCandido:
Actually, I'm not sure about the representation of the `CouplingsRef` as a running object:
https://github.com/NNPDF/eko/blob/9fc14bafe7ee0e48ac2a043b2fd6926cddb6f018/src/eko/io/types.py#L138-L146

There are possibly two running objects, but they have to be coordinated: either the second is not running, or they have to share the reference scale, enforced in `__post_init__`:
https://github.com/NNPDF/eko/blob/9fc14bafe7ee0e48ac2a043b2fd6926cddb6f018/src/eko/io/types.py#L153-L161

But instead for `num_flavs_ref` we (i.e. I) chose a different strategy: if there is, the number of flavors at reference scale is a unique one.
I wonder if it is better to have separate reference values for `alphas/em`, with a further single scale (always present since `alphas` is always running), and then a further field to mark if `alphaem` is running or not.

This option would denest the structure, save a check (since it will be only possible to represent consistent objects), improve symmetry with `num_flavs_ref`, and we can still restore the running attributes as constructed properties.

What do you think about? @felixhekhorn @giacomomagni @niclaurenti @andreab1997

src/eko/io/dictlike.py
```diff
@@ -80,8 +78,18 @@ def from_dict(cls, dictionary):
         # finally construct the class, just passing the arguments by name
         return cls(**dictionary)
 
-    @property
-    def raw(self):
+    @classmethod
+    def from_dict(cls, dictionary):
```
giacomomagni, line 82: why do you want to  duplicate this?

niclaurenti:
> Actually, I'm not sure about the representation of the `CouplingsRef` as a running object:
> 
> https://github.com/NNPDF/eko/blob/9fc14bafe7ee0e48ac2a043b2fd6926cddb6f018/src/eko/io/types.py#L138-L146
> 
> There are possibly two running objects, but they have to be coordinated: either the second is not running, or they have to share the reference scale, enforced in `__post_init__`:
> 
> https://github.com/NNPDF/eko/blob/9fc14bafe7ee0e48ac2a043b2fd6926cddb6f018/src/eko/io/types.py#L153-L161
> 
> But instead for `num_flavs_ref` we (i.e. I) chose a different strategy: if there is, the number of flavors at reference scale is a unique one. I wonder if it is better to have separate reference values for `alphas/em`, with a further single scale (always present since `alphas` is always running), and then a further field to mark if `alphaem` is running or not.
> 
> This option would denest the structure, save a check (since it will be only possible to represent consistent objects), improve symmetry with `num_flavs_ref`, and we can still restore the running attributes as constructed properties.
> 
> What do you think about? @felixhekhorn @giacomomagni @niclaurenti @andreab1997

You mean that you would like to remove `QrefQED` and restore the flag `alphaem_running`?
For me it's fine, but in this case we should drop the idea of implementing the decoupled running as I proposed few weeks ago.
In ant case, I don't think we really need the decoupled running given that I wanted to implement it only for benchmarking against `APFEL`.

src/ekomark/data/operators.py
```diff
@@ -17,7 +17,7 @@
             n_integration_cores=0,
             debug_skip_non_singlet=False,
             debug_skip_singlet=False,
-            Q2grid=[100],
+            mugrid=[100],
```
giacomomagni, line 20:
```suggestion
            mugrid=[10],
```

src/eko/runner/compute.py
```diff
@@ -0,0 +1,8 @@
+"""Compute parts from recipes."""
+from .parts import Part
+from .recipes import Recipe
+
+
+def compute(recipe: Recipe) -> Part:
+    """Compute EKO component in isolation."""
+    return Part("ciao")
```
giacomomagni, line 8: Don't we need a better name here?!

AleCandido:
> You mean that you would like to remove `QrefQED` and restore the flag `alphaem_running`?
> For me it's fine, but in this case we should drop the idea of implementing the decoupled running as I proposed few weeks ago.
> In ant case, I don't think we really need the decoupled running given that I wanted to implement it only for benchmarking against `APFEL`.

Ok, good that you mentioned: what is the decoupled running?

Everything that is currently supported will still be supported in the proposed layout. That class is only used to specify the reference value, and we always said that if both are running, the two reference values have to be given at the same scale (since it's the border condition of a vector-valued differential equation).

If you then want to have more of those for some reason, you should be free to instantiate multiple ones per run.
But yes, the proposal is exactly what you said above.

giacomomagni (commented):
For my knowledge are these changes compatible with the current Eko on master??

niclaurenti:
> Ok, good that you mentioned: what is the decoupled running?
It's the running without the mixed terms (i.e. the QED corrections to the QCD beta function and the QCD corrections to the QED beta function). So in this case the two `Qref` can be different.
> If you then want to have more of those for some reason, you should be free to instantiate multiple ones per run. But yes, the proposal is exactly what you said above (i.e. _remove `QrefQED` and restore the flag `alphaem_running`_).
In any case, I'm perfectly fine with your proposal.

src/eko/runner/compute.py
AleCandido: This part has to be implemented in #217 and further PRs. I actually started from there, and then I remembered I had to first fix the runcard.

Ignore every new file in `src/eko/runner/` for the time being. As soon as we close this PR, we will move the focus on them.

AleCandido:
> For my knowledge are these changes compatible with the current Eko on master?? In the sense that ekos can be loaded and used safely, provided `pineko` will be adjusted...

@giacomomagni The point of this PR is precisely to break the runcard once more, since we know that there are leftovers of the December rework that are meaningless (`rotations` in the `OperatorCard`) or undesirable (`_mu[2]grid`).
Since we're going to break the runcards, I asked you to provide as many improvements as possible, but nothing came, so I just kept going with what I knew that had to be fixed.

> So, please, @andreab1997, @giacomomagni, and @felixhekhorn provide feedback on the current status of runcards, such that I can make a proposal for the final layout (of course also @niclaurenti and @t7phy are invited to comment, but I expect the played much less with these runcards).

However, as mentioned this should break the *producer* interface, in the sense that existing runcards might have to be manually/automatically upgraded, but it should not break the *consumer* interface. I.e. if you computed an EKO, you should still be able to read it (inside `eko.io.struct` the `mu2grid` terminology was already used).

AleCandido:
> It's the running without the mixed terms (i.e. the QED corrections to the QCD beta function and the QCD corrections to the QED beta function). So in this case the two `Qref` can be different.

If they are decoupled (that is possibly inconsistent with usage in DGLAP) you will be able to use two different `CouplingsRef` instances: one for the `alphas` running, and the other for `alphaem` running. You can set the reference value of the unused coupling to `nan`, and just do nothing for that.

If they are decoupled, but the reference values are given at the same scale, you can also share the same instance, and just shut down the mixed terms :)

src/ekomark/data/operators.py
AleCandido: This makes sense, it is a consistent update :)

I'll add it.

giacomomagni:
> @giacomomagni The point of this PR is precisely to break the runcard once more, since we know that there are leftovers of the December rework that are meaningless (`rotations` in the `OperatorCard`) or undesirable (`_mu[2]grid`). Since we're going to break the runcards, I asked you to provide as many improvements as possible, but nothing came, so I just kept going with what I knew that had to be fixed.
>

okay, I'm sorry about not having provided any feedback. Main reason is because I am honestly still quite confused to about all these refactors (I am not skilled enough to understand everything at the first sight, and time is finite. My bad of course, not your.)

> However, as mentioned this should break the _producer_ interface, in the sense that existing runcards might have to be manually/automatically upgraded, but it should not break the _consumer_ interface. I.e. if you computed an EKO, you should still be able to read it (inside `eko.io.struct` the `mu2grid` terminology was already used).

I'm asking these question of compatibility, from the point of view of the user, just to plan and be aware of which tools have then to be updated and if it will be possible to reuse existing material (for instance JR is still asking for ekos a la IC paper style... ). If the second it's not possible fine, I understand, provided that what will be given now is as good as or even better of you had before.  

> EDIT: the part about the _consumer_ interface will be true up to the point in which we will actually use the specified number flavors in `mugrid` to implement the evolution part of "numerical FONLL": since now you access an operator with a scale, and after you will need a scale and a number of flavors, this will break compatibility (but it is a genuinely new feature: it is really difficult to add it without breaking anything...)

okay fine. I hope we can address this on a separate PR once we start implementing numerical FONLL.

src/eko/runner/compute.py
giacomomagni: okay could you plese then add some comments or TODOs otherwise it's difficult to understand. 
If I understand correctly #217 now contains a file changed which can be already included here, is this the case?

AleCandido:
> okay, I'm sorry about not having provided any feedback. Main reason is because I am honestly still quite confused to about all these refactors (I am not skilled enough to understand everything at the first sight, and time is finite. My bad of course, not your.)

Not really: I know it's confusing, so at some point we can really have an "EKO meeting" to put everyone on the same page.
I currently proposed to do it during/in place of a "code meeting" (since usual code meetings are actually `NNPDF/nnpdf` repo meetings). If there will be one in 10 days, we can discuss during that, otherwise we might organize another one separately.

AleCandido:
> I'm asking these question of compatibility, from the point of view of the user, just to plan and be aware of which tools have then to be updated and if it will be possible to reuse existing material (for instance JR is still asking for ekos a la IC paper style... ). If the second it's not possible fine, I understand, provided that what will be given now is as good as or even better of you had before.

I understand, sorry if we lacked explanation before :bow:

> okay fine. Is this fully included already in this PR?

The runcard upgrade is intended to be fully included in this PR. But of course this is prone to having made good choices, if you then believe that they were not and are causing problems, we might change again (first I want it to be used, then I can promise it will be stable, even if of course I'd like to be stable asap).
Any other change in this PR should not affect the user and just be internal.

The largest past of the UI of `eko` package is now in `eko.io`, even though that's not yet the *Public Interface*.
However, you mostly expect a user to write runcards (and they're being updated here) and interact with `eko.EKO` objects (full path `eko.io.struct.EKO`), and they are mostly not updated (only internal ones).

The only update to `eko.EKO` is in its `raw` representation https://github.com/NNPDF/eko/pull/220/files#diff-b42c25cd64ec550e3bf712384a042ff99c0a6fe17cde2ce09fd56c5df6b9adae. But `eko.EKO` is mostly used with its underlying folder, so with the `EKO.open` constructors (and aliases, i.e. `EKO.create`, `EKO.read`, and `EKO.edit`), so in principle users using this suggested way should be unaffected (so I expect all EKOs to be the exact same after this PR, and usable exactly as before).

src/eko/runner/compute.py
AleCandido: You're right #217 is just an empty PR at the moment, so do not consider its content, I will reset the branch to this one as soon as we'll be ready to merge.

That is just the next one I'm going to work on, simply it happened to be historically the first one I opened (you can infer from the number) because of bad planning (i.e. I forgot I had to do this first).

AleCandido:
Sandbox and benchmarks are currently broken, because my deserialization is not good enough with `Generic` types...

AleCandido:
Ok, cf5b666 was completely unexpected, and it's not the end of the story.
I.e. actual benchmarks are actually broken, but I'm quite convinced it is not this PR's fault (I haven't tried yet on `master`, just because I didn't want to recompile, but I'll do).

- the runner was using `EKO.read()`, even though potentially manipulating it
- the theory card was using the intermediate format (`"order"` key), but in benchmarks we're actually creating them in the legacy format (so `"PTO"` and `"QED"` keys), and upgrading them
- plots were completely broken (and they still are, just a bit less), since they were using the old EKO format
- I'm now getting a clear QED error for `poe lha`, but this branch should be based on current `master`
	```
	AssertionError: 49 columns passed, passed data had 50 columns
    [...]
	ValueError: 49 columns passed, passed data had 50 columns
	```

Mea culpa: I haven't run benchmarks in a while.
But is someone running them? @felixhekhorn @niclaurenti @giacomomagni

AleCandido:
> Ok, [cf5b666](https://github.com/NNPDF/eko/commit/cf5b6666e6e70fed23bc544cebcad546dcbda24f) was completely unexpected, and it's not the end of the story.

This commit has been moved to `master`, and became a8ee0a8c90462bfa83a56bcaf3855b7fae461b38

giacomomagni:
> > Ok, [cf5b666](https://github.com/NNPDF/eko/commit/cf5b6666e6e70fed23bc544cebcad546dcbda24f) was completely unexpected, and it's not the end of the story.
> 
> This commit has been moved to `master`, and became [a8ee0a8](https://github.com/NNPDF/eko/commit/a8ee0a8c90462bfa83a56bcaf3855b7fae461b38)

Yes theys seems `lha_paper_bench` seems broken at LO also on the `N3LO_ad_variations` branch, which was rebased after the QED merge.
I thought @niclaurenti checked that benchmarks were fine before merging #135 ...

We definitely need a bot that run `poe lha` before merging anything at this point!! 
Is this only and ekomark problem or there is something more fishy going on?

niclaurenti:
> > > Ok, [cf5b666](https://github.com/NNPDF/eko/commit/cf5b6666e6e70fed23bc544cebcad546dcbda24f) was completely unexpected, and it's not the end of the story.
> > 
> > 
> > This commit has been moved to `master`, and became [a8ee0a8](https://github.com/NNPDF/eko/commit/a8ee0a8c90462bfa83a56bcaf3855b7fae461b38)
> 
> Yes `lha_paper_bench` seems broken at LO (without SV) also on the `N3LO_ad_variations` branch, which was rebased after the QED merge. I thought @niclaurenti checked that benchmarks were fine before merging #135 ...
> 
> We definitely need a bot that run `poe lha` before merging anything at this point!! Are benchmarks now passing on `master` ? Was it only an `ekomark` problem or there is something more fishy going on?

Did you run
```
obj = BenchmarkFFNS()
obj.benchmark_plain(0)
```
?

AleCandido:
> We definitely need a bot that run `poe lha` before merging anything at this point!!

Let me finish the "jets rework" and then I will implement also that :)

AleCandido:
> Are benchmarks now passing on `master` ?
> Was it only an `ekomark` problem or there is something more fishy going on?

Most of the benchmarks should be now passing on `master`, even though you should try. @felixhekhorn started yesterday from cf5b666 and tried to fix the rest.

We're still fixing the operator plots, that almost no one is using, but they are used by sandbox. While doing this, we found another item to be discussed, but if there is pressure to fix `sandbox.py` as well, I believe we should be able to do it. Correct @felixhekhorn?

giacomomagni:
> Let me finish the "jets rework" and then I will implement also that :)

I can give a try and implement a minimal thing.

AleCandido:
> I thought we can have something even more easier to start. Like the fit-bot in nnpdf, you add the label and it runs the lha benchmark and then you look at log file by eyes. I know it will be rudimental, but at least it will save us from these bugs, which are quite dangerous. I can make a PR, this solution will involve just adding a workflow (for the moment we can do and eko only) and adding `argparse` or similar to `lha_paper_bench.py` so it will be easier to select the bench you want to run.

I'm not worried about an optimal solution, what essentially does (or should do) a job in that workflow is:
```sh
git clone ... # through actions/checkout
poetry install -E mark --with tests
poe lha
```
and that's it.

The only further bits for which I pointed it to you are:
1. the container, that is mandatory, otherwise you would have to install every time the externals (I know LHA has no external, but the others have, and they are useful as well), and this is prone to failure (because they change) and makes the workflow more complex - the easiest solution is just the container, and it's only two simple lines
    ```yaml
    container:
      image: docker://ghcr.io/nnpdf/bench-evol:latest
    ```
2. the event: you can do the label thing, a la fit-bot, but that's a rather bad solution, at least for our use case, because it does not ensure the workflow is run, so it's not much better than asking people to do it (it is better, because at least the run is recorded, but not much), while the other events are just a few lines, and they achieve the purpose https://github.com/NNPDF/yadism/blob/46ca667626d5f98f98d9e7a01566ec1a7f3d0834/.github/workflows/benchmarks.yml#L5-L15
	  ```yaml
	  push:
	    branches:
	      - master
	  pull_request:
	    types:
	      - closed
	      - review_requested
	  pull_request_review:
	    types:
	      - submitted
      ```

If you wish, you can also add the label, but in that case much better to keep also `workflow_dispatch`, that is just the button :)

AleCandido:
In e8d1f4f I finally fixed all the unit tests, but not the isolated benchmarks yet.

Unfortunately, all workflows are still failing, because part of the isolated benchmarks are run also in the unit tests workflows...

AleCandido:
@felixhekhorn @andreab1997 finally completed, the moment you confirm, we can merge :)

benchmarks/eko/benchmark_evol_to_unity.py
```diff
@@ -10,25 +9,27 @@
...
-    theory.quark_masses.t.value = 173.0
+    theory.heavy.num_flavs_init = 4
+    theory.heavy.intrinsic_flavors = [4, 5]
+    theory.heavy.masses.c.value = 1.0
+    theory.heavy.masses.b.value = 4.75
+    theory.heavy.masses.t.value = 173.0
     operator.mu0 = float(np.sqrt(2))
-    operator.mu2grid = [10]
-    operator.rotations.xgrid = XGrid(np.linspace(1e-1, 1, 30))
+    operator.mugrid = [(10, 5)]
```
andreab1997, line 31: Why now are we adding `10` as `mu` while before we were adding `10` as `mu2`? I expected `np.sqrt(10)`

benchmarks/ekobox/benchmark_evol_pdf.py
```diff
@@ -17,20 +17,20 @@
 def benchmark_evolve_single_member(
     tmp_path, cd, lhapdf_path, theory_card: TheoryCard, operator_card: OperatorCard
 ):
-    mu2grid = [20.0, 100.0, 10000.0]
+    mugrid = [(3.0, 4), (10.0, 5), (100.0, 5)]
```
andreab1997, line 20: Even here I expected `np.sqrt(20)`

andreab1997 (commented):
This is not a full review, I just started looking to the first ~20 files changed. I will continue later on.

benchmarks/eko/benchmark_evol_to_unity.py
AleCandido: Because in principle is irrelevant :)

benchmarks/ekobox/benchmark_evol_pdf.py
AleCandido: Even here it should be irrelevant exactly which scale you're probing. It was a random number before, I used just another one, and avoided a function call.

andreab1997: Ok looks fine

andreab1997 (approved):
Looks good to me, I had a brief look and I did not find anything potentially harmful

AleCandido:
> Looks good to me, I had a brief look and I did not find anything potentially harmful

Apart for the known breaking changes ^^

felixhekhorn (approved):
as discussed I would have kept the `_`, but also fine like this ...

however, I suggest this to be the last in the series of merges as I (naively) expect this the one with the biggest impact (potentially splitting this even in a `v0.14`)

AleCandido:
> however, I suggest this to be the last in the series of merges as I (naively) expect this the one with the biggest impact (potentially splitting this even in a `v0.14`)

Discussing with @andreab1997, I convinced him to accept this change together with the others, thus in v0.13.0, exactly for the reason that we will soon need it, and we want to update our dependants the least amount of times. If we have multiple breaking changes (e.g. including this and N3LO), much better batching them.

Whether you want to merge first something else and then this, I'm perfectly open to discussion. I'd just avoid making a PR ready and reviewed waiting for another one to complete, and merge in `master` as soon as we have them (but if we have other ones reasonably ready, I'm not particularly opinionated on how to sort them).

Issue #122, opened 2022-05-25
Cut `banana` out of `ekobox`


AleCandido:
Maybe the solution is just a forth package, `ekobox` and `ekomark` will depend on it, but at least `ekobox` won't depend on `ekomark`, and we're sure will not depend on `banana`.

felixhekhorn:
@scarlehoff we were thinking about this as a solution for the problems in https://github.com/NNPDF/nnpdf/pull/1529

scarlehoff:
This is in the milestone for 0.9.4 (which has been reelased) but it is not closed. What is the status? (to know whether I have to update the conda recipe or not: https://github.com/conda-forge/eko-feedstock/pull/3)

AleCandido:
The person that released v0.9.4 simply forgot to update the milestone (milestones in EKO have been just introduced, people are not familiar with them...). I'm updating it.

felixhekhorn:
> The person that released v0.9.4 simply forgot to update the milestone (milestones in EKO have been just introduced, people are not familiar with them...). I'm updating it.

sorry :innocent:

AleCandido:
No problem :)

felixhekhorn:
the relevant PR is #226

PR #225, opened 2023-03-14
Polarized ToyLH from genpdf

giacomomagni:
This PR is to allow generation of a Polarised ToyLH pdf as `lhapdf.PDF` object.

@Radonirinaunimi to generate it you can do: 

```
genpdf generate ToyLH_pol 21 1 2 3 4 -p toy_pol -i
```

### DIFF
--- a/src/ekobox/genpdf/__init__.py
+++ b/src/ekobox/genpdf/__init__.py
@@ -60,6 +60,13 @@ def take_data(parent_pdf_set=None, members=False, xgrid=None, Q2grid=None):
             all_blocks.append(
                 [generate_block(toylh.xfxQ2, xgrid, Q2grid, br.flavor_basis_pids)]
             )
+        elif parent_pdf_set in ["toylh_polarized", "toy_pol"]:
+            info = copy.deepcopy(load.Toy_info)
+            toylh = toy.mkPDF("ToyLH_polarized", 0)
+            all_blocks.append(
+                [generate_block(toylh.xfxQ2, xgrid, Q2grid, br.flavor_basis_pids)]
+            )
+
         else:
             info = load.load_info_from_file(parent_pdf_set)
             # iterate on members
@@ -104,6 +111,8 @@ def generate_pdf(
     If `parent_pdf_set` is the name of an available PDF set,
     it will be used as parent. In order to use the toy PDF
     as parent, it is enough to set `parent_pdf_set` to "toy" or "toylh".
+    In order to use the toy Polarized PDF
+    as parent, it is enough to set `parent_pdf_set` to "toylh_polarized" or "toy_pol".
     If `parent_pdf_set` is not specified, a debug PDF constructed as
     x * (1-x) for every flavor will be used as parent.
     It is also possible to provide custom functions for each flavor
@@ -114,7 +123,7 @@ def generate_pdf(
     on custom combinations of PIDs, it is also possible to pass a list
     containing the desired factors for each flavor.
 
-    The default behaviour is to generate only one member for a PDF set
+    The default behavior is to generate only one member for a PDF set
     (the zero member) but it can be changed setting to True the `members` flag.
 
     The `info_update` argument is a dictionary and provide to the user as a way


felixhekhorn (approved):

felixhekhorn:
In a future PR to address #122 we should defer the `from banana import toy` from top level to that specific `elif` (of course likewise for the unpolarized case) (or should we do it here?)

felixhekhorn:
@t7phy in case you add your toy function inside `banana` (as we discussed yesterday) you maybe also want to mirror this PR in the similar way

giacomomagni:
> In a future PR to address #122 we should defer the `from banana import toy` from top level to that specific `elif` (of course likewise for the unpolarized case) (or should we do it here?)

okay fine, seems doable without any downside.

giacomomagni:
> @t7phy in case you add your toy function inside `banana` (as we discussed yesterday) you maybe also want to mirror this PR in the similar way

yes let's fix also that. I can't see any commit banana though.

Radonirinaunimi (approved):

felixhekhorn:
> > @t7phy in case you add your toy function inside `banana` (as we discussed yesterday) you maybe also want to mirror this PR in the similar way
> 
> yes let's fix also that. I can't see any commit banana though.

just to repeat again a piece of private conversation: @t7phy is going for LHAPDF-like benchmarks first

felixhekhorn (approved):
I'm happy with the change, but are you sure this will close #122 completely? (i.e. there is no other dependency?)

giacomomagni:
> I'm happy with the change, but are you sure this will close #122 completely? (i.e. there is no other dependency?)

All the other occurrences of `banana` are in `ekomark`

src/ekobox/genpdf/__init__.py
```diff
@@ -54,12 +53,20 @@ def take_data(parent_pdf_set=None, members=False, xgrid=None, Q2grid=None):
             pid: lambda x, _Q2: x * (1 - x) for pid in br.flavor_basis_pids
         }
     if isinstance(parent_pdf_set, str):
-        if parent_pdf_set in ["toylh", "toy"]:
+        if parent_pdf_set in ["toy_pol", "toy"]:
+            # import banana only here to avoid
+            # explict dependecy
```
giacomomagni, line 58:
```suggestion
            # explict dependency
```

src/ekobox/genpdf/__init__.py
```diff
@@ -54,12 +53,20 @@ def take_data(parent_pdf_set=None, members=False, xgrid=None, Q2grid=None):
             pid: lambda x, _Q2: x * (1 - x) for pid in br.flavor_basis_pids
         }
     if isinstance(parent_pdf_set, str):
-        if parent_pdf_set in ["toylh", "toy"]:
+        if parent_pdf_set in ["toy_pol", "toy"]:
+            # import banana only here to avoid
+            # explict dependency
```
giacomomagni, line 58:
```suggestion
            # explicit dependency
```

AleCandido:
> All the other occurrences of `banana` are in `ekomark`

Not sure this solves the issue, unfortunately, since `ekobox` depends on `ekomark`.

However, I'm fine with this change, if it helps anyhow.

giacomomagni:
> Not sure this solves the issue, unfortunately, since `ekobox` depends on `ekomark`.

True, there is a single point where `ekomark` is used which is [here](https://github.com/NNPDF/eko/blob/e772437e4cd25087b970ab4d260682972272acf0/src/ekobox/evol_pdf.py#L9), so we have to leave the issue open.

felixhekhorn:
> > Not sure this solves the issue, unfortunately, since `ekobox` depends on `ekomark`.
> 
> True, there is a single point where `ekomark` is used which is [here](https://github.com/NNPDF/eko/blob/e772437e4cd25087b970ab4d260682972272acf0/src/ekobox/evol_pdf.py#L9), so we have to leave the issue open.

if that is true, we can/should move that module to `ekobox` (can/should be done in a separate PR) - also content-wise it is easy to argue that this module should belong to `ekobox`: it _is_ a user interface and in fact has nothing to do with benchmarking

felixhekhorn:
> it _is_ a user interface and in fact has nothing to do with benchmarking

just to be fair: it is used inside `ekomark`, because logs are on PDFs - this way we would invert the dependency: `ekomark` then depends on `ekobox`, which maybe is more acceptable

giacomomagni:
> > it _is_ a user interface and in fact has nothing to do with benchmarking
> 
> just to be fair: it is used inside `ekomark`, because logs are on PDFs - this way we would invert the dependency: `ekomark` then depends on `ekobox`, which maybe is more acceptable

Okay let's do one step at the time and address this in another PR, since this is out of the scope of this one.
If there are no other problems I'll merge by the end of the afternoon.

AleCandido:
> if that is true, we can/should move that module to `ekobox` (can/should be done in a separate PR) - also content-wise it is easy to argue that this module should belong to `ekobox`: it _is_ a user interface and in fact has nothing to do with benchmarking

Just notice that that function was originally in `eko`, then in `ekobox`, then in `ekomark`, because we argued that it didn't make sense that the internal module depended on the user-facing one.
Well, it now makes sense. Fine by me :)

PR #226, opened 2023-03-14
Invert ekomark-ekobox dependency

### LINKED ISSUE #122: Cut `banana` out of `ekobox`

felixhekhorn:
Closes #122


AleCandido (approved):
Since #225 is already approved, please merge before that one

giacomomagni (commented):
I'm fine with this. 
I imagine this will break the dependency in many tools right?

AleCandido:
> I'm fine with this. I imagine this will break the dependency in many tools right?

Only those using `apply_pdf` directly, in principle (and outside our packages I'm not sure it is used that much).

andreab1997 (approved):

andreab1997:
Fine for me, I don't see anything harmful. For sure I will need to update something as @giacomomagni was saying but nothing too complicate I believe.

Issue #120, opened 2022-05-20
Improve Notebook usage in docs


felixhekhorn:
To not block #117 let's discuss [the issue raised concerning notebooks](https://github.com/N3PDF/eko/pull/117#discussion_r876884900) here

Maybe we could try something along the lines of https://nbsphinx.readthedocs.io/en/0.8.8/custom-formats.html#Example:-Jupytext

PR #232, opened 2023-03-25
Time Like Splitting Functions

t7phy:


t7phy:
This PR is almost final and ready to be merged, however there is just one thing left that needs to be checked but I am facing a problem in that once I modify the gamma functions, it seems to me that NUMBA is not recompiling them. How do I force NUMBA to recompile after any change I make?

AleCandido:
> This PR is almost final and ready to be merged, however there is just one thing left that needs to be checked but I am facing a problem in that once I modify the gamma functions, it seems to me that NUMBA is not recompiling them. How do I force NUMBA to recompile after any change I make?

Cancel the `__pycache__` directories.

t7phy:
> > This PR is almost final and ready to be merged, however there is just one thing left that needs to be checked but I am facing a problem in that once I modify the gamma functions, it seems to me that NUMBA is not recompiling them. How do I force NUMBA to recompile after any change I make?
> 
> Cancel the `__pycache__` directories.
> 
> https://numba.pydata.org/numba-doc/dev/developer/caching.html#cache-clearing

I actually did delete them in the directory where the as files are stored, however that did not force recompile.

AleCandido:
If you want to just remove Numba cached files, you can filter them with:

```sh
find . -type f | grep '\.nb[i|c]$' | xargs rm -f
```

t7phy:
> If you want to just remove Numba cached files, you can filter them with:
> 
> ```shell
> find . -type f | grep '\.nb[i|c]$' | xargs rm -f
> ```
> 
> However, this would not fix your problem here:
> 
> > I actually did delete them in the directory where the as files are stored, however that did not force recompile.
> 
> Can you explicitly write exactly which directory you canceled?

eko/src/ekore/anomalous_dimensions/unpolarized/time_like/__pycache__

tests/ekore/anomalous_dimensions/unpolarized/time_like/test_init.py
```diff
@@ -2,9 +2,8 @@
 
 import ekore.anomalous_dimensions.unpolarized.time_like as ad_ut
 
-
-def test_init():
-    with pytest.raises(NotImplementedError):
-        ad_ut.gamma_ns((1, 0), 0, 1.0, 4)
-    with pytest.raises(NotImplementedError):
-        ad_ut.gamma_singlet((1, 0), 1.0, 4)
+# def test_init():
```
AleCandido, line 5: Either you keep it, or you drop it.

I guess now this has been implemented, so the commented test is outdated. Please, replace with suitable tests.

tests/eko/evolution_operator/test_init.py
```diff
@@ -17,7 +17,7 @@
 
 
 def test_quad_ker_errors():
-    for p, t in [(False, True), (True, True)]:
+    for p, t in [(True, True)]:
```
AleCandido, line 20: Replace also this one with suitable tests for implemented objects.

AleCandido: Moreover, you can drop the `for` loop, and just inline the value of `p` and `t`.

This test is now specific for the polarized + time-like case.

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
```diff
@@ -0,0 +1,693 @@
+"""The unpolarized time-like NLO Altarelli-Parisi splitting kernels."""
+
+import numba as nb
+import numpy as np
+from numpy import power as npp
```
AleCandido, line 5: The comment to the other PR still holds: this is just the Python built-in `**` operator.

Built-ins are better because more elementary, and more familiar to the reader.
However, it is definitely not top-priority.

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
```diff
@@ -0,0 +1,693 @@
...
+                    - 423 * npp(N, 8)
+                    - 76 * npp(N, 9)
+                    - 288 * zeta2 * npp(N, 2)
+                    - 720 * zeta2 * npp(N, 3)
+                    - 504 * zeta2 * npp(N, 4)
+                    + 180 * zeta2 * npp(N, 5)
+                    + 576 * zeta2 * npp(N, 6)
+                    + 504 * zeta2 * npp(N, 7)
+                    + 216 * zeta2 * npp(N, 8)
+                    + 36 * zeta2 * npp(N, 9)
```
AleCandido, line 351: Collect as much as possible: it simplifies the evaluation, and improves readability.

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
AleCandido: I assume that Numba compiles `npp` and `**` to the same thing, including support for complex numbers. However, this can be verified with a very brief example :)

If I'm wrong, ignore the comment, of course ^^

tests/ekore/anomalous_dimensions/unpolarized/time_like/test_init.py
t7phy: Yes, I will definitely fix this but first I would like to sort the above issue. Do you have any suggestions on why even deleting the pycache doesn't force recompile?

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
t7phy: I just followed the usage as was done in space like case (for consistency) which also uses numpy.power.

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
t7phy: I had used mathematica simplifications to obtain all the expressions.

AleCandido:
> `eko/src/ekore/anomalous_dimensions/unpolarized/time_like/__pycache__`

In principle, it is the correct one. However, try to cancel them all, at least inside `anomalous_dimensions`.

Worst case solution is of course to reclone the repo in a new location. However, that might work for multiple reasons: Numba objects have to be stored somewhere, but that somewhere might be local, but also non-local to the project (e.g. in case you didn't install EKO in development mode, or if you set Numba to use a global cache).

But, as a first attempt, if deleting is not working, you can try to edit the file you want to recompile. I'm not sure if Numba uses the time-stamp or a file hash, but changing the file content will change both of them.

t7phy:
> > `eko/src/ekore/anomalous_dimensions/unpolarized/time_like/__pycache__`
> 
> In principle, it is the correct one. However, try to cancel them all, at least inside `anomalous_dimensions`.
> 
> Worst case solution is of course to reclone the repo in a new location. However, that might work for multiple reasons: Numba objects have to be stored somewhere, but that somewhere might be local, but also non-local to the project (e.g. in case you didn't install EKO in development mode, or if you set Numba to use a global cache).
> 
I had installed in as per documentations for installing EKO for development.

> But, as a first attempt, if deleting is not working, you can try to edit the file you want to recompile. I'm not sure if Numba uses the time-stamp or a file hash, but changing the file content will change both of them.

That is indeed the problem. Modifying the file should lead numba to recompile but it is not doing that. Yesterday, I literally deleted the entire content of as2.py and saved it, then I ran NLO benchmarks and they still ran. So something very weird is happening, which I don't understand.

tests/eko/evolution_operator/test_init.py
t7phy: This is done

t7phy:
Hi @felixhekhorn and @AleCandido , the benchmarks have now been successful for all orders. This implementation is now ready for merging. I await your reviews before pressing the green button :) 
cc. @enocera

tests/eko/evolution_operator/test_init.py
```diff
@@ -17,32 +17,31 @@
 
 
 def test_quad_ker_errors():
-    for p, t in [(False, True), (True, True)]:
```
felixhekhorn, line 20: you can not remove existing tests :upside_down_face:

tests/ekore/anomalous_dimensions/unpolarized/time_like/test_init.py
```diff
@@ -1,10 +0,0 @@
-import pytest
-
-import ekore.anomalous_dimensions.unpolarized.time_like as ad_ut
-
-
-def test_init():
```
felixhekhorn, line 6: don't simply delete - replace with something real

src/ekore/anomalous_dimensions/unpolarized/time_like/as1.py
felixhekhorn, line 1: remember you promised unit tests for things which are the same in space-like

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
```diff
@@ -0,0 +1,687 @@
...
+    S12 = S11 + 1 / N2
+    S13 = S12 + 1 / N3
+    S14 = S13 + 1 / N4
+    S15 = S14 + 1 / N5
+    S16 = S15 + 1 / N6
+
+    ZETA2 = zeta2
+    ZETA3 = zeta3
+
+    SPMOM = (
```
felixhekhorn, line 61: you can do here or later (i.e. with #202 ): remember that we use different implementations - this should be something like $g_3$, if I remember correctly

src/ekore/anomalous_dimensions/unpolarized/time_like/as1.py
```diff
@@ -0,0 +1,156 @@
+"""The unpolarized LO Altarelli-Parisi splitting kernels."""
```
felixhekhorn, line 1:
```suggestion
"""The unpolarized |LO| Altarelli-Parisi splitting kernels."""
```
and similar

src/ekore/anomalous_dimensions/unpolarized/time_like/as3.py
```diff
@@ -0,0 +1,791 @@
...
+
+from ....harmonics import w1, w2, w3, w4
+from ....harmonics.constants import zeta2, zeta3
+
+
+@nb.njit(cache=True)
+def gamma_nsp(N, nf):
+    r"""Compute the NNLO non-singlet positive anomalous dimension.
+
+    Based on :cite:`Mitov:2006ic`.
```
felixhekhorn, line 15: can you give an equation?

benchmarks/FF_LHAPDF_bench.py
```diff
@@ -0,0 +1,149 @@
+"""Benchmark FFs from LHAPDF"""
+
+from banana import register
+
+from eko import interpolation
+from ekomark.benchmark.runner import Runner
+
+register(__file__)
+
+base_operator = {"ev_op_iterations": 50, "backward_inversion": "exact"}
```
felixhekhorn, line 10: are you sure 50 is needed? I'm slightly worried we might run into numerical problems ... we usually use 10, isn't that sufficient?

benchmarks/FF_LHAPDF_bench.py
```diff
@@ -0,0 +1,149 @@
...
+        }
+
+        operator_card = {
+            **base_operator,
+            "Q2grid": list(Q2grid),
+            "time_like": True,
+            "interpolation_xgrid": interpolation.lambertgrid(100, 0.01),
+        }
+
+        self.run([theory_card], [operator_card], [FF_sets_lo[7]])
```
felixhekhorn, line 105: why not taking the PDF index as argument?

felixhekhorn (changes requested):
please provide
- [ ] unit tests
- [x] documentation

tests/eko/evolution_operator/test_init.py
felixhekhorn: sorry - I confused myself, you're right!

src/ekore/anomalous_dimensions/unpolarized/time_like/as1.py
t7phy: done

benchmarks/FF_LHAPDF_bench.py
t7phy: changed to 10, it seems to be fine too

benchmarks/FF_LHAPDF_bench.py
t7phy: I wanted to have the option to just change FF sets while benchmarking in a quick way, since there are very few of them, it seemed simple enough to have them in a list in the file itself

src/ekore/anomalous_dimensions/unpolarized/time_like/as1.py
felixhekhorn: please also modify the docstrings of the fncs

benchmarks/FF_LHAPDF_bench.py
```diff
@@ -0,0 +1,149 @@
...
+    "NNFF10_PIm_nlo",
+    "NNFF10_PIp_nlo",
+    "NNFF10_PIsum_nlo",
+    "NNFF10_KAm_nlo",
+    "NNFF10_KAp_nlo",
+    "NNFF10_KAsum_nlo",
+    "NNFF10_PRm_nlo",
+    "NNFF10_PRp_nlo",
+    "NNFF10_PRsum_nlo",
+    "MAPFF10NLOPIm",
```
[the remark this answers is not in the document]
t7phy: we mostly use MAP ones instead of NN where ever possible. However, as far as settings as described on info pages of these FF sets go, they are mostly same.

benchmarks/FF_LHAPDF_bench.py
felixhekhorn: just fine - simply add an argument `pdf_index=7` such that there are less hardcoded stuff

t7phy:
> please provide
> 
>     * [ ]  unit tests
> 
>     * [ ]  documentation

documentation, as in the readthedocs page?

src/ekore/anomalous_dimensions/unpolarized/time_like/as1.py
t7phy: The problem is that I am not aware of any clear cut unit tests for the time-like splitting functions (otherwise they would already be there). Do you have any in mind that might work? @felixhekhorn @enocera

src/ekore/anomalous_dimensions/unpolarized/time_like/as1.py
t7phy: also done

tests/ekore/anomalous_dimensions/unpolarized/time_like/test_init.py
t7phy: looking at the analogous tests in space-like case, I understand the ones which check the dimensions of the output, however the others are not fully clear to me. perhaps we could discuss a little about this over zoom?

src/ekore/anomalous_dimensions/unpolarized/time_like/as1.py
felixhekhorn: here specifically, I seem to recall that `tl.as1.gamma_qq` === `sl.as1.gamma_gg` and same for `qq`. Instead, for the off-diagonal stuff mirrored

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
felixhekhorn, line 1: as said, here I would add some Mela reference numbers a unit tests

src/ekore/anomalous_dimensions/unpolarized/time_like/as3.py
felixhekhorn:
```suggestion
    Implements :eqref:`15` from  :cite:`Mitov:2006ic` via the N-space translation from A. Vogt.
```
and similar ... This _is_ the reference if A. Vogt is saying so ... so, we should give him due credit (maybe even add a own bibtex entry?) and also the paper. The fact that it is a different representation is not relevant for the reference here (though it is for the content of the function) ...

benchmarks/FF_LHAPDF_bench.py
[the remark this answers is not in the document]
felixhekhorn: what does "mostly" mean? if they are not the same and you don't want same, maybe drop them? or if they are the exact same then keep them ...

src/ekore/anomalous_dimensions/unpolarized/time_like/as1.py
t7phy: I have added individual tests for each splitting functions, do you also want me to add a check between space like and time like ones?

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
t7phy: let's leave this for the time being and come back during #202 . As these are taken from Vogt's parametrized implementation, it's safer to let them be for the time being (as computationally they will only make a difference after #202).

benchmarks/FF_LHAPDF_bench.py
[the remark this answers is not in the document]
t7phy: well, that depends on what "settings" mean? If you mean settings that specifically concern the benchmark, then I look at nf, xmin, xmax, for example, and  these are the same. However, if you look at Qmax, these are different, but I do not look at them because the Q to which I evolve to is allowed by both. That is why I said "mostly".

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
felixhekhorn: Can you add a comment somewhere saying "The reference values are coming from Mela" or similar?

src/ekore/anomalous_dimensions/unpolarized/time_like/as3.py
felixhekhorn: the reference equation for nsm is Eq. 16: "for both ξ = − and ξ = v. " - to be exact, it is implicit defined there (due to the difference on the lhs). As for the website, I think it is also fine to just say "coming from A. Vogt" people in the field will know ... the only alternative, as said, can be a `misc` bibtex entry ... decide yourself

t7phy: No no, I am not talking about NNLO anymore, for which I have already included the equations as you suggested. I am talking about NLO, where just for the NSP and NSM, I had to use the parametrized expressions from Vogt, however the paper that he mentions on his page only contains NSP and not NSM.

felixhekhorn: then for ns- I would cite "Mitov:2006wy Eq. B.8 with a parametrization provided by A. Vogt" (by the way right above these equations there is again the statement about odd and even moments ...)

t7phy: well, what happened is I had initially taken all expressions from Mitov:2006wy paper, used mathematica to get them in the form which can then be used in python. It appears that the non singlet expressions in the paper might contain some errors and therefor the nsp and nsm were replaced with Vogt's parametrization whereas the qqs, qg, gq, and gg were left as they were (from Mitov's paper). So for nsp and nsm, the citation has to be as per Vogt here, for which as I mentioned above, the paper he mentions does not contain nsm as far as I can tell.

felixhekhorn: fine, I accept as is
PS: last thing: did you see Gluck:1992zx has an erratum?

t7phy: Yes, the erratum does not concern us, as it is for P_gg for which, I use Mitov's expression.

tests/ekore/anomalous_dimensions/unpolarized/time_like/test_as3.py
```diff
@@ -1,3 +1,5 @@
+# it is possible to obtain result values for testing from vbertone/MELA repo
```
AleCandido, line 1:
```suggestion
"""Test values have been obtained running MELA.

MELA is an alternative Mellin space evolution code, hosted at https://github.com/vbertone/MELA.

"""
```

The second part you can put in a single place, and omit from individual test modules. Even in the docs, if you wish :)

tests/ekore/anomalous_dimensions/unpolarized/time_like/test_as3.py
```diff
@@ -0,0 +1,57 @@
+# Testing values obtained from |MELA| routines
```
AleCandido, line 1: For module comments, function and class comments, and even globals descriptions, always prefer docstrings.
```suggestion
"""Testing values obtained from |MELA| routines."""
```

doc/source/theory/TimeLike.rst
```diff
@@ -0,0 +1,26 @@
...
+The relevant setting in the operator card is thus called ``time_like = True``.
+
+We implement the time-like |DGLAP| anomalous dimensions up to |NNLO| in :class:`~ekore.anomalous_dimensions.unpolarized.time_like`.
+The implementation for the |LO| and |NLO| splitting functions is based on :cite:`Mitov:2006wy, Gluck:1992zx` and the implementation for
+the |NNLO| splitting functions is based on :cite:`Mitov:2006ic, Moch:2007tx, Almasy:2011eq`.
+Supplying new anomalous dimensions and new matching conditions is the only change required for the eko program (e.g. the
+solution strategies are unaffected).
+
+The time-like splitting functions have been benchmarked using select |FF| sets from |LHAPDF|. In addition, they have been tested 
+against routines from |MELA|, which is an alternative Mellin Space evolution code, available at https://github.com/vbertone/MELA.
```
felixhekhorn, line 26: you have to define the abbreviation for Mela - and the existing abbreviation for lhapdf is with lower case... please check this page also live! you can do so by `poe docs` and then `poe docs-server` for more details also check the `pyproject.toml`

doc/source/theory/TimeLike.rst
```diff
@@ -0,0 +1,26 @@
...
+The relevant setting in the operator card is thus called ``time_like = True``.
+
+We implement the time-like |DGLAP| anomalous dimensions up to |NNLO| in :class:`~ekore.anomalous_dimensions.unpolarized.time_like`.
+The implementation for the |LO| and |NLO| splitting functions is based on :cite:`Mitov:2006wy, Gluck:1992zx` and the implementation for
+the |NNLO| splitting functions is based on :cite:`Mitov:2006ic, Moch:2007tx, Almasy:2011eq`.
+Supplying new anomalous dimensions and new matching conditions is the only change required for the eko program (e.g. the
+solution strategies are unaffected).
+
+The time-like splitting functions have been benchmarked using select |FF| sets from |LHAPDF|. In addition, they have been tested 
+against functions from |MELA|, which is an alternative Mellin Space evolution code, available at https://github.com/vbertone/MELA.
```
felixhekhorn, line 26:
```suggestion
The time-like evolution has been benchmarked using selected |FF| sets from |LHAPDF|. In addition, the anomalous dimensions have been tested 
against functions from |MELA|, which is an alternative Mellin Space evolution code, available at https://github.com/vbertone/MELA.
```

doc/source/shared/abbreviations.rst
```diff
@@ -111,7 +111,9 @@
 .. |T| raw:: html
 
    <span class="green">✓</span>
+.. |MELA| raw:: html
 
+   <a href="https://github.com/vbertone/mela"><abbr title="a Mellin Evolution LibrAry">
```
felixhekhorn, line 116: you did not look at the page ... else you would have seen this:
grafik
which very likely is not what you want ...
```suggestion

.. |MELA| raw:: html

   <a href="https://github.com/vbertone/mela"><abbr title="a Mellin Evolution LibrAry">MELA</abbr></a>
```

t7phy: I made the changes you suggested was unable to run it because the poe docs was running giving errors, but I have fixed it now.

felixhekhorn: Then we need to fix these errors! what is the error? did you install the necessary packages? there is a poetry group or an python extra - again look to `pyproject.toml`

t7phy: No, nothing to fix, I was missing pandoc, its working now.

felixhekhorn: Are you saying we have an undeclared dependency on pandoc? coming from where?

AleCandido: > Are you saying we have an undeclared dependency on pandoc? coming from where?

It is used by `nbsphinx` https://nbsphinx.readthedocs.io/en/0.9.1/installation.html#pandoc

AleCandido: If we find a better alternative to Jupyter Notebooks, we could consider dropping them (a better alternative -> code that is easy to run, not just written there, and possibly interleaved with text).

Maybe we can think to resort to doctests, we're not using pictures anyhow, nor tables or any other kind of fancy output, so code and text should be supported enough by Sphinx itself.

felixhekhorn: > It is used by `nbsphinx` https://nbsphinx.readthedocs.io/en/0.9.1/installation.html#pandoc

Fine, but `nbsphinx` is in the `docs` group ... as far as I understand `pip` is not sufficient, but `apt` is needed ...

felixhekhorn: There is already a dedicated issue for this discussion #120

felixhekhorn:
You may not be responsible for everything, but for the major part yes: [please run always pre-commit](https://results.pre-commit.ci/run/github/212659217/1680690547.l60YlYIWRke4FyA7wdL40A)

We have finally the CI check back! so I can stop complaining personally and let the red cross speak ...

t7phy:
> You may not be responsible for everything, but for the major part yes: [please run always pre-commit](https://results.pre-commit.ci/run/github/212659217/1680690547.l60YlYIWRke4FyA7wdL40A)
> 
> We have finally the CI check back! so I can stop complaining personally and let the red cross speak ...

but it has always been active for me. I do not know why it never checked the test files or the doc files, but I have no reason to disable it.

felixhekhorn:
> but it has always been active for me. I do not know why it never checked the test files or the doc files, but I have no reason to disable it.

well, the red cross says something different - and for the fixes in the last two commits and did nothing more then add a random new line somewhere and run `pre-commit` - so my local version is the CI version ...

in any case I will not attempt the docstyle fixes here to not further spoil this PR, but open a new dedicated one

t7phy:
> > but it has always been active for me. I do not know why it never checked the test files or the doc files, but I have no reason to disable it.
> 
> well, the red cross says something different - and for the fixes in the last two commits and did nothing more then add a random new line somewhere and run `pre-commit` - so my local version is the CI version ...
> 
> in any case I will not attempt the docstyle fixes here to not further spoil this PR, but open a new dedicated one

So I think everything in this PR is fine now?

t7phy:
> > So I think everything in this PR is fine now?
> 
> let me run the benchmarks first ... no idea how you managed without [7e325cf](https://github.com/NNPDF/eko/commit/7e325cf124f84a1616df354048e520a0d6c81d7d) and [193afc8](https://github.com/NNPDF/eko/commit/193afc8fc2b562dd6f247ffaea34ef309b48d8b5) : they are strictly necessary ...

I ran all benchmarks before the shift was made from Q2 to mu.

felixhekhorn (approved):
I believe this is ok now

two remarks
- I think this is significant slower than space-like counterparts because here is no reuse of harmonic sums at all (which proves they are a bottleneck) ... the problem can be tackled in #202 
- at NNLO V has a bigger error O(0.2%) than S O(0.04%) in the small x region (where small here x=0.01) which is a bit odd, but fine

AleCandido:
After isolated benchmarks completion feel free to merge. Problems with `pre-commit` are not caused by this PR.

PR #236, opened 2023-03-29
Fix docs

andreab1997:
According to #222 I am correcting the docs, in particular minus signs, and reconciling with the source papers. @felixhekhorn  if you can please add the part regarding the thresholds for ekos and alphas.


felixhekhorn:
ported as https://github.com/NNPDF/eko/pull/222/commits/69ef82f3a401e9fd76310b551aaf98bf2ffc01b2 to #222

Issue #157, opened 2022-10-20
Drop Q2 everywhere


AleCandido:
We should update code and docs not to use $Q^2$ in place of $\mu_F^2$, in order to avoid ambiguity, and resulting confusion and bugs.

- [ ] fix docs
- [ ] fix code

Also $x$ would be better replaced by $z$ (but here it is less crucial).

AleCandido:
A step forward has been done in #172, but there has been no check that the transition consistently happened in the code, and for sure there has been no transition in the docs.

AleCandido:
`Q2grid` has been dropped in 6b830a910110bbde3756c9dc8317e5eb99ff2c75

Verify the status of transition with:
```bash
rg `q2` -i -l
```

Some occurrences of `q2` are going to be kept for a while, since they're used in legacy compatibility.

felixhekhorn:
even more: everything should be `evolgrid` by now

Issue #175, opened 2022-12-08
Legacy to versions


AleCandido:
At the moment, since we are transitioning, there are a lot of legacy objects:
- legacy runner
- legacy runcards
- legacy output

About the output, there is already an issue discussing the opportunity of versioning it #106 
But we need to version all the input and output we use, while the runner is a different story.

So, we need to act in two different ways.
For the IO, we will declare a candidate v1 to the current format, for both inputs (runcards) and output (operator). What is now called `legacy` will be instead the v0, and all the code dealing with legacy IO will be moved to `eko.io.v0`.
The only supported actions will be loading and upgrading to v1 format. Everything else will be dropped.
If in future we will come to a v2 version, we will create a `eko.io.v1` package, with loaders and upgraders to v2.

The only thing we need to decide is if versioning the input and output together (a single `__data_version__`, as it is at the moment), or using two separate version numbers.
Only integer versions will be supported, since any compatible change does not require a new version.

About the runner, we will simply reimplement the current one with new elements, and find a better name than `legacy`.

felixhekhorn:
I'm inclined to say let's split input and output - since they are for the most part decoupled. The one and only interaction they have, is that the output should contain the input cards.

The runner version is the program version (so v0.11 atm) and it need to be able to act on both: suitable input and output

AleCandido:
Ok, agreed for input and output, I'll make a proposal for this after #172 

There will be an extra piece, i.e. import theory from *current* NNPDF layout (it will update with the theory db layout itself) to *current* theory version. At the moment, it is mostly a copy of the `v0` theory, with the only difference that `v0` theory will be frozen, and it contains a few extra element (that are not present in DB entries).

Issue #97, opened 2022-02-23
Lazy loading for rust


AleCandido:
We'd like PineAPPL as well to be able to consume an EKO one `Q2` at a time, since evolving grids requesting a huge number of `Q2` points would require splitting the grid otherwise.

In order to do this, the easiest is to directly provide and maintain a Rust crate that is able to manage the EKO output format.
Indeed, PyO3 would make it possible to run Python code from Rust, but it is more a handle on the interpreter itself, rather than proper bindings in this direction (even though it might be used for this, but it can be painful).

So, the easiest thing it's to make a standalone crate in this repository (unfortunately [`eko`](https://crates.io/crates/eko) is already lost as a name, but we can still use something like *get eko*, i.e. `geko`).
In order to do this, we don't really need many ingredients:
- [ ] a [`tar`](https://crates.io/crates/tar) library, for which the one linked is an obvious candidate
- [ ] a [`yaml`](https://crates.io/crates/yaml-rust) library, that is another easy choice
- [ ] an `npy` library, and here it becomes difficult; the alternatives are
  1. [`npyz`](https://crates.io/crates/npyz) not much maintained, but should be complete enough 
  2. [`ndarray-npy`](https://crates.io/crates/ndarray-npy), slightly more maintained in the recent future, but it does not seem to have a complete feature set
  3. [`npy`](https://crates.io/crates/npy), that is mentioned by the previous, and the number 1. claims to be its fork, but it looks untouched from 2018

felixhekhorn:
as you already said: i. or ii. seem to be likely candidates ...

AleCandido:
Yes, but unfortunately no one of them is very much maintained, so we have to be ready to fix them ourselves (if needed).

We might want to have a look at the code, `.npy` is not a very complicated format, it is just the binary dump of an array + some headers. In any case, the crates are a better starting point than writing everything from scratch, if we're lucky they will completely fulfill our purpose.

felixhekhorn:
maybe we should provide the possibility to do the `xgrid_reshape` on get-time

felixhekhorn:
@alecandido can you please remind me what you meant by "operators headers syncer" above?

alecandido:
It's more than two years ago, but I guess it was:
https://github.com/NNPDF/eko/blob/3bfc89d42c4a501dc7d4ec75aa48f00947e71143/src/eko/io/inventory.py#L186-L203
and friends (e.g. `__iter__`)

[...]

PR #242, opened 2023-04-04
First jets implementation

AleCandido:
Supersedes #217 

- [x] complete `OperatorGrid` migration
- [x] add inventories initialization
- [x] restore old runner
	- it will be dropped in the next PR, but let's use one last time to benchmark the new one
- [x] fix tests in the workflow
- [x] fix isolated benchmarks
- [x] fix benchmarks
- [x] check `managed` runner equivalence with `legacy`
- [x] revert default `solve`
    https://github.com/NNPDF/eko/blob/ddc38952c375b40cc1160fb0541bcd3c68d12c48/src/eko/runner/__init__.py#L55-L56
- [x] check and fix tutorials
- ~~src/eko still contains some q2 in names or docstrings, you might want to update~~
  - postponed, and already tracked by #157
- [x] decide about normalize

### DIFF
--- a/src/eko/couplings.py
+++ b/src/eko/couplings.py
@@ -8,7 +8,6 @@ See :doc:`pQCD ingredients </theory/pQCD>`.
 
 """
 import logging
-import warnings
 from typing import Iterable, List
 
 import numba as nb
--- a/src/eko/couplings.py
+++ b/src/eko/couplings.py
@@ -453,6 +453,7 @@ class Couplings:
         # create new threshold object
         self.a_ref = np.array(couplings.values) / 4.0 / np.pi  # convert to a_s and a_em
         matching_scales = (np.array(masses) * np.array(thresholds_ratios)).tolist()
+        self.thresholds_ratios = list(thresholds_ratios)
         self.atlas = matchings.Atlas(matching_scales, (couplings.scale**2, nf_ref))
         self.hqm_scheme = scheme_name
         logger.info(
@@ -744,7 +745,7 @@ class Couplings:
             # - if there is yet a step to go
             if k < len(path) - 1:
                 # q2_to is the threshold value
-                L = np.log(self.atlas.walls[seg.nf - shift + 1])
+                L = np.log(self.thresholds_ratios[seg.nf - shift])
                 m_coeffs = (
                     compute_matching_coeffs_down(self.hqm_scheme, seg.nf - 1)
                     if is_downward
--- a/src/eko/couplings.py
+++ b/src/eko/couplings.py
@@ -744,7 +744,6 @@ class Couplings:
             # apply matching conditions: see hep-ph/9706430
             # - if there is yet a step to go
             if k < len(path) - 1:
-                # q2_to is the threshold value
                 L = np.log(self.thresholds_ratios[seg.nf - shift])
                 m_coeffs = (
                     compute_matching_coeffs_down(self.hqm_scheme, seg.nf - 1)
--- a/src/eko/io/inventory.py
+++ b/src/eko/io/inventory.py
@@ -102,10 +102,9 @@ class Inventory(Generic[H]):
 
         # in case of contentless, check header availability instead
         if self.contentless:
-            if self.lookup(stem, header=True):
-                self.cache[header] = None
-                return None
-            raise LookupError(f"Impossible to retrieve {header} from {self}.")
+            self.lookup(stem, header=True)
+            self.cache[header] = None
+            return None
 
         # for contentful inventories, check operator availability
         oppath = self.lookup(stem)
@@ -117,7 +116,7 @@ class Inventory(Generic[H]):
         return op
 
     def __setitem__(self, header: H, operator: Optional[Operator]):
-        """Set operator for given :math:`mu^2`.
+        """Set operator for given header.
 
         Header and operator are automatically dumped on disk.
 
@@ -178,6 +177,10 @@ class Inventory(Generic[H]):
 
         """
         yield from self.cache
+
+    def __len__(self):
+        """Return the number of elements in the cache."""
+        return len(self.cache)
 
     def sync(self):
         """Sync the headers in the cache with the content on disk.
--- a/src/eko/io/inventory.py
+++ b/src/eko/io/inventory.py
@@ -201,3 +201,8 @@ class Inventory(Generic[H]):
     def __invert__(self):
         """Alias for :meth:`sync`."""
         self.sync()
+
+    def empty(self):
+        """Empty the in-memory cache."""
+        for header in self.cache:
+            del self[header]
--- a/src/eko/io/items.py
+++ b/src/eko/io/items.py
@@ -23,7 +23,8 @@ class Header:
 class Evolution(Header):
     """Information to compute an evolution operator.
 
-    Describe evolution with a fixed number of light flavors.
+    It describes the evolution with a fixed number of light flavors between two
+    scales.
 
     """
 
@@ -43,7 +44,7 @@ class Evolution(Header):
 
     Intermediate ones always have final scales :attr:`mu2` corresponding to
     matching scales, and initial scales :attr:`mu20` corresponding to either
-    matching scales or the global initial scale of the EKO.
+    matching scales or the global initial scale of the |EKO|.
 
     Note
     ----
@@ -58,15 +59,15 @@ class Evolution(Header):
                                                             |
         nf = 6 --------------------------------------------------------
 
-    where each lane corresponds to DGLAP evolution with the relative number of
-    running flavors, and the vertical bridges are the perturbative matchings
+    where each lane corresponds to |DGLAP| evolution with the relative number
+    of running flavors, and the vertical bridges are the perturbative matchings
     between two different "adjacent" schemes.
 
     """
 
     @classmethod
     def from_atlas(cls, segment: matchings.Segment, cliff: bool = False):
-        """Create instance from analogous Atlas object."""
+        """Create instance from analogous :class:`eko.matchings.Atlas` object."""
         return cls(**asdict(segment), cliff=cliff)
 
     @property
@@ -89,7 +90,7 @@ class Matching(Header):
 
     @classmethod
     def from_atlas(cls, matching: matchings.Matching, inverse: bool = False):
-        """Create instance from analogous Atlas object."""
+        """Create instance from analogous :class:`eko.matchings.Atlas` object."""
         return cls(**asdict(matching), inverse=inverse)
 
     @property
@@ -123,8 +124,8 @@ class Target(Header):
 class Operator:
     """Operator representation.
 
-    To be used to hold the result of a computed 4-dim operator (from a given
-    scale to another given one).
+    To be used to hold the result of a computed 4-dim operator (either a raw
+    evolution operator or a matching condition).
 
     Note
     ----
--- a/src/eko/io/items.py
+++ b/src/eko/io/items.py
@@ -5,8 +5,8 @@ from typing import BinaryIO, Optional, Union
 
 import lz4.frame
 import numpy as np
-import numpy.lib.npyio as npyio
 import numpy.typing as npt
+from numpy.lib import npyio
 
 from .. import matchings
 from . import exceptions
--- a/src/eko/io/runcards.py
+++ b/src/eko/io/runcards.py
@@ -174,7 +174,7 @@ class Legacy:
         new["order"] = [old["PTO"] + 1, old["QED"]]
         alphaem = self.fallback(old.get("alphaqed"), old.get("alphaem"), default=0.0)
         ms = self.heavies("m%s", old)
-        mus = self.heavies("Qm%s", old)
+        ks = self.heavies("k%sThr", old)
         new["couplings"] = dict(
             alphas=old["alphas"],
             alphaem=alphaem,
@@ -183,7 +183,7 @@ class Legacy:
             max_num_flavs=old["MaxNfAs"],
         )
         new["heavy"] = {
-            "num_flavs_init": nf_default(old["Qref"] ** 2.0, default_atlas(ms, mus))
+            "num_flavs_init": nf_default(old["Qref"] ** 2.0, default_atlas(ms, ks))
             if old["nf0"] is None
             else old["nf0"],
             "num_flavs_max_pdf": old["MaxNfPdf"],
@@ -198,6 +198,7 @@ class Legacy:
         if old["HQ"] == "POLE":
             new["heavy"]["masses"] = [[m, nan] for m in ms]
         elif old["HQ"] == "MSBAR":
+            mus = self.heavies("Qm%s", old)
             new["heavy"]["masses"] = [[m, mu] for m, mu in zip(ms, mus)]
         else:
             raise ValueError(f"Unknown mass scheme '{old['HQ']}'")
--- a/src/eko/io/struct.py
+++ b/src/eko/io/struct.py
@@ -200,6 +200,14 @@ class EKO:
         """Drop operator from memory."""
         del self.operators[Target.from_ep(ep)]
 
+    def __delattr__(self, name: str):
+        """Empty an inventory cache."""
+        attr = getattr(self, name)
+        if isinstance(attr, Inventory):
+            attr.empty()
+        else:
+            super().__delattr__(name)
+
     @contextlib.contextmanager
     def operator(self, ep: EPoint):
         """Retrieve an operator and discard it afterwards.
--- a/src/eko/io/struct.py
+++ b/src/eko/io/struct.py
@@ -182,10 +182,6 @@ class EKO:
             else:
                 self.recipes_matching[recipe] = None
 
-    def __ilshift__(self, recipes: List[Recipe]):
-        """Shortcut for :meth:`load_recipes`."""
-        self.load_recipes(recipes)
-
     # operator management
     # -------------------
     def __getitem__(self, ep: EPoint) -> Optional[Operator]:
@@ -374,7 +370,7 @@ class EKO:
         elif mode in "a":
             load = True
         else:
-            raise ValueError
+            raise ValueError(f"Unknown file mode: {mode}")
 
         tmpdir = pathlib.Path(tempfile.mkdtemp(prefix=TEMP_PREFIX))
         if load:
--- a/src/eko/matchings.py
+++ b/src/eko/matchings.py
@@ -160,11 +160,12 @@ class Atlas:
 
         prev = path[0]
         matched: MatchedPath = [prev]
-        for seq in path[1:]:
-            matching = Matching(prev.target, max(prev.nf, seq.nf))
+        for seg in path[1:]:
+            matching = Matching(prev.target, max(prev.nf, seg.nf))
 
             matched.append(matching)
-            matched.append(seq)
+            matched.append(seg)
+            prev = seg
 
         return matched
 
--- a/src/eko/matchings.py
+++ b/src/eko/matchings.py
@@ -194,7 +194,7 @@ def is_downward_path(path: Path) -> bool:
 
     - in the number of active flavors when the path list contains more than one
       :class:`PathSegment`, note this can be different from each
-      :attr:`PathSegment.is_downward_q2`
+      :attr:`PathSegment.is_downward`
     - in :math:`Q^2` when just one single :class:`PathSegment` is given
 
     """
--- a/src/eko/matchings.py
+++ b/src/eko/matchings.py
@@ -30,7 +30,7 @@ class Segment:
 
     def __str__(self):
         """Textual representation, mainly for logging purpose."""
-        return f"PathSegment({self.origin} -> {self.target}, nf={self.nf})"
+        return f"Segment({self.origin} -> {self.target}, nf={self.nf})"
 
 
 Path = List[Segment]
@@ -188,14 +188,14 @@ def nf_default(mu2: SquaredScale, atlas: Atlas) -> FlavorsNumber:
 
 
 def is_downward_path(path: Path) -> bool:
-    """Determine if a path is downward.
+    r"""Determine if a path is downward.
 
     Criterias are:
 
     - in the number of active flavors when the path list contains more than one
-      :class:`PathSegment`, note this can be different from each
-      :attr:`PathSegment.is_downward`
-    - in :math:`Q^2` when just one single :class:`PathSegment` is given
+      :class:`Segment`, note this can be different from each
+      :attr:`Segment.is_downward`
+    - in :math:`\mu^2`, when just one single :class:`Segment` is given
 
     """
     if len(path) == 1:
--- a/src/eko/quantities/heavy_quarks.py
+++ b/src/eko/quantities/heavy_quarks.py
@@ -65,19 +65,6 @@ MatchingScale = SquaredScale
 MatchingScales = HeavyQuarks[MatchingScale]
 
 
-def scales_from_ratios(
-    ratios: MatchingRatios, masses: HeavyQuarkMasses
-) -> MatchingScales:
-    """Convert ratios to squared scales.
-
-    .. todo::
-
-        make this a method
-
-    """
-    return MatchingScales(*(np.power(ratios, 2.0) * np.power(masses, 2.0)).tolist())
-
-
 # TODO: upgrade the following to StrEnum (requires py>=3.11) with that, it is
 # possible to replace all non-alias right sides with calls to enum.auto()
 
@@ -119,11 +106,6 @@ class HeavyInfo(DictLike):
     """Scheme used to specify heavy quark masses."""
     matching_ratios: MatchingRatios
     """Matching scale of heavy quark masses"""
-
-    @property
-    def matching_scales(self) -> MatchingScales:
-        """Compute matching scales."""
-        return scales_from_ratios(self.matching_ratios, self.masses)
 
     @property
     def squared_ratios(self) -> List[float]:
--- a/src/eko/quantities/heavy_quarks.py
+++ b/src/eko/quantities/heavy_quarks.py
@@ -1,7 +1,7 @@
 """Heavy quarks related quantities."""
 import enum
 from dataclasses import dataclass
-from typing import Generic, List, Sequence, TypeVar, Union
+from typing import Generic, List, Sequence, TypeVar
 
 import numpy as np
 
@@ -58,7 +58,7 @@ class HeavyQuarks(list, Generic[T]):
 
 QuarkMass = LinearScale
 QuarkMassRef = ReferenceRunning[QuarkMass]
-HeavyQuarkMasses = HeavyQuarks[Union[QuarkMass, QuarkMassRef]]
+HeavyQuarkMasses = HeavyQuarks[QuarkMassRef]
 MatchingRatio = float
 MatchingRatios = HeavyQuarks[MatchingRatio]
 MatchingScale = SquaredScale
--- a/tests/eko/io/test_inventory.py
+++ b/tests/eko/io/test_inventory.py
@@ -1,0 +1,93 @@
+from dataclasses import dataclass
+
+import numpy as np
+import pytest
+
+from eko.io import access, inventory, items
+
+
+@dataclass(frozen=True)
+class FakeH(items.Header):
+    blub: int
+
+
+def test_contentless(tmp_path):
+    acw = inventory.AccessConfigs(tmp_path, False, True)
+    iw = inventory.Inventory(tmp_path, acw, FakeH, contentless=True, name="Bla")
+    assert "Bla" in str(iw)
+    one = FakeH(1)
+    assert one not in iw
+    assert len(list(tmp_path.glob("*"))) == 0
+    # set an element
+    iw[one] = "blub"
+    assert one in iw
+    assert iw[one] is None
+    ls = list(tmp_path.glob("*"))
+    assert len(iw) == 1
+    assert len(ls) == 1
+    assert ls[0].suffix == inventory.HEADER_EXT
+    # let's make a second read-only instance
+    acr = inventory.AccessConfigs(tmp_path, True, True)
+    ir = inventory.Inventory(tmp_path, acr, FakeH, contentless=True, name="noihaha")
+    assert "Bla" not in str(ir)
+    assert one not in ir
+    assert ir[one] is None
+    # deletion is empty for contentless
+    del iw[one]
+    assert one in iw
+    assert iw[one] is None
+    assert len(list(tmp_path.glob("*"))) == 1
+    iw.empty()
+    assert len(list(tmp_path.glob("*"))) == 1
+    # non-existant
+    two = FakeH(2)
+    assert two not in iw
+    with pytest.raises(inventory.LookupError):
+        iw[two]
+
+
+def test_contentfull(tmp_path):
+    def o():
+        return items.Operator(np.random.rand(2, 2, 2, 2))
+
+    acw = inventory.AccessConfigs(tmp_path, False, True)
+    iw = inventory.Inventory(tmp_path, acw, FakeH, name="Bla")
+    one = FakeH(1)
+    assert one not in iw
+    assert len(list(tmp_path.glob("*"))) == 0
+    # set an element
+    o_one = o()
+    iw[one] = o_one
+    assert one in iw
+    assert iw[one] is not None
+    assert len(iw) == 1
+    assert len(list(tmp_path.glob("*" + inventory.HEADER_EXT))) == 1
+    assert len(list(tmp_path.glob("*" + inventory.COMPRESSED_EXT))) == 1
+    # let's make a second read-only instance
+    acr = inventory.AccessConfigs(tmp_path, True, True)
+    ir = inventory.Inventory(tmp_path, acr, FakeH, name="noihaha")
+    assert "Bla" not in str(ir)
+    assert one not in ir
+    # if we sync we know the header is there
+    ~ir
+    assert one in ir
+    assert len(ir) == 1
+    # do an actual read to load
+    np.testing.assert_allclose(ir[one].operator, o_one.operator)
+    # but we can not write
+    with pytest.raises(access.ReadOnlyOperator):
+        ir[one] = o()
+    # let's overwrite the op in the writable, so they diverge
+    o_one_p = o()
+    iw[one] = o_one_p
+    assert len(iw) == 1
+    np.testing.assert_allclose(iw.cache[one].operator, o_one_p.operator)
+    np.testing.assert_allclose(ir.cache[one].operator, o_one.operator)
+    # let's actually read again
+    ~ir
+    np.testing.assert_allclose(ir[one].operator, o_one_p.operator)
+    # non-existant
+    two = FakeH(2)
+    assert two not in ir
+    with pytest.raises(inventory.LookupError):
+        ir[two]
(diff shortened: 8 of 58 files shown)


AleCandido:
@felixhekhorn while propagating the new mu2grid I noticed a couple of things:

1. we have a mostly unused parameter of some operator operations of `EKO`, i.e. `compress` - should we assume that we always compress our operators and simplify the implementation?
2. since now we are specifying what I called "evolution points" (i.e. $(\mu^2, n_f)$, alternative proposals are welcome), maybe it would be more consistent to collapse the `nf0` parameter with `mu0`, to have a single evolution point

1. is an optional argument of some internal functions, and I do not expect it to be used anywhere, so it should be non-breaking. `nf0` is EKO specific (in the sense that is not included in the NNPDF theory), but it would break both theory card and operator card.

felixhekhorn:
1. I agree - we can assume it to be always compressed (which in practice is always the case)
2. indeed this a bit more tricky since it is another (more or less) big break ... I've no clear opinion ... maybe better to postpone this to after this PR?

felixhekhorn:
Well, now you're suggesting a *much* more drastic change than before :upside_down_face: (e.g. because I know this would also have a highly non-trivial impact on yadism) - while I agree we may want to update a bit, are you really sure you want to do this here? this PR will already be big and I think we could decouple this change ... let me comment on your points nevertheless:

> at some point I'd like to rename

even fine (as long as we keep the name "Atlas" - which I like :innocent: )

1. Mmm ... in practice when you're crossing a threshold you're always interested in the ratio as this will enter your matching conditions as an explicit parameter, so not sure about dropping it here ...
2. I guess we can (as long as tests keep passing) ...
3. either is or
4. this is fine, I guess ...
5. then we should follow what we do in practice: have a fixed point at construction time and no "second" argument in `path`
6. well, we should keep the possibility to specify `0` or `-1` as `nf_ref` - when it will be auto determined

AleCandido:
> Mmm ... in practice when you're crossing a threshold you're always interested in the ratio as this will enter your matching conditions as an explicit parameter, so not sure about dropping it here ...

It is not used at all by the atlas. Let's decouple tasks: the `Atlas` is drawing the path, if we need the ratios information for something else, we will propagate on its own.

AleCandido:
> * either is or
> 
> * this is fine, I guess ...

Among the two, I prefer 3. (or both).

> * well, we should keep the possibility to specify `0` or `-1` as `nf_ref` - when it will be auto determined

I already wrote a function for the automated determination:

https://github.com/NNPDF/eko/blob/d0f6e7a5a82b81d6b76357effc5ac7c9a066c2d1/src/eko/io/runcards.py#L388-L406

So, the method you're asking for is `Atlas.nf()`.

If ever we will do something like this, we should use `None`, that is much more explicit, but I would avoid usage in the current runcards, and keep it:
- for legacy upgrade (thus where it is)
- or as a tool for writing runcards in `ekobox`

Following the spirit of the default-less strategy, we do not want to recommend a specific number of flavors at a given scale, since they are all available.
(remark: I'd keep following just *the spirit of* the default-less strategy, and instead being much milder on the strategy application itself, since some defaults, where sensible, simplify a lot the usage, marking the options as "advanced" - but always preferring to keep the code simple, without complex parameter determinations happening behind the scenes)

AleCandido:
I updated `thresholds.py`, and it is now much cleaner!

The amount of code left is minimal, and everything should be well documented. However, it broke EKO completely.

Since I already broke it, I decided to go all the way in and change even the name (still very open to alternatives, please propose), it will make it even simpler to detect all the places in which it is used.

tests/eko/test_matchings.py
```diff
@@ -0,0 +1,141 @@
...
+        # is hashable?
+        d = {}
+        d[p] = 1
+        assert d[p] == 1
+
+    def test_str(self):
+        p = Segment(0, 1, 3)
+        s = str(p)
+        assert s.index("0") > 0
+        assert s.index("3") > 0
```
felixhekhorn, line 25:
```suggestion
        assert s.index("1") > 0
```

tests/eko/test_matchings.py
```diff
@@ -0,0 +1,141 @@
+"""
+    Tests for the threshold class
+"""
```
felixhekhorn, line 3:
```suggestion
"""Tests for the threshold class"""
```
no?

src/eko/matchings.py
```diff
@@ -0,0 +1,201 @@
...
+
+
+Path = List[Segment]
+
+
+@dataclass(frozen=True)
+class Matching:
+    """Matching between two different segments.
+
+    The meaning of the flavor index is the PID of the corresponding heavy quark.
```
felixhekhorn, line 43: This refers to the index in the *upper* patch, correct? We need to have an absolute sorting and you need to mention that here ... (think of: we evolve from 20 -> 30 then we add the bottom matching instead if we evolve 30 -> 20 we have to remove the bottom matching)

src/eko/matchings.py
```diff
@@ -0,0 +1,201 @@
...
+        """Create basic atlas."""
+        self.walls = [0] + matching_scales + [np.inf]
+        self.origin = origin
+
+        logger.info(str(self))
+
+    def __str__(self):
+        """Textual representation, mainly for logging purpose."""
+        walls = " - ".join([f"{w:.2e}" for w in self.walls])
+        return f"ThresholdsAtlas [{walls}], ref={self.origin[0]} @ {self.origin[1]}"
```
felixhekhorn, line 72:
```suggestion
        return f"Atlas [{walls}], ref={self.origin[0]} @ {self.origin[1]}"
```

felixhekhorn:
I wonder whether we should make a release before merging this one ... this is due to the current (=master) version being already propagated https://github.com/NNPDF/pineappl/pull/227 and due to the bad experiences with #172

AleCandido:
> I wonder whether we should make a release before merging this one ... this is due to the current (=master) version being already propagated [NNPDF/pineappl#227](https://github.com/NNPDF/pineappl/pull/227) and due to the bad experiences with #172

The actual merge is not going to happen immediately, so I would reconsider the situation in a while.
In case, if we are already breaking with something else, we could make a pre-release of the next minor, make also this one into a pre-release, and eventually stabilize with an actual minor when it will be proved to work.

src/eko/io/runcards.py
```diff
@@ -272,6 +274,21 @@ def update(theory: Union[RawCard, TheoryCard], operator: Union[RawCard, Operator
...
 
+def default_atlas(masses: list, matching_ratios: list):
+    r"""Create default landscape.
+
+    This method should not be used to write new runcards, but rather to have a
+    consistent default for comparison with other softwares and existing PDF
+    sets.
+    There is no one-to-one relation between number of running flavors and final
+    scales, unless matchings are all applied. But this is a custom choice,
+    since it is possible to have PDFs in different |FNS| at the same scales.
```
AleCandido, line 285: This is copy-pasted from below :P

I'd propose to drop this, and to prepend an underscore to this function.
At some point, I will also go through #175 and reconsider these objects organically.

AleCandido:
@felixhekhorn in 7f8a62864fb2f4587bfc9b3bc5a0bf8bf3766bee I tried to simplify the xgrid reshape function, since that one and the corresponding flavor one are quite complex and very redundant.

Of course, I started by fixing the corresponding test, that was failing for a different reason...

However, there is still a lot to be improved: the similarity with the flavor one is now hidden and the new is still too complex, in order to account for the factorization of the two (xgrid and flavor) into some shared pieces.

1. this function is almost trivial, but if we use over and over is better factored out:
    https://github.com/NNPDF/eko/blob/7f8a62864fb2f4587bfc9b3bc5a0bf8bf3766bee/src/eko/io/manipulate.py#L24-L42
2. I wonder if `swap` here is really needed: if what I wrote in the docstring is correct, I'd rather transpose the result, rather than adding a swap argument
    https://github.com/NNPDF/eko/blob/7f8a62864fb2f4587bfc9b3bc5a0bf8bf3766bee/src/eko/io/manipulate.py#L49-L60
3. here I would have liked to immediately assign `eko.rotations.xxxgrid` on the same line, but I postponed in order to avoid setting trivial ones when no rotations will be performed
    https://github.com/NNPDF/eko/blob/7f8a62864fb2f4587bfc9b3bc5a0bf8bf3766bee/src/eko/io/manipulate.py#L85-L107
    I wonder if there is a way to further simplify it...
4. instead about this part I'm pretty happy, and I believe that we could already factorize and apply also in `flavor_reshape`
    https://github.com/NNPDF/eko/blob/7f8a62864fb2f4587bfc9b3bc5a0bf8bf3766bee/src/eko/io/manipulate.py#L109-L136

Can you confirm I didn't break anything, and try to see if there are ways to further improve, especially the first part (about checks and rotations computation).

felixhekhorn:
All fine, more or less - but you're adding one giant break on an other ...

AleCandido:
> All fine, more or less - but you're adding one giant break on an other ...

Nope, this should not break anything :)

src/eko/io/struct.py
```diff
@@ -627,6 +627,11 @@ def mu2grid(self) -> List[SquaredScale]:
         """Provide the list of :math:`Q^2` as an array."""
         return [mu2 for mu2, _ in self._operators]
 
+    @property
+    def epgrid(self) -> List[EPoint]:
```
felixhekhorn, line 631: I think `mu2grid` should be removed in favor of this function (likewise for `mu20`), but think this is a massive change I didn't dare do it straight away ...

AleCandido: Indeed, I was worried of dropping `mu2grid` immediately...

I started calling in some places `evolgrid` (sounds better than `epgrid`, that is more a code name), but here I avoided just because you could use `list(self)` for this, leveraging the definition of `__iter__` (and I believe we should do also here, if you want to explicitly expose this property, don't you think so?).

felixhekhorn: adjusted in https://github.com/NNPDF/eko/pull/242/commits/e91e526183efad77be23790b7b085fb34f52316d - okay, but if you agree then I start to propagate it around

src/eko/io/inventory.py
```diff
@@ -0,0 +1,138 @@
...
+    """Determine operator file name, from the associated header."""
+    stem = encode(header)
+    return stem + OPERATOR_EXT[1 if err else 0]
+
+
+@dataclass(frozen=True)
+class Inventory:
+    """Assets manager.
+
+    In particular, manaage autosave, autoload, and memory caching.
```
felixhekhorn, line 46:
```suggestion
    In particular, manage autosave, autoload, and memory caching.
```

AleCandido:
I finally split `struct.py` in several modules, I started hating scrolling up and down so much...

There will be a few tests to be fixed :)

AleCandido:
@felixhekhorn I will try to maintain a TODO list in the OP, feel free to add some items you believe to be relevant.

If you wish, restoring the old runner and fixing some tests is open for help. At the moment, I will focus on completing the `managed` runner.

`eko.evolution_operator` can be greatly improved, including some functions of `PhysicalOperator` and much involved code. However, I decided to focus on `OperatorGrid` alone in this PR, to have the alternative runner and nothing else.
The improvement of the individual pieces of the computation will come in a further PR (and this gives a sense to the *raw* in `raw-jets`).

src/eko/matchings.py
AleCandido: As mentioned in person: `hq` is a `FlavorIndex`, not a `FlavorsNumber`, so it refers to the PID of the heavy quark associated to the matching, e.g. 5 for bottom.

It is not the lower, nor the upper patch. Of course, numerically it coincides with the `FlavorsNumber` in the upper patch, but this is a consequence, conceptually they are different objects.

AleCandido:
@felixhekhorn fixed the last big bug in the new runner in d46d1b3, but we still have a tiny one left that is mostly affecting T24 (so the bottom), and only in some cases.
It seems to me that also the other distributions are slightly deteriorated, but it is a <1% effect (and most likely a consequence of the former).

So, @giacomomagni and @niclaurenti, you could start having a look at the code, since in most of the cases is already perfectly working.
In particular, check if you find something that seems to break compatibility with the current `master` (i.e. if you call the same `eko.runner.solve`/`eko.solve` function with the same runcards, it might crash the program).

AleCandido:
Sorry for the bad news, but I just realized there is another reason for which we can't use yet the new runner (beyond the "T24 bug", that we'll hunt with @felixhekhorn): PineAPPL compatibility.

At the moment, @cschwan maintained an EKO loader in PineAPPL for the purpose, but here I'm breaking again the internal structure, so it won't work any longer.
Thus, `grid.evolve()` won't work any longer, breaking completely Pineko.

The goal of this PR is to put the new runner side-by-side to the old one, and keep the old as the default for another release.
From everyone else's point of view it should be a non-breaking operation, but PineAPPL relies on EKO internals...

So, we should definitely tag `master` before merging this, if we have a working version with everything else. And after that (git-wise, chronologically is irrelevant) we can merge this PR into `master`.
But before tagging the content of this PR, I will try to implement #97, at least a first step. I.e. importing the Rust EKO loader from PineAPPL, and update it to match with the current structure (just loading metadata and operators, optionally also runcards).

cschwan:
@AleCandido can you explain how this would affect the implementation of [`pineappl evolve`](https://github.com/NNPDF/pineappl/blob/ee84ca62765d5087234cb054bce1637b038af8ce/pineappl_cli/src/evolve.rs#L10-L178) in terms of how the files would change? In principle I don't mind you changing the format again as long as I can support it inside `pineappl evolve`.

Or do you really mean `Grid::evolve`, for which this is probably already foreseen in <https://github.com/NNPDF/pineappl/pull/186>?

AleCandido:
> @AleCandido can you explain how this would affect the implementation of [`pineappl evolve`](https://github.com/NNPDF/pineappl/blob/ee84ca62765d5087234cb054bce1637b038af8ce/pineappl_cli/src/evolve.rs#L10-L178) in terms of how the files would change? In principle I don't mind you changing the format again as long as I can support it inside `pineappl evolve`.
> 
> Or do you really mean `Grid::evolve`, for which this is probably already foreseen in [NNPDF/pineappl#186](https://github.com/NNPDF/pineappl/pull/186)?

NNPDF/pineappl#186 is fine with the current `master` version, but this one is drastically changing the structure of the internal `operators/` folder.
So, yes, I really meant `pineappl evolve`, sorry (but also the idea of giving a lazy loader to `grid.evolve()`, in order to avoid the need of loading a full EKO all at once, it would be helpful to evolve jets).

Before, we already had the `recipes/` and `parts/` folders, but they were unused. Their role was actually to support the feature implemented in this PR, but, lacking an implementation, their designed structure was sketchy.
Since here I actually had to use them, I finalized the design, and propagated to `operators` as well, to have a uniform layout (such that I can manage with a single object).

In particular, at the moment the file names were determined by a complex prescription, essentially embedding the structure labeling the operator in a reversible way (lossless).
This was initially motivated since operators were labeled just by single floats (the scales), and less motivated when it became a pair `(float, int)` (the scale and the number of flavors).
For `recipes` and `parts` this would have been even more complex, so I decided to change strategy for a more scalable one.

I named the structure associated to each of those folders an `Inventory`, and each inventory manages `Item`s, each of which is made of a `Header` and a `Content`. The content is always a rank-4 array (possibly two, when it includes the integration error), as it was before for `operators/`, if present (`contentless` inventories are allowed, since it was needed for `recipes/`).
The associated recipe is whatever `struct`-like object (a product type), possibly made of basic fields (null, booleans, floats, integers, and strings).

The serialization of an `Item` is a pair of files (in the case of contentful inventories, like `operators/`), one containing the header, serialized in YAML, and the other containing the NumPy dump (as before).
The file names are `<hash>.yaml` and `<hash>.np{y,z}.lz4` (as before, `.npy` if error is not present, else `.npz`), where `<hash>` is the hash of the Python data structure.

I already see a limitation of this approach and a line of improvement: instead of computing the native Python hash over the data structure, I should have computed a standard hash function (like `sha1`, or something similar) on the YAML serialization, since it would have been more predictable.
However, a reader doesn't need to compute any hash, since it the only operations required are: 
- load an operator, given a file name
- load all headers (i.e. all the `yaml` files in the folder), keeping the header-name relation in a map
- given a header (`(mu2, nf)` for operators), retrieve a file name and load the corresponding content

Given that `pineappl evolve` just requires iterating over the whole content, you don't even need the last point (what in Python would be called `__getitem__`).

src/eko/couplings.py
```diff
@@ -446,24 +446,19 @@ def assert_positive(name, var):
...
         self.a_ref = np.array(couplings.values) / 4.0 / np.pi  # convert to a_s and a_em
-        self.thresholds = thresholds.ThresholdsAtlas(
-            masses,
-            couplings.scale**2,
-            nf_ref,
-            thresholds_ratios=thresholds_ratios,
-            max_nf=max_nf,
-        )
+        matching_scales = (np.array(masses) * np.array(thresholds_ratios)).tolist()
+        self.thresholds_ratios = list(thresholds_ratios)
```
giacomomagni, line 456: `thresholds_ratios` maybe is already a list given the line above

giacomomagni (commented):
Hi, 
I tried running the benchmark and it seems to work smoothly and faster!! :))

Just a few minor comments: 
* ´src/eko´ still contains some `q2` in names or docstrings
 you might want to update.

* here there is a leftover `rotations`: 
https://github.com/NNPDF/eko/blob/4ff16c0120e1240300e2c6e51a3d3c960fd4e41c/src/ekomark/plots.py#L231


* I gave a try in testing the eko converter, but this is what get: 

```
(eko_dev) giacomomagni@Giacomos-MacBook-Pro eko % eko convert ../../ICpheno/charm_asym/ekos/o24c2cf_t6011e3.tar 
[09:49:17] INFO     Atlas [0.00e+00 - 2.28e+00 - 2.42e+01 - 2.98e+04 - inf], ref=0.0 @ 0                                                                                                                              matchings.py:68
           INFO     Atlas [0.00e+00 - 2.28e+00 - 2.42e+01 - 2.98e+04 - inf], ref=0.0 @ 0                                                                                                                              matchings.py:68
/Volumes/Git_Workspace/physicstools/NN3PDF/eko/src/eko/io/legacy.py:180: UserWarning: Unrecognized array loaded, 'alphas'
  warnings.warn(f"Unrecognized array loaded, '{name}'")
  File "/Users/giacomomagni/.conda/envs/eko_dev/bin/eko", line 6, in <module>
    sys.exit(command())
  File "/Users/giacomomagni/.conda/envs/eko_dev/lib/python3.10/site-packages/click/core.py", line 1130, in __call__
    return self.main(*args, **kwargs)
  File "/Users/giacomomagni/.conda/envs/eko_dev/lib/python3.10/site-packages/click/core.py", line 1055, in main
    rv = self.invoke(ctx)
  File "/Users/giacomomagni/.conda/envs/eko_dev/lib/python3.10/site-packages/click/core.py", line 1657, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/giacomomagni/.conda/envs/eko_dev/lib/python3.10/site-packages/click/core.py", line 1404, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/giacomomagni/.conda/envs/eko_dev/lib/python3.10/site-packages/click/core.py", line 760, in invoke
    return __callback(*args, **kwargs)
  File "/Volumes/Git_Workspace/physicstools/NN3PDF/eko/src/ekobox/cli/convert.py", line 24, in subcommand
    legacy.load_tar(old, new)
  File "/Volumes/Git_Workspace/physicstools/NN3PDF/eko/src/eko/io/legacy.py", line 65, in load_tar
    with EKO.create(dest) as builder:
  File "/Volumes/Git_Workspace/physicstools/NN3PDF/eko/src/eko/io/struct.py", line 587, in __exit__
    eko.close()
  File "/Volumes/Git_Workspace/physicstools/NN3PDF/eko/src/eko/io/struct.py", line 471, in close
    self.dump()
  File "/Volumes/Git_Workspace/physicstools/NN3PDF/eko/src/eko/io/struct.py", line 458, in dump
    with tarfile.open(archive, "w") as tar:
  File "/Users/giacomomagni/.conda/envs/eko_dev/lib/python3.10/tarfile.py", line 1662, in open
    return cls.taropen(name, mode, fileobj, **kwargs)
  File "/Users/giacomomagni/.conda/envs/eko_dev/lib/python3.10/tarfile.py", line 1672, in taropen
    return cls(name, mode, fileobj, **kwargs)
  File "/Users/giacomomagni/.conda/envs/eko_dev/lib/python3.10/tarfile.py", line 1484, in __init__
    fileobj = bltn_open(name, self._mode)
FileNotFoundError: [Errno 2] No such file or directory: '../../ICpheno/charm_asym/ekos/../../ICpheno/charm_asym/ekos/o24c2cf_t6011e3-new.tar'
```

I'm not 100% sure this is related to this PR, but I wonder if this feature can ever work and you are just not maintaining a dead code.

src/eko/couplings.py
AleCandido: Not really: since somewhere else it was a tuple or an array (don't remember which and where), I downgraded the input to `Iterable[float]`.

This `list` is a temporary compatibility step, that won't be any longer required as soon as we will have enforced clearer typing in `eko`.

AleCandido:
> I tried running the benchmark and it seems to work smoothly and faster!! :))

Faster is unexpected, it should've been slightly slower...

The reason is that is doing the exact same operations, but in the middle it is dumping and loading arrays on disk, to limit memory usage (but dumping and loading have a cost).

The only true difference is that I wanted to limit the complexity of objects on disk, so I decided to restrict to just arrays, including for `parts/`, that is the elements that are multiplied in VFNS (evolution patches and matchings).
Because of this, we're now multiplying only pure rank-4 arrays, while before we were multiplying sparser arrays with the "dot notation" (i.e. `"S.g" * "g.g" = "S.g"`), but I don't know if this were faster or slower (there were 16 elements for pure QCD, and 30 for QCDxQED, while now we are using 196 elements always, but instead of using strings and dictionaries to implement the multiplication, now it is a pure linear algebra product implemented in NumPy).

I'm already glad that is working, if it's faster is an unexpected, but welcome, feature :)

AleCandido:
> * I gave a try in testing the eko converter, but this is what get:

Unfortunately, the converter has not being upgraded at all. But I was considering dropping completely the support for the legacy format, to simplify the code (not for legacy runcards, those should be still supported).

Do you believe there will be still need for it?

In principle, we could more easily convert v0.12 EKOs in this new EKOs. In case there is the need for a converter, would this be enough?

giacomomagni:
>
> Do you believe there will be still need for it?
> 

The main (and hopefully unique) reason why I'm intersted in coverting old old ekos, it's to access the old not updated N3LO charm in 3FNS (and impossible to reproduce, apart for fishing a 2 years ago commit...), since they were refrained in updating them. Now it's a while that no one is asking for it, so I'm tempted to say let's drop this, but I fear someone can pop up asking for a plot using that stuff...

src/eko/evolution_operator/grid.py
```diff
@@ -39,40 +46,20 @@ class OperatorGrid(sv.ModeMixin):
...
         configs: Configs,
         debug: Debug,
-        thresholds_config,
-        couplings,
-        interpol_dispatcher,
+        atlas: Atlas,
+        couplings: Couplings,
+        interpol_dispatcher: InterpolatorDispatcher,
     ):
-        """Initialize `OperatorGrid`.
```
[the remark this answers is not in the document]
AleCandido: In principle both, but the whole content of this file is more or less deprecated, since it is the *old runner*, and it is not used any longer by the *new* runner.

I would skip the docstrings update, just because they were outdated even before this PR, and in principle it is not this PR business to fix them (but to supersede). However, if you'd like to do it, feeel free.

AleCandido:
From https://github.com/NNPDF/eko/pull/242#pullrequestreview-1416260154

- [ ] `src/eko` still contains some q2 in names or docstrings, you might want to update
- [ ] here there is a leftover `rotations`: 
    https://github.com/NNPDF/eko/blob/4ff16c0120e1240300e2c6e51a3d3c960fd4e41c/src/ekomark/plots.py#L231

src/eko/io/inventory.py
```diff
@@ -102,10 +102,9 @@ def __getitem__(self, header: H) -> Optional[Operator]:
 
         # in case of contentless, check header availability instead
         if self.contentless:
-            if self.lookup(stem, header=True):
-                self.cache[header] = None
-                return None
-            raise LookupError(f"Impossible to retrieve {header} from {self}.")
+            self.lookup(stem, header=True)
+            self.cache[header] = None
+            return None
```
AleCandido, line 107: I would revert this, since they are two different situations:
1. you request a `header` that is present, but not loaded in `cache`, thus you put in the `cache`
2. you request a `header` that is not present, consequently you should receive an error

With your change, you can add elements to the cache by requesting them, and they would never be associated to items on disk, so you would lose them after the run.

src/eko/io/inventory.py
```diff
@@ -0,0 +1,211 @@
...
+            path
+            for path in self.path.iterdir()
+            if path.name.startswith(stem)
+            if "".join(path.suffixes) in EXT
+        ]
+
+        if len(found) == 0:
+            raise LookupError(f"Item '{stem}' not available in {self}.")
+        elif len(found) > 1:
+            raise LookupError(
```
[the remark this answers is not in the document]
AleCandido: I'm only checking if `.startswith(stem)`, where `stem` is potentially whatever `str`ing, so I can't guarantee there is exactly one.

I'm mainly worried that I'm just returning the first element, so the user would never know he asked for un ambiguous `stem`.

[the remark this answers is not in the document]
felixhekhorn: completely faked test added in https://github.com/NNPDF/eko/pull/242/commits/33a6e5d49ee925ffd4e7d6901ede3890f7e0abdf

src/eko/io/inventory.py
AleCandido: Ah, you're right. I just forgot `lookup` was raising :P

src/eko/quantities/heavy_quarks.py
```diff
@@ -65,19 +65,6 @@ def t(self, value: T):
...
-    ratios: MatchingRatios, masses: HeavyQuarkMasses
-) -> MatchingScales:
-    """Convert ratios to squared scales.
-
-    .. todo::
-
-        make this a method
-
-    """
-    return MatchingScales(*(np.power(ratios, 2.0) * np.power(masses, 2.0)).tolist())
```
felixhekhorn, line 78: @AleCandido you should remove that ambiguity here https://github.com/NNPDF/eko/blob/ec083e7ed167554107defd3723c50008046e596a/src/eko/quantities/heavy_quarks.py#L61 which was make this function failing - and since it is not actually used anywhere I decided to to the simplest thing and remove it all together

src/eko/quantities/heavy_quarks.py
```diff
@@ -65,19 +65,6 @@ def t(self, value: T):
 MatchingScales = HeavyQuarks[MatchingScale]
 
 
-def scales_from_ratios(
```
AleCandido, line 68: Why don't you like this?

src/eko/quantities/heavy_quarks.py
```diff
@@ -120,11 +107,6 @@ class HeavyInfo(DictLike):
     matching_ratios: MatchingRatios
     """Matching scale of heavy quark masses"""
 
-    @property
-    def matching_scales(self) -> MatchingScales:
```
AleCandido, line 119: and this property?

src/eko/quantities/heavy_quarks.py
AleCandido: Ah, I see. But that ambiguity I need for parsing, since we might have pole masses or MSbar ones.

The alternative is to make them always `QuarkMassRef`, slightly increasing the runcards complexity, but with the benefit of having a single type around (the reference for pole masses can be `nan`)

AleCandido: Ok, so we can drop the `Union`. If we were able to completely avoid `Union`s (apart from `Optional`, that's a different story) my life would be much easier :)

src/eko/io/items.py
```diff
@@ -0,0 +1,204 @@
...
+        content = np.load(extracted_stream)
+
+        if isinstance(content, np.ndarray):
+            op = content
+            err = None
+        elif isinstance(content, npyio.NpzFile):
+            op = content["operator"]
+            err = content["error"]
+        else:
+            # TODO: We might consider dropping this exception since np.load will always
```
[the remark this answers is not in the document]
AleCandido: The other option is to drop the one of the two `if`s conditions, and replace them with a plain `else`.
However, I find it more readable this way, since the user knows what's happening just reading the code, without the need of any further tool.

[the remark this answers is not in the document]
felixhekhorn: Fine - but this triggers a completely  fake test: https://github.com/NNPDF/eko/blob/ef1289d8abbc8f9e7daafab28e8bca050343226c/tests/eko/io/test_items.py#L77

AleCandido:
I rebased on `master` after the merge of #258, i.e. py3.11, and it seems that also the tests in this PR are passing even on that Python version.

I believe that we should be good to go. We're not in a special hurry to merge this PR, but we want it a bit sooner, rather than later, because we'd like to use to compute jets, and to finally close a big PR, reducing the level of branching in EKO (this is the last big PR left in EKO).

felixhekhorn:
> this is the last big PR left in EKO

I agree on everything, but on that one :innocent:

felixhekhorn:
> this is the last big PR left in EKO

I agree on everything, but on that one :innocent:

tests/eko/io/test_metadata.py
```diff
@@ -0,0 +1,33 @@
+import logging
+
+import pytest
+import yaml
+
+from eko.io import bases, metadata, paths
+
+
+def test_metadata(tmp_path, caplog):
```
felixhekhorn, line 9: @andreab1997 I guess one should add here a test to check that rotation are actually persistent ... (can't find the issue at the moment)

felixhekhorn: if you can, please (we need to get people involved :upside_down_face: )

tests/eko/io/test_inventory.py
```diff
@@ -0,0 +1,104 @@
...
+    ~ir
+    np.testing.assert_allclose(ir[one].operator, o_one_p.operator)
+    # non-existant
+    two = FakeH(2)
+    assert two not in ir
+    with pytest.raises(inventory.LookupError):
+        ir[two]
+
+
+def test_fs(tmp_path):
```
felixhekhorn, line 96: @AleCandido actually I can come up with a much more realistic test case which indeed reveals a bug in the current implementation, i.e. https://github.com/NNPDF/eko/pull/242/commits/605798cda32ff6c0ba9b6f05471b9dbd37be688f will fail with
```
E           eko.io.inventory.LookupError: Too many items associated to 'pJnoDyjlzDs=' in Inventory 'operators':
E           [PosixPath('/tmp/eko-jayn2gj1/operators/pJnoDyjlzDs=.npz.lz4'), PosixPath('/tmp/eko-jayn2gj1/operators/pJnoDyjlzDs=.npy.lz4')]
```
i.e. one contains errors and is thus `.npz.lz4` and the other (created before) does not and is thus `.npy.lz4` - so adding or deleting errors on an operator triggers a problem

AleCandido: `Operator`s are supposed to be immutable, you should not modify existing ones.

AleCandido: I.e. if you do, it is at your own risk.

`manipulate` does it consistently, if you want a more generic change, the advice is to fill a new `EKO`, or you have to know how to update it consistently.

felixhekhorn: fine

src/eko/io/struct.py
```diff
@@ -587,126 +170,58 @@ def permissions(self):
...
+        """Drop operator from memory."""
+        del self.operators[Target.from_ep(ep)]
+
+    def __delattr__(self, name: str):
+        """Empty an inventory cache."""
+        attr = getattr(self, name)
+        if isinstance(attr, Inventory):
+            attr.empty()
+        else:
+            super().__delattr__(name)
```
felixhekhorn, line 205: @AleCandido what is the use case of this line? why would I delete an eko attribute? and can I even do it on an ordinary Python object? I think this should be replaced by a `ValueError` or similar ...

AleCandido: Yes, you can do it on an ordinary Python object. In particular, on whatever Python object.
After all, Python is a dynamic language, otherwise it would be C.

Indeed, it is just the counterpart of setting new attributes: as you can set an attribute to an instance, you can also delete it.
And setting new attributes only to `self` and only inside `__init__` is just a convention and nothing else. `self` is not even a keyword, and you could choose whatever name you prefer inside `__init__`. And `__init__` would be a perfectly normal function, apart that is called right after the creation of a new instance, with the arguments provided to the constructor.

AleCandido: So, the `else` branch is there just to avoid modification of the default behavior of Python objects, apart from the specific case in which I want to use `del` with a different meaning.

In any case, this syntax:
```py
del eko.operators
```
it is only syntactic sugar for:
```py
eko.operators.empty()
```

I would not worry too much about what `del` does in general (you're just surprised because you're encountering a new Python operator for the first time).

P.S.: btw `del x.y` is `__delattr__` and `del x[y]` is `__delitem__`, and they are quite different, despite using the same keyword: the first is defined for all objects, while the second is only used by some containers (`list`, `dict`, and other mutable containers)

felixhekhorn: fine

src/eko/io/items.py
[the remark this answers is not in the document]
AleCandido: Only together with the full-coverage-at-all-costs requirement :)

giacomomagni (approved):

giacomomagni (approved):

src/eko/io/items.py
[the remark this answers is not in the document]
felixhekhorn: fine ... recall that [I just *did* find another bug](https://github.com/NNPDF/eko/commit/b94e4a21bdc97422505a323fc491ffdefdba51a7) doing this and it was exactly in the one line missing ...

AleCandido:
`Atlas.normalize` is only affecting the computation of couplings and MSbar masses.

The main issue is simply that we didn't move consistently to `EvolutionPoint`s in the full `eko` package, so there are still some functions that make use of bare target scales (like in couplings), and because of this we're allowing `nf=None`, and this propagates down to the `Atlas`.

The part involved is not going to be immediately part of the public API, and the only places where this is directly used are `ekobox` (which we can consistently update in a single PR, living in this repo) and `yadism`, that is already lacking behind (and using an old version of EKO).

I will open a dedicated issue, and the upgrade will be done in future PRs, while refactoring EKO.

src/eko/evolution_operator/grid.py
```diff
@@ -94,17 +82,6 @@ def __init__(
...
-            "iterate-exact",
-            "iterate-expanded",
-            "truncated",
-            "ordered-truncated",
-            "decompose-exact",
-            "decompose-expanded",
-            "perturbative-exact",
-            "perturbative-expanded",
-        ]:
-            raise ValueError(f"Unknown evolution mode {method}")
```
niclaurenti, line 107: why this check has been removed?

AleCandido: Full explanation in the commit message: https://github.com/NNPDF/eko/commit/93e662d5c311377df6ec01f8cdd06b1855b91101

PR #245, opened 2023-04-10
Time like matching conditions

t7phy:


felixhekhorn:
Please keep the PRs in draft mode if they are not ready yet - note that we now have an trigger on becoming a full PR (the LHA benchmarks), so (unlike before) now the state really matters

src/ekore/operator_matrix_elements/unpolarized/time_like/as1.py
```diff
@@ -0,0 +1,77 @@
...
+
+# from eko.constants import CF
+CF = 4 / 3
+
+
+@nb.njit(cache=True)
+def A_hg(N, L):
+    r"""Compute the |NLO| heavy-gluon |OME|.
+
+    Implements :eqref:`27` from :cite:`Cacciari:2005ry`
```
felixhekhorn, line 14:
```suggestion
    Implements :eqref:`27` from :cite:`Cacciari:2005ry`.
```

t7phy:
> Please keep the PRs in draft mode if they are not ready yet - note that we now have an trigger on becoming a full PR (the LHA benchmarks), so (unlike before) now the state really matters

As I mentioned in my message, the PR is ready as is, for your review.

src/ekore/operator_matrix_elements/unpolarized/time_like/as1.py
```diff
@@ -0,0 +1,77 @@
+"""The unpolarized, time-like |NLO| matching conditions."""
+
+import numba as nb
+import numpy as np
+
+# from eko.constants import CF
+CF = 4 / 3
```
felixhekhorn, line 7: mh? why?

tests/ekore/operator_matrix_elements/unpolarized/time_like/test_as1.py
felixhekhorn, line 1: Can you add a comment somewhere á là "coming from Mela" if that is true?

src/ekore/operator_matrix_elements/unpolarized/time_like/as1.py
felixhekhorn, line 1: As mentioned, I believe we need https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/src/ekore/operator_matrix_elements/unpolarized/space_like/as1.py#L102 here - you can copy as you did for the ad or just import ...

src/ekore/operator_matrix_elements/unpolarized/time_like/__init__.py
```diff
@@ -1,15 +1,60 @@
...
+    -------
+    A_non_singlet : numpy.ndarray
+        non-singlet |OME|
+
+    """
+    if matching_order[0] > 1:
+        raise Exception("Time-like matching conditions are only known upto NLO")
+    else:
+        A_ns = np.zeros((matching_order[0], 2, 2), np.complex_)
+        A_ns[0] = as1.A_ns()
```
felixhekhorn, line 32:
```suggestion
    A_ns = np.zeros((matching_order[0], 2, 2), np.complex_)
    A_ns[0] = as1.A_ns()
```
you can simplify a bit (same below)

felixhekhorn (changes requested):

tests/ekore/operator_matrix_elements/unpolarized/time_like/test_as1.py
t7phy: done

src/ekore/operator_matrix_elements/unpolarized/time_like/as1.py
t7phy: I was testing something and forgot to change it back, corrected now

src/ekore/operator_matrix_elements/unpolarized/time_like/as1.py
```diff
@@ -0,0 +1,98 @@
...
+        )
+    )
+    return result
+
+
+@nb.njit(cache=True)
+def A_gg(L):
+    r"""Compute the |NLO| gluon-gluon |OME|.
+
+    They are defined as the Mellin transform of :eqref:`B.6` from :cite:`Buza_1998`.
```
felixhekhorn, line 46: If you copy, please
1. put the Cacciari equation as reference and say it is the same as space-like
2. add a unit test to check they are the same (it is trivial, I know)

src/ekore/operator_matrix_elements/unpolarized/time_like/__init__.py
```diff
@@ -1,15 +1,58 @@
...
+        :math:`\ln(\mu_F^2 / m_h^2)`
+
+    Returns
+    -------
+    A_non_singlet : numpy.ndarray
+        non-singlet |OME|
+
+    """
+    if matching_order[0] > 1:
+        raise Exception("Time-like matching conditions are only known upto NLO")
```
giacomomagni, line 29: Maybe it's better to have a warning here: imagine you want to do the best we have, then you do NNLO evolution + NLO matching, but this will prevent to do it...

src/ekore/operator_matrix_elements/unpolarized/time_like/as1.py
```diff
@@ -0,0 +1,101 @@
...
+    return result
+
+
+@nb.njit(cache=True)
+def A_gg(L):
+    r"""Compute the |NLO| gluon-gluon |OME|.
+
+    Implements the Mellin transform of
+    :eqref:`24` from :cite:`Cacciari:2005ry`.
+    It is identical to the one in
```
giacomomagni, line 48: why don't we import from there directly then?

tests/ekore/operator_matrix_elements/unpolarized/time_like/test_init.py
giacomomagni, line 1: here you might test also the NNLO errors for completeness

src/ekore/operator_matrix_elements/unpolarized/time_like/__init__.py
t7phy: So for this, I have not changed it to either of the two things we discussed that day but rather I have used a method which was used in MAP fragmentation functions, as suggested by Emanuele. Basically now the as1 matching conditions are used for both NLO and NNLO ADs.

src/ekore/operator_matrix_elements/unpolarized/time_like/as1.py
t7phy: This is to be consistent with the approach we also followed in ADs whereby at LO where, there were two functions which were identical but we didn't import them. Alessandro and Felix were okay with these since these were small duplications and my rational for this was as follows: 1. it keeps the time_like part self contained and 2. it allows citing the proper time-like papers which also contain these expressions.

src/ekore/operator_matrix_elements/unpolarized/time_like/__init__.py
giacomomagni: yes indeed this is the best you can do

giacomomagni (approved):

felixhekhorn (approved):

Issue #145, opened 2022-08-31
Public Interface


AleCandido:
We have to decide the public interface of EKO, for the following reason: at the moment, everything is potentially public.
In this way, we won't ever have a v1.0.0, since it is impossible that all internal details will be forever stable.

We also have to decide how to discriminate the public interface from the non-public one. Essentially, I know two ways:
1. [Python underscore convention](https://peps.python.org/pep-0008/#descriptive-naming-styles): if a name is led by an underscore, e.g. `_myfunc`, then it is private
2. [Numpy export style](https://github.com/numpy/numpy/blob/917335c6f1277e44c998c8475173ba66b4d409b5/numpy/__init__.py#L137-L155): everything and only what is accessible in the top-level scope is public, i.e. 
    ```py
    import numpy as np

    np.myobj
    ```
    then `myobj` is public

The point about majors is preserving backward compatibility, if we only add stuffs to the public API we won't need a new major ever.
So new major are needed only for refactoring the public API. Otherwise, EKO will be v1.x.y forever :)

felixhekhorn:
Let's do 2) first and then, if needed, afterwards 1)

I'll edit the top entry to add a list of needed entries

felixhekhorn:
We also have to agree on a strategy for `ekobox` and `ekomark`

AleCandido:
> We also have to agree on a strategy for `ekobox` and `ekomark`

That's fair, but less urgent. `ekomark` in particular is really a byproduct, mainly for internal use (while `ekobox` is somewhat more relevant, getting closer to the user).

felixhekhorn:
I believe [pandas is following](https://github.com/pandas-dev/pandas/blob/01693d64fe7b4335c327d62c5a5861f07c98f8c9/pandas/__init__.py#L25) our current best proposal and it provides the two functions we need ([here](https://github.com/pandas-dev/pandas/blob/01693d64fe7b4335c327d62c5a5861f07c98f8c9/scripts/validate_unwanted_patterns.py#L156) and [here](https://github.com/pandas-dev/pandas/blob/01693d64fe7b4335c327d62c5a5861f07c98f8c9/scripts/validate_unwanted_patterns.py#L205)) together with the [pre-commit hook](https://github.com/pandas-dev/pandas/blob/01693d64fe7b4335c327d62c5a5861f07c98f8c9/.pre-commit-config.yaml#L333)

AleCandido:
First, we definitely need to prioritize this issue.

[...]

PR #247, opened 2023-04-20
Fix backward matching

niclaurenti:
Pull request made to fix the bug in the backward matching:
in the function `build_ome` the argument `backward_method` arrives as an object of the type `InversionMethod.EXACT` or `InversionMethod.EXPANDED` but we are checking it to be equal to strings. In conclusion the backward matching is never applied (the matching operator is always the upwards one) unless we add the `.value`.


src/eko/evolution_operator/operator_matrix_element.py
```diff
@@ -49,7 +49,7 @@ def build_ome(A, matching_order, a_s, backward_method):
     # .end
     ome = np.eye(len(A[0]), dtype=np.complex_)
     A = np.ascontiguousarray(A)
-    if backward_method == "expanded":
+    if backward_method.value == "expanded":
```
AleCandido, line 52: The proper way is to compare with the actual object:

```suggestion
    if backward_method is InversionMethod.EXPANDED:
```

felixhekhorn:
- as conjectured: this most likely is another side effect of #172 (since the thing was introduced there)
- can you please fix the tests? (it seems they are not using what is used in practice)
- I wonder whether to fix in the opposite way, i.e. fixing the RHS instead of LHS: `if backward_method.value == InversionMethod.EXACT` (constant literals should be available inside nuba, see https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/src/eko/evolution_operator/__init__.py#L409)
- minor comment (not to be fixed in this PR): we should kill strings from numba since they are slowing down
- I immediately triggered the LHA benchmark https://github.com/NNPDF/eko/actions/runs/4755747561

felixhekhorn:
Also should we add a forward-backward run to the iso-bench? actually, can we bring https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/benchmarks/eko/benchmark_evol_to_unity.py#L109-L112 back to live?

felixhekhorn:
> Also should we add a forward-backward run to the iso-bench? actually, can we bring
> 
> https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/benchmarks/eko/benchmark_evol_to_unity.py#L109-L112
> back to live?

maybe the problem was time ... @giacomomagni do you remember?

AleCandido:
@felixhekhorn should we merge the quick fix, and worry about tests at a later stage?

niclaurenti:
> @felixhekhorn should we merge the quick fix, and worry about tests at a later stage?

Before merging I would like to perform the benchmark of evolving to 1.65 and then back to 100 to see whether it is fixed or not

giacomomagni:
> > Also should we add a forward-backward run to the iso-bench? actually, can we bring
> > https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/benchmarks/eko/benchmark_evol_to_unity.py#L109-L112
> > 
> > back to live?
> 
> maybe the problem was time ... @giacomomagni do you remember?

If I recall correctly problems where both time but also accuracy, because you had to test against and identity matrix, which was not really good.

giacomomagni (commented):
Some minor comments.
Could you please update also here:

https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/src/eko/evolution_operator/operator_matrix_element.py#L34

https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/src/eko/evolution_operator/operator_matrix_element.py#L117

and maybe we should use `None` instead of `""` here and similar:

https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/src/eko/evolution_operator/operator_matrix_element.py#L143

Otherwise also adding `.value`  in config would be more consistent with the other modules of Eko. Simply changing here:

https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/src/eko/evolution_operator/grid.py#L86

AleCandido:
> Otherwise also adding `.value` in config would be more consistent with the other modules of Eko. Simply changing here:

`config` is part of the issue: if it hadn't passed through an opaque dictionary, we might have been warned of the type inconsistency by some static analyzer (I will soon add `mypy` explicitly, but PyRight in the editor should have already noticed...). I would keep propagating the more specific types, in order to avoid similar issues later on.

niclaurenti:
Hi @giacomomagni

> and maybe we should use `None` instead of `""` here and similar:

I agree with using `None` instead of `""`

> Otherwise also adding `.value` in config would be more consistent with the other modules of Eko. Simply changing here:

but I would pass the whole object to `config` and not just the string

tests/eko/evolution_operator/test_ome.py
```diff
@@ -41,7 +41,7 @@ def test_build_ome_as():
         aS = A_singlet((o, 0), N, sx_singlet, nf, L, is_msbar, sx_ns)
 
         for a in [aNS, aS]:
-            for method in ["", "expanded", "exact"]:
+            for method in ["", InversionMethod.EXPANDED, InversionMethod.EXACT]:
```
AleCandido, line 44: Even the iteration just works:
```suggestion
            for method in InversionMethod:
```

AleCandido:
Actually, thinking about the alternatives `""` and `None` for `backward_inversion`, I believe we can avoid both: let's always have an inversion strategy, this will avoid us to always consider the case in which it is not there.

If we need a default we could pick one of the two (in case we really want a default, and we really want to avoid one of them, we can add a `NULL` variant to `InversionMethod`, but I'd prefer to avoid).

giacomomagni:
> Actually, thinking about the alternatives `""` and `None` for `backward_inversion`, I believe we can avoid both: let's always have an inversion strategy, this will avoid us to always consider the case in which it is not there.
> 
I think the `None` option was done in other to avoid passing an additional variable `is_backward` to `quad_ker` or similar.
But for sure there are other options.

felixhekhorn:
> Actually, thinking about the alternatives `""` and `None` for `backward_inversion`, I believe we can avoid both: let's always have an inversion strategy, this will avoid us to always consider the case in which it is not there.
> 
> If we need a default we could pick one of the two (in case we really want a default, and we really want to avoid one of them, we can add a `NULL` variant to `InversionMethod`, but I'd prefer to avoid).
> 
> @felixhekhorn

as already said - most likely we can not stick to our no-default-strategy in the light of #145 ... so let's assume something

> I think the `None` option was done in other to avoid passing an additional variable `is_backward` to `quad_ker` or similar. But for sure there are other options.

indeed, now as you say I remember:

https://github.com/NNPDF/eko/blob/9538753139cfcbafc81f1851449deab4fe59f9b7/src/eko/evolution_operator/operator_matrix_element.py#L52-L70

AleCandido:
> > I think the `None` option was done in other to avoid passing an additional variable `is_backward` to `quad_ker` or similar. But for sure there are other options.
> 
> indeed, now as you say I remember:

But then we are already passing to Numba something that is either `expanded` or "everything else".
Maybe we can simplify even `backward_method` by using a boolean `backward_expanded`: we save a string and string comparison, with something that is fully equivalent (we'd be throwing away the option of having other backward methods, but it doesn't seem to me that relevant).

niclaurenti:
So in the end what we decided? I revert the changes to use again strings?

AleCandido:
> @AleCandido @felixhekhorn So in the end what we decided? I revert the changes to use again strings?

Nono, I was talking about strings inside Numba (where for sure we can not propagate Python's enum, unless we convert to integers).

About your changes, keep them as they are. I would not really consider the option of using strings again.

felixhekhorn (approved):

AleCandido (approved):

giacomomagni (approved):

PR #251, opened 2023-04-28
Fix apfel_utils

niclaurenti:
Very fast PR to fix `exomark/benchmark/external/apfel_utils`

### DIFF
--- a/src/ekomark/benchmark/external/apfel_utils.py
+++ b/src/ekomark/benchmark/external/apfel_utils.py
@@ -72,8 +72,8 @@ def compute_apfel_data(
 
     # Run
     apf_tabs = {}
-    for q2 in operators["mu2grid"]:
-        apfel.EvolveAPFEL(theory["Q0"], np.sqrt(q2))
+    for mu2 in np.array(operators["mugrid"]) ** 2:
+        apfel.EvolveAPFEL(theory["Q0"], np.sqrt(mu2))
         print(f"Executing APFEL took {(time.perf_counter() - apf_start)} s")
 
         tab = {}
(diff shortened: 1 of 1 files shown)


giacomomagni:
Thanks @niclaurenti can we do the very same for Pegasus in this PR?

src/ekomark/benchmark/external/apfel_utils.py
```diff
@@ -72,8 +72,8 @@ def compute_apfel_data(
 
     # Run
     apf_tabs = {}
-    for q2 in operators["mu2grid"]:
-        apfel.EvolveAPFEL(theory["Q0"], np.sqrt(q2))
+    for mu2 in np.array(operators["mugrid"]) ** 2:
+        apfel.EvolveAPFEL(theory["Q0"], np.sqrt(mu2))
```
giacomomagni, line 76: here you can avoid the square now ...

felixhekhorn:
Just to say that this is another side effect from our favourite #172

src/ekomark/benchmark/external/apfel_utils.py
niclaurenti: yes but in

https://github.com/NNPDF/eko/blob/889b855ca818eb82e39bfc51799e8a7a25d8a884/src/ekomark/benchmark/external/apfel_utils.py#L114

the table is constructed using the squares. To either i loop over the squares and I use `sqrt` or I do the opposite.
I did in this way in order to change the minimum possible

felixhekhorn: if you then square below ...

felixhekhorn (approved):
I assume you checked this fix is all you need :upside_down_face: - looks good!

AleCandido:
Thanks for this! @niclaurenti @giacomomagni

PR #253, opened 2023-04-28
Refactor harmonic cache

giacomomagni:
Since we have introduced a lot of new elements we will need to restructure the harmonic cache.

Reimplementation of #202 

* [x] Add the cache, tests and relative cleaning
* [x] Rewrite all the anomalous dimension modules and relative tests
* [x] Rewrite all the matching conditions modules and relative tests
* [x] Benchmark and check for any numba error


src/ekore/anomalous_dimensions/unpolarized/space_like/as2.py
```diff
@@ -70,21 +71,23 @@ def gamma_nsp(n, nf, sx):
...
-    Sp2p = harmonics.S2(n / 2)
-    Sp3p = harmonics.S3(n / 2)
-    g3n = harmonics.g_functions.mellin_g3(n, S1)
+    S1 = c.get(c.S1, cache, n)
+    S2 = c.get(c.S2, cache, n)
+    Sp1p = c.get(c.S1h, cache, n)
+    Sp2p = c.get(c.S2h, cache, n)
+    Sp3p = c.get(c.S3h, cache, n)
+    # TODO: check is this True or False ?
+    g3n = c.get(c.g3, cache, n, is_singlet=False)
```
giacomomagni, line 90: @felixhekhorn Do we want to still compute something at the level of `__init__` ?? Otherwise we will need to pass an additional
`is_singlet` argument around...

src/ekore/harmonics/__init__.py
```diff
@@ -2,266 +2,131 @@
...
-    sx_qed_ns.append(S2p1h)
-    S3p1h = S3((n + 1.0) / 2.0)
-    sx_qed_ns.append(S3p1h)
-    g3N = g_functions.mellin_g3(n, s1)
-    sx_qed_ns.append(g3N)
-    S1p2 = polygamma.recursive_harmonic_sum(s1, n, 2, 1)
-    g3Np2 = g_functions.mellin_g3(n + 2.0, S1p2)
-    sx_qed_ns.append(g3Np2)
-    return sx_qed_ns
+# TODO: CLEAN
```
AleCandido, line 12: Remember doing this before attempting the merge.

src/ekore/harmonics/cache.py
```diff
@@ -0,0 +1,236 @@
...
+from .g_functions import mellin_g3
+from .polygamma import recursive_harmonic_sum
+
+# here a register of all possible functions
+S1 = 0  # = S_1(N)
+S2 = 1
+S3 = 2
+S4 = 3
+S5 = 4
+Sm1 = 5
```
AleCandido, line 16: If you wish, a slightly more flexible option might be:
```suggestion
_index = iter(range(1000))
S1 = next(_index)  # = S_1(N)
S2 = next(_index)
S3 = next(_index)
S4 = next(_index)
S5 = next(_index)
Sm1 = next(_index)
```

Because you do not really care about the actual number, for as long as it is a stable one, and in this way they would be assigned at import time (not from literals, but still fully constants).
This would also make it simpler to insert in the middle, e.g.
```py
_index = iter(range(1000))
S1 = next(_index)  # = S_1(N)
S2 = next(_index)
S3 = next(_index)
S4 = next(_index)
S5 = next(_index)
S6 = next(_index)
Sm1 = next(_index)
```
modifying only the line involved.
Furthermore, it would stress that the actual number is irrelevant.

However, they are minor points, so it is definitely a detail.

AleCandido: The number in the iterator is just a stupidly high number.
You don't care about the exact one, if available it could have been even infinite, since if you're not iterating until infinite you simply use the initial segment, whatever the length is.

src/ekore/harmonics/__init__.py
giacomomagni: Sure I marked them explicitly!

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
```diff
@@ -7,12 +7,14 @@
 from eko import constants
 from eko.constants import zeta2, zeta3
 
-from ....harmonics import w1, w2, w3
+from ....harmonics import cache as c
+
+# TODO: CLEAN: all the calls to polygamma could be reduced to some harmonics...
 from ....harmonics.polygamma import cern_polygamma as polygamma
```
giacomomagni, line 13: @t7phy I noticed here there are still some explicit calls to `cern_polygamma`, 
would you mind simplify them in terms of the harmonics so we can use the cache?

src/ekore/anomalous_dimensions/polarized/space_like/as3.py
```diff
@@ -341,10 +349,11 @@ def gamma_nss(N, nf, sx):
         :math:`\\gamma_{ns,s}^{(2)}(N)`
 
     """
-    S1 = sx[0]
-    S3 = sx[2]
-    Sm1, Sm2, Sm3, _ = harmonics.smx(N, sx, False)
-    Sm21 = harmonics.Sm21(N, S1, Sm1, False)
+    # TODO: CHECK is this true or false ??
```
giacomomagni, line 352: here we need to double check this has to be continued on odd or even...

src/ekore/harmonics/cache.py
giacomomagni: thanks for this suggestion, we can indeed do this way

giacomomagni:
* [ ] Add the cache, tests and relative cleaning
* [x] Rewrite all the anomalous dimension modules and relative tests
* [ ] Rewrite all the matching conditions modules and relative tests
* [ ] Benchmark and check for any numba error

src/ekore/harmonics/__init__.py
AleCandido: I know. I just opened the conversation to make it explicit even in the GitHub interface :)

src/ekore/anomalous_dimensions/polarized/space_like/as3.py
giacomomagni: fixed in e96738befac3cd2d8c5f54dca3264cf2ec37bd2d

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
t7phy: I had a look at the two functions which use them, gamma_nsp and gamma_nsm, and 4 expressions caught my attention: SPMOM, SLV, SSTR2P and SSTR3P. SPMOM is equivalent to g3 (as far as I can tell). However, I am unable to identify SLV, SSTR2P and SSTR3P, and which functions to correspond to from the harmonics. 
The other problem I am having is that for some reason running pytest for functions with the new cache is giving me Fatal Python Error: Aborted. So the new cache (and therefore eko) is basically not working on my machine which is of course making things difficult for me to do even trial and error to figure out the exact functions.

giacomomagni: If the time-like expressions follows the same syntax as the unpolarised,  I'd say the equivalences are:
```
SPMOM -> g3
```
then you can use this rules to simplify the others in terms of harmonics and g3:
```
HarmonicRules={
 PolyGamma[0, n__] -> S[1, - 1 + n] - EulerGamma,
 PolyGamma[1, n__] -> - S[2, - 1 + n] + Zeta[2],
 PolyGamma[2, n__] -> 2 S[3, - 1 + n] - 2 Zeta[3],
 PolyGamma[3, n__] -> - 6 S[4, - 1 + n] + 6 Zeta[4],
 PolyGamma[4, n__] -> 24 S[5, - 1 + n] - 24 Zeta[5],
 PolyGamma[5, n__] -> - 120 S[6, - 1 + n] + 120 Zeta[6],
 PolyGamma[6, n__] -> 720 S[7, - 1 + n] - 720 Zeta[7]
};
``` 
ending up in: 
```
SSTR3P ->  S[3, N/2]
SSTR2P ->  S[2, N/2]
SSCHLP ->  g3 + S1/N^2 - 1/12 \[Pi]^2 (2/(N + N^2) + S[1, 1/2 + N/2] - S[1, N/2]) - 5 Zeta[3]/8
```
If you find the same and test are passing we can commit this. 
Note that all the terms with g3 should appear inside `S_{-2,1}` but we know this is not straightforward since there are alle the difference `(-1)^n` prescriptions around, so we can give up on this...

Regarding the compilation error, I suspect numba is messing up something. But `pytest` should always work since it runs with numba disabled (as you can see below ... ). I will try to investigate  tomorrow if this implementation is actually able to run with `nb.jit`

t7phy: couple things: 
1. for SSCHLP, if we have S[1, 1/2 + N/2], I guess we will need to add this to cache also right, maybe as S1p3h ..?
2. I modified SPMOM, SSTR3P, SSTR2P as above, however this is giving errors in tests: https://github.com/NNPDF/eko/actions/runs/4855646050/jobs/8654404902
3. I don't think the error is coming from numba, I had a quick look at the error text and it didnt seem so...

giacomomagni: hi, 
 1. Yes we are already caching S[1, 1/2 + N/2] as `S1ph`, so no problem.
 2. This errors suggest the replacement was not fully successfull. Can you open a PR?
 3. These errors are not coming from numba, I meant the one about segmentation fault, you mentioned before. 
 Do you have any more detail?

src/ekore/harmonics/cache.py
giacomomagni: done in https://github.com/NNPDF/eko/pull/253/commits/cd34b1c666755450c64c19ec49a463a2f959266b

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
t7phy: As of now, on my laptop, pytest is just hanging after 'Collected ** items'. It's working fine on master but not on this branch.

felixhekhorn: Just to say that for me it is running without problems ...

felixhekhorn: > PS: something you can always try is to remove the `__pycache__` folders and recompile

Note that I'm discussing this also a bit with the numba developers https://github.com/numba/numba/issues/8926, so far without large movements ...

the upshot is to delete the "correct" pycache - which should be the one containing the *first* numba call, which for us is `quad_ker` (and friends), so `eko/ev_op/__pycache__`

t7phy: but pytest doesn't use the compiled version of the functions, right?

felixhekhorn: > but pytest doesn't use the compiled version of the functions, right?

No it should not - numba should be deactivated there

giacomomagni: I fear we do have some problems... I tried to run the pegasus benchmarks (`python pegasus_bench.py`), 
first time it took a while but all good. Now if I rerun I get: 

```
 ──────────────────────────────────────────── 
  Theories: 2 OCards: 1 PDFs: 1 ext: pegasus  
 ──────────────────────────────────────────── 
Computing for theory=af2df25, ocard=cd6c3a9 and pdf=ToyLH ...
ThresholdsAtlas [0.00e+00 - 4.00e+00 - 2.02e+01 - 3.00e+04 - inf], ref=None @ None
Interpolation: number of points = 49, polynomial degree = 4, logarithmic = True
ThresholdsAtlas [0.00e+00 - 4.00e+00 - 2.02e+01 - 3.00e+04 - inf], ref=2.0000000000000004 @ 3
ThresholdsAtlas [0.00e+00 - 2.00e+00 - 1.01e+01 - 1.50e+04 - inf], ref=2.0000000000000004 @ 3
Strong Coupling: a_s(µ_R^2=2.000000)^(nf=3)=0.027852=0.350000/(4π)
{'polarized': False}
{'time_like': False}
Prepare threshold operator
Evolution: computing operators 2.000000e+00 -> 4.000000e+00, nf=3
Evolution: µ_R^2 distance: 1.000000e+00 -> 2.000000e+00
Scale Variation: (µ_F/µ_R)^2 = 5.000000e-01, mode: exponentiated
Evolution: a_s distance: 3.536335e-02 -> 2.785212e-02
Evolution: order: (2, 0), solution strategy: iterate-exact
zsh: segmentation fault  python pegasus_bench.py
```
ad disabling numba it works again:

```
(eko_dev) giacomomagni@Giacomos-MacBook-Pro benchmarks % export NUMBA_DISABLE_JIT=1
(eko_dev) giacomomagni@Giacomos-MacBook-Pro benchmarks % python pegasus_bench.py   
 ──────────────────────────────────────────── 
  Theories: 2 OCards: 1 PDFs: 1 ext: pegasus  
 ──────────────────────────────────────────── 
Computing for theory=af2df25, ocard=cd6c3a9 and pdf=ToyLH ...
ThresholdsAtlas [0.00e+00 - 4.00e+00 - 2.02e+01 - 3.00e+04 - inf], ref=None @ None
Interpolation: number of points = 49, polynomial degree = 4, logarithmic = True
ThresholdsAtlas [0.00e+00 - 4.00e+00 - 2.02e+01 - 3.00e+04 - inf], ref=2.0000000000000004 @ 3
ThresholdsAtlas [0.00e+00 - 2.00e+00 - 1.01e+01 - 1.50e+04 - inf], ref=2.0000000000000004 @ 3
Strong Coupling: a_s(µ_R^2=2.000000)^(nf=3)=0.027852=0.350000/(4π)
{'polarized': False}
{'time_like': False}
Prepare threshold operator
Evolution: computing operators 2.000000e+00 -> 4.000000e+00, nf=3
Evolution: µ_R^2 distance: 1.000000e+00 -> 2.000000e+00
Scale Variation: (µ_F/µ_R)^2 = 5.000000e-01, mode: exponentiated
Evolution: a_s distance: 3.536335e-02 -> 2.785212e-02
Evolution: order: (2, 0), solution strategy: iterate-exact
Evolution: computing operators - 1/49 took: 35.609883 s
...
``` 
 
I also verified that if I remove the `__pycache__` then I'm also able to rerun with numba (modulo compilation time)
 Any idea?

felixhekhorn: also in the CI we did not converge to anything ... I started a new run here https://github.com/NNPDF/eko/actions/runs/4861110401/jobs/8665750747 but I don't expect anything ...

giacomomagni: What can it be? The optional argument of `cache.get` ? 
But we do have other functions using optional arguments and numba decorated?

giacomomagni: From https://numba.pydata.org/numba-doc/latest/developer/caching.html :

```
Requirements for cacheable function:

The LLVM module must be self-contained, meaning that it cannot rely on other compiled units without linking to them.

The only allowed external symbols are from the [NRT](https://numba.pydata.org/numba-doc/latest/developer/numba-runtime.html#arch-numba-runtime) or other common symbols from system libraries (i.e. libc and libm).

Debugging note:

Look for the usage of inttoptr in the LLVM IR or target_context.add_dynamic_add() in the lowering code in Python. They indicate potential usage of runtime address. Not all uses are problematic and some are necessary. Only the conversion of constant integers into pointers will affect caching.

Misuse of dynamic address or dynamic symbols will likely result in a segfault.

Linking order matters because unused symbols are dropped after linking. Linking should start from the leaf nodes of the dependency graph.
```

giacomomagni: Okay the problem are the recursive calls of get inside get.

felixhekhorn: > Okay the problem are the recursive calls of get inside get.

good catch! can we do a two step process? like a top-level `get` which in turn is calling `compute_composed` for the composed sums?

felixhekhorn: So: an infinite recursion can also lead to a segmentation fault, but on a first glance we're not falling into this scenario ... can you maybe try to investigate on a smaller code base?

AleCandido: @giacomomagni another line of investigation: can you make a minimal example, in a single script, with as little code as possible, that is generating the issue?

giacomomagni: here is how I find it:
This script works (many times) if you compute S1 only (having commented all the `elif` containing nested calls to `get` in `src/harmonics/cache.py` ), but it fails as one of these `elif` are included for instance when you ask for `g3`: 

```
import numba as nb
from ekore.harmonics import cache as c



@nb.njit(cache=True)
def my_func(cache):
    n = 1.0
    S1 = c.get(c.S1, cache, n)
    # g3 = c.get(c.g3, cache, n)
    print(S1)

cache = c.reset()
my_func(cache)
```
Of course each time you change something in `src/harmonics/cache.py` but not in the snippet you need to remove the `__pycache__` manually.
To have more insight of  what happens you can also do `export NUMBA_DEBUG_CACHE=1`

giacomomagni: > So: an infinite recursion can also lead to a segmentation fault, but on a first glance we're not falling into this scenario ... can you maybe try to investigate on a smaller code base?

The problem is not the infinite recursion (if that was the issue you will not even able to run anything with `export NUMBA_DISABLE_JIT=1`), but I believe is how numba deals with the `__pycache__`

giacomomagni:
https://github.com/NNPDF/eko/pull/253/commits/f10329d5251770d9f71674b9e7ac6f181b61397e should make numba happy. If you have any suggestion how to improve the code, please fell free to commit or comment.

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
AleCandido: > This script works (many times) if you compute S1 only (having commented all the `elif` containing nested calls to `get` in `src/harmonics/cache.py` ), but it fails as one of these `elif` are included for instance when you ask for `g3`:

Thanks for the script, but if possible an even more minimal version would be helpful.
Can you replace the `harmonics` import with the actual cache, and trim it down until you have something minimal that generates the issue?
I.e., it would be really helpful to have a minimal *self-contained* failing script (no imports but Numba).

I do not believe that you need the whole `ekore` to make it crash (otherwise it could only be something like a memory issue), so it would make sense to make it more and more minimal. Your script is a great starting point, but taking into account `ekore` still too complex.

giacomomagni: Sure here you have:

```
import numba as nb
import numpy as np

# fake functions
@nb.njit(cache=True)
def f1():
    return 1
@nb.njit(cache=True)
def f2(x):
    return 2 + x

# fake keys
F1 = 0
F2 = 1

# reset
@nb.njit(cache=True)
def reset():
    return np.full(2, np.nan, np.float_)

@nb.njit(cache=True)
def not_working_get(key, cache):
    s = cache[key]
    if not np.isnan(s):
        return s 
    if key == F1:
        s = f1()
    elif key == F2:
        s = f2(not_working_get(F1, cache))
    cache[key] = s
    return s

@nb.njit(cache=True)
def working_get(key, cache):
    s = cache[key]
    if not np.isnan(s):
        return s
    if key == F1:
        s = f1()
    elif key == F2:
        s = f2(f1())
    cache[key] = s
    return s

cache = reset()
# this will work
working_get(F2, cache)
# this will work once
# not_working_get(F2, cache)
print(cache[F2])
```

In any case the most important proof is [here](https://github.com/NNPDF/eko/actions/runs/4863651845/jobs/8671636590)

AleCandido: > In any case the most important proof is [here](https://github.com/NNPDF/eko/actions/runs/4863651845/jobs/8671636590)

Sorry, maybe I didn't properly express myself. I didn't want a proof that there is an issue (I fully believe you), just a minimal example to play with and try to understand what's going wrong and why.

And sorry to be a bit dense, but what does exactly mean that "this will work only once"? How can I make a sensible failing `assert`?

giacomomagni: "this will work only once" means: you run it, it works. You run it again without modifying anything you get seg fault.
You delete `__pycache__` and run it, it will work again.

AleCandido: I see, now I can investigate a bit more

AleCandido: Just notice that, in the script above, I'm not even invoking recursion, because I'm directly calling it with `exit = True`.

The presence of recursion breaks the cache on its own.

AleCandido: Recursion in Numba has limited supported, but for what I found at the moment, the main problem is mutual recursion. No mention found yet about cache.
https://numba.readthedocs.io/en/stable/proposals/typing_recursion.html

My best proposal is to **avoid caching the cache getter**, i.e. the recursive function.

src/ekore/harmonics/cache.py
```diff
@@ -0,0 +1,229 @@
...
+    """Compute Sm2 if not yet in cache."""
+    if np.isnan(cache[Sm2]):
+        cache = update(w2.S2, S2, cache, n)
+        cache = update(w2.S2, S2mh, cache, (n - 1) / 2)
+        cache = update(w2.S2, S2h, cache, n / 2)
+        cache[Sm2] = w2.Sm2(n, cache[S2], cache[S2mh], cache[S2h], is_singlet)
+    return cache
+
+
+@nb.njit(cache=True)
```
AleCandido, line 84:
```suggestion
@nb.njit
```
https://github.com/NNPDF/eko/pull/253#discussion_r1183392401

AleCandido: It is true that everything that is used under Numba has to be numbified, but in principle this property doesn't extend to cache: we should be able to mix cached and uncached Numba functions.

Since the problem seems to derive from this one, let's simply move this single function out of cache.

AleCandido: The following seems to work fine:
```py
import numba as nb

@nb.njit(cache=True)
def callee(x):
    return x == 0

@nb.njit
def caller(exit):
    if exit:
        return 1.
    return callee(caller(0))

@nb.njit(cache=True)
def supercaller():
    return caller(True)

print(supercaller())
```

AleCandido: Sorry, I take it back: if you replace `caller(True)` with `caller(False)` in `supercaller`, it is crashing again.

giacomomagni: yes, from what I saw yesterday as soon as you have a not cached functions all the `supercallers` cache will not work.

src/ekore/anomalous_dimensions/unpolarized/time_like/as2.py
giacomomagni: thanks for the reference, unfortunately it's only a draft, sigh

felixhekhorn (approved):
Unfortunately, our hope of a small speed up doesn't seem to be true (LHA bot ran here in 33m and on master on 31m) - nevertheless, I'm still convinced this is a good idea ... (also the impact on `eko` is minimal which is good)

(in a future PR we may want to harmonize the arguments of the various functions `n,cache,nf` vs. `n,nf,cache` - but that is really minor ...)

AleCandido:
> (LHA bot ran here in 33m and on master on 31m)

The exact timing should not be measured on GitHub Actions machines: they are not uniform, and most likely virtual, so you never know the exact performances on comparable hardware.

However, the message is that the order of magnitude is unchanged.

giacomomagni:
> Unfortunately, our hope of a small speed up doesn't seem to be true (LHA bot ran here in 33m and on master on 31m) - nevertheless, I'm still convinced this is a good idea ... (also the impact on `eko` is minimal which is good)

Are you sure this do not depends also on other factors? I checked that some older benchmark took longer other were quicker... Most likely we haven't gained much apart in code cleaning and #143


> (in a future PR we may want to harmonize the arguments of the various functions `n,cache,nf` vs. `n,nf,cache` - but that is really minor ...)

True

Issue #241, opened 2023-04-04
LHA SV fails for smallest x in CI


felixhekhorn:
we still have the 100% error [in github](https://github.com/NNPDF/eko/actions/runs/4606070102/jobs/8138911316#step:7:421):
```
Computing for theory=45286e9, ocard=cc12964 and pdf=ToyLH ...
Compute external result

Log added, hash=2e5273d953ef355a2f48badae3729333e94e6808ae941048e2e2e943771b0e63


 ─── 
  S  
 ─── 
               x       Q2         eko     eko_error         LHA  percent_error
0   1.000000e-07  10000.0    0.000000  0.000000e+00  251.422304    -100.000000
1   1.000000e-06  10000.0  132.769958  1.641839e-04  133.133286      -0.272906
```
but I cannot reproduce this locally
```
In [2]: dfl(-1)
Out[2]: 

- theory: `45286e92f0fbedc3ed4a84b47547190efc4bde84583e5782716a5c5c70448aed`
- obs: `8beae75234d8ba49de7079e5f76d3916dd810708d31bb54f85505faf1c470055`
- using PDF: *ToyLH*

S
               x       Q2         eko     eko_error         LHA  percent_error
0   1.000000e-07  10000.0  250.741696  3.320808e-04  251.422304      -0.270703
1   1.000000e-06  10000.0  132.769958  1.641844e-04  133.133286      -0.272906
```
Note that the obs hash is not the same ... I seem to remember to have seen this before ... @giacomomagni ? or @t7phy ?

_Originally posted by @felixhekhorn in https://github.com/NNPDF/eko/issues/238#issuecomment-1495732624_

giacomomagni:
Having rerun this locally after recompiling everything from scratch, I confirm I'm not able to reproduce the error.

felixhekhorn:
@giacomomagni can you confirm my obs hash?

felixhekhorn:
> EDIT: Note that ls(o) is currently broken:

I think this should have been fixed in https://github.com/NNPDF/eko/pull/231 - and indeed I can not reproduce it. Can you check if you regenerate the DB it works also for you? (backup the old if you do not want to loose it - maybe a simple rename there can save you)

felixhekhorn:
> but `obs` is different from both of you...

[...]

PR #258, opened 2023-05-03
Extend support to Python 3.11

AleCandido:
Thanks to new Numba release https://github.com/numba/numba/releases/tag/0.57.0


felixhekhorn:
There seems to be two different type of errors around:
1. a numerical issue present in all Python versions 3.8 - 3.1:
```
FAILED tests/eko/test_interpolation.py::TestBasisFunction::test_log_eval_N - AssertionError: 
Arrays are not almost equal to 7 decimals
 ACTUAL: 0.718281828459045
 DESIRED: 2.718281828459045
```
this might be related to #241 
2. an IO issue present only in 3.8
```
/home/runner/work/eko/eko/src/eko/io/dictlike.py:192: in load_typing
    return load_field(variant, value)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

type_ = numpy.ndarray[typing.Any, numpy.dtype[+ScalarType]], value = None

    def load_field(type_, value):
        """Deserialize dataclass field."""
        # TODO: nice place for a match statement...
        if type(type_) is typing.NewType:
            return load_field(type_.__supertype__, value)
        try:
            # before py3.10 typing.NewType was just a function, so the check above
            # would  fail
            return load_field(type_.__supertype__, value)
        except AttributeError:
            pass
    
        if typing.get_origin(type_) is not None:
            # this has to go first since for followin ones I will assume they are
            # valid classes, cf. the module docstring
            return load_typing(type_, value)
    
>       assert inspect.isclass(type_)
E       AssertionError
```
@AleCandido this looks like numpy has changed something

giacomomagni:
> On top there is a new numba issue inside OME appearing (sounds like something is not an array, but it should be ...)

Maybe it's because A is defined only inside `if` and thus can be `None` ?

AleCandido:
The other problem is that some of the updated deps is now causing issues with typing in py3.8...

Since it is limited to "new dep + py3.8", I'd be really tempted to say: well, let's drop py3.8. However, that is definitely not an option, it is still incredibly popular (see "Python Minor" download plots for [NumPy](`https://pypistats.org/packages/numpy`) and [all packages](https://pypistats.org/packages/__all__) on PyPI).
To be fair, even py3.7 is still very used, actually the most used one: but that's their fault, it is really approaching EOL...

felixhekhorn:
Okay, problem 1 is solved: this was due to float numerics, i.e. `np.log` and `np.exp` are not inverse functions to each other. This fact we knew all along, but still so far it did work - I assume they might have been on a subset of float numbers. Apparently now one of the two changed its implementation and now they diverge also for our test case. Solution was to relax the comparisons in `log_evaluate_Nx` - should we adjust also the others? (`evaluate_Nx` and `evaluate_x` with the later being adjusted already halfway)

giacomomagni:
> `np.log` and `np.exp` are not inverse functions

Just for curiosity where did you see it?  https://numpy.org/doc/stable/reference/generated/numpy.log.html ?

giacomomagni:
> I deduced from here https://github.com/NNPDF/eko/actions/runs/4936234492/jobs/8823523488#step:11:124 (where a `np.log(np.exp(-1))` call took place) - and it makes sense, I think ...

I see then the definition in the documentation is not fully accurate 😬

AleCandido:
> I see then the definition in the documentation is not fully accurate :grimacing: 

I believe the definition to be accurate enough, simply is defined something different of what you think: they are defining the $log(x)$ function, as a function on real numbers, for the user who might wonder what the `np.log(x)` is implementing (or, if you wish, approximating). E.g. in case you had doubts on the basis.

However, the `np.log(x)` is a function `float -> float`, and not $\mathbb{R} \to \mathbb{R}$, and as a `float` function is not the inverse of `np.exp(x)`.

AleCandido:
The remaining issue arises from `npt.NDArray`: until `numpy==1.23.1` it was a class, so there were two different behaviors between py3.8 and py>3.8, but both valid

- **py>3.8** is able to find the origin class of the type hint
    https://github.com/NNPDF/eko/blob/999fb30bebf2d7c79651615c673695cd645c4e06/src/eko/io/dictlike.py#L144-L147
    once resolved, that class (i.e. `np.ndarray`) is immediately used
- **py3.8** was not able to reconstruct the origin, but since `npt.NDArray` was a class on its own, it was passing this gate
    https://github.com/NNPDF/eko/blob/999fb30bebf2d7c79651615c673695cd645c4e06/src/eko/io/dictlike.py#L149
    and resolved later on as NumPy array
    https://github.com/NNPDF/eko/blob/999fb30bebf2d7c79651615c673695cd645c4e06/src/eko/io/dictlike.py#L155-L159

However, since `numpy==1.23.2`, `npt.NDArray` is not any longer a class, and for this reason py3.8 is failing.
From the [list of PRs merged](https://numpy.org/doc/stable/release/1.23.2-notes.html#pull-requests-merged), I believe the responsible to be https://github.com/numpy/numpy/pull/22032.

However, the weird behavior is first an issue in py3.8, for which I do not understand why `typing.get_origin` is behaving differently than in the other Python versions (the function has been introduced in py3.8 itself, and no public change is reported in the [docs](https://docs.python.org/3/library/typing.html#typing.get_origin)).

AleCandido:
My change is minimal (just sorting differently some conditions that were already present), and seems like it is doing its job.

Even @felixhekhorn changes are minimal, and perfectly reasonable (I would approve them with the tick, but I'm the author of the PR).

Let's wait for the CI, and if @giacomomagni approves them as well I would merge immediately.

Btw, I just stepped into py3.11 on my system and I'd like to be able to run EKO :P
(@felixhekhorn is even waiting since longer)

giacomomagni (approved):

AleCandido:
Thanks @giacomomagni I'm just waiting a little more to give @felixhekhorn the chance to have a final look.

I will merge today in any case.

felixhekhorn (approved):

PR #261, opened 2023-05-08
Drop fact scale

giacomomagni:
Since we are close to make a new release I'd propose to drop this deprecated feature now.


felixhekhorn (approved):
let's wait for the CI to confirm my last commit didn't break the benchmarks - else this is fine

Issue #272, opened 2023-05-22
In legacy runcards `nf0` is not correctly determined


andreab1997:
In https://github.com/NNPDF/eko/blob/099433602d209331763429ece166d5cf4b49a984/src/eko/io/runcards.py#L186 we use `Qref` instead of `Q0` which results in a wrong determination of `nf0`.

PR #273, opened 2023-05-22
Fix `nf0` in runcards

andreab1997:
Solve #272


AleCandido (approved):
Thanks for the fix (and sorry for the bug in the first place... 🐛)

felixhekhorn:
just for tracking: this is a patch to #242

PR #274, opened 2023-05-22
Fix apfel benchmarks

niclaurenti:
Fixing apfel benchmarks. Tomorrow I will run also the other benchmarks to see if they are broken or not

### DIFF
--- a/src/eko/io/runcards.py
+++ b/src/eko/io/runcards.py
@@ -225,7 +225,7 @@ class Legacy:
             mugrid = old["mugrid"]
         else:
             mu2grid = old["Q2grid"] if "Q2grid" in old else old["mu2grid"]
-            mugrid = np.sqrt(mu2grid)
+            mugrid = np.sqrt(mu2grid).tolist()
         new["mugrid"] = flavored_mugrid(
             mugrid,
             list(self.heavies("m%s", old_th)),
@@ -314,7 +314,7 @@ def flavored_mugrid(mugrid: list, masses: list, matching_ratios: list):
 
     """
     atlas = default_atlas(masses, matching_ratios)
-    return [(float(mu), nf_default(mu**2, atlas)) for mu in mugrid]
+    return [(mu, nf_default(mu**2, atlas)) for mu in mugrid]
 
 
 # TODO: move to a more suitable place
--- a/src/eko/io/runcards.py
+++ b/src/eko/io/runcards.py
@@ -309,7 +309,7 @@ def flavored_mugrid(mugrid: list, masses: list, matching_ratios: list):
 
     """
     atlas = default_atlas(masses, matching_ratios)
-    return [(mu, nf_default(mu**2, atlas)) for mu in mugrid]
+    return [(float(mu), nf_default(mu**2, atlas)) for mu in mugrid]
 
 
 # TODO: move to a more suitable place
--- a/src/eko/io/runcards.py
+++ b/src/eko/io/runcards.py
@@ -173,11 +173,16 @@ class Legacy:
 
         new["order"] = [old["PTO"] + 1, old["QED"]]
         alphaem = self.fallback(old.get("alphaqed"), old.get("alphaem"), default=0.0)
+        if "Qedref" in old:
+            em_running = bool(np.isclose(old["Qedref"], old["Qref"]))
+        else:
+            em_running = False
         ms = self.heavies("m%s", old)
         ks = self.heavies("k%sThr", old)
         new["couplings"] = dict(
             alphas=old["alphas"],
             alphaem=alphaem,
+            em_running=em_running,
             scale=old["Qref"],
             num_flavs_ref=old["nfref"],
             max_num_flavs=old["MaxNfAs"],
--- a/src/ekomark/benchmark/external/apfel_utils.py
+++ b/src/ekomark/benchmark/external/apfel_utils.py
@@ -72,8 +72,8 @@ def compute_apfel_data(
 
     # Run
     apf_tabs = {}
-    for mu in operators["mugrid"]:
-        apfel.EvolveAPFEL(theory["Q0"], mu)
+    for mu2 in operators["Q2grid"]:
+        apfel.EvolveAPFEL(theory["Q0"], np.sqrt(mu2))
         print(f"Executing APFEL took {(time.perf_counter() - apf_start)} s")
 
         tab = {}
@@ -111,7 +111,7 @@ def compute_apfel_data(
             evol_pdf = rotate_flavor_to_evolution @ pdfs
             tab = dict(zip(evol_basis, evol_pdf))
 
-        apf_tabs[mu**2] = tab
+        apf_tabs[mu2] = tab
 
     ref = {
         "target_xgrid": target_xgrid,
--- a/src/ekomark/benchmark/external/apfel_utils.py
+++ b/src/ekomark/benchmark/external/apfel_utils.py
@@ -73,7 +73,7 @@ def compute_apfel_data(
     # Run
     apf_tabs = {}
     for mu in operators["mugrid"]:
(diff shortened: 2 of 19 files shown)


src/ekomark/benchmark/runner.py
```diff
@@ -245,10 +246,11 @@ def log(self, theory, _, pdf, me, ext):
                 flavor_rotation=rotate_to_evolution,
                 qed=qed,
             )
-        for q2 in q2s:
+        for (q2, ref_pdfs), (q2a, res) in zip(ext["values"].items(), pdf_grid.items()):
+            np.testing.assert_allclose(q2, q2a)
```
AleCandido, line 250:
```suggestion
        np.assert_allclose(ext["values"].keys(), pdf_grid.keys())
        meaningful_name = zip(ext["values"].items(), pdf_grid.values())
        for (q2, ref_pdfs), res in meaningful_name:
```
To marginally simplify, you could split the check on its own (assuming NumPy is supporting your aggregated `.keys()` with allclose).

Usage of meaningful names always helps, making it clear at first site.

All in all, it is pretty fine also as it as.

felixhekhorn: fine ... but then I prefer my solution

src/ekomark/benchmark/runner.py
```diff
@@ -245,10 +246,11 @@ def log(self, theory, _, pdf, me, ext):
...
             log_tab = dfdict.DFdict()
-            ref_pdfs = ext["values"][q2]
-            res = pdf_grid[q2]
+            # ref_pdfs = ext["values"][q2]
+            # res = pdf_grid[q2]
```
felixhekhorn, line 253:
```suggestion
```

src/ekomark/benchmark/external/apfel_utils.py
```diff
@@ -72,8 +72,8 @@ def compute_apfel_data(
 
     # Run
     apf_tabs = {}
-    for mu2 in operators["Q2grid"]:
-        apfel.EvolveAPFEL(theory["Q0"], np.sqrt(mu2))
+    for mu in operators["mugrid"]:
```
felixhekhorn, line 75: why did you revert?

src/ekomark/benchmark/runner.py
felixhekhorn, line 1: Please also drop [q2s as said here](https://github.com/NNPDF/eko/pull/274#issuecomment-1558964805)

src/ekomark/benchmark/external/apfel_utils.py
```diff
@@ -72,8 +72,8 @@ def compute_apfel_data(
...
+        apfel.EvolveAPFEL(theory["Q0"], np.sqrt(mu**2))
```
AleCandido, line 76:
```suggestion
        apfel.EvolveAPFEL(theory["Q0"], mu)
```

src/ekomark/benchmark/external/apfel_utils.py
AleCandido: Considering the line below, if `mugrid` is available the current version makes sense (and it avoids the old `Q2` nomenclature)

src/eko/io/runcards.py
```diff
@@ -309,7 +309,7 @@ def flavored_mugrid(mugrid: list, masses: list, matching_ratios: list):
 
     """
     atlas = default_atlas(masses, matching_ratios)
-    return [(mu, nf_default(mu**2, atlas)) for mu in mugrid]
+    return [(float(mu), nf_default(mu**2, atlas)) for mu in mugrid]
```
AleCandido, line 312: Why do you explicitly need to call `float`?

It is not explicit in the type above, but I thought `mugrid` as a `list[float]`.

src/ekomark/benchmark/external/apfel_utils.py
niclaurenti: @felixhekhorn I think that the other day we were looking by mistake to an old version of the code. In fact in the current master, in ekomark operator has `mugrid` instead of `Q2grid`

src/eko/io/runcards.py
niclaurenti: It's because otherwise we had problems with the dumping into a .yaml since it was a numpy object

felixhekhorn: After https://github.com/NNPDF/eko/blob/633fbb48f6a5d85b49757f7417c1c0ca4b5f16ff/src/eko/io/runcards.py#L223 elements are numpy-objects (so tuple for yaml)

AleCandido: But that's an actual array, so it is not a `list` at all.

The function 
https://github.com/NNPDF/eko/blob/633fbb48f6a5d85b49757f7417c1c0ca4b5f16ff/src/eko/io/runcards.py#L224-L228
should be simply called with `mugrid.tolist()`, instead of just `mugrid` (my bad, most likely I changed the function afterward - we definitely need `mypy` running...).

felixhekhorn: yes - after that (my) line `mugrid` is an numpy-array containing numpy objects which then via your snippet arrive to https://github.com/NNPDF/eko/blob/633fbb48f6a5d85b49757f7417c1c0ca4b5f16ff/src/eko/io/runcards.py#L312
and so the observable card receives a list of tuples with the first element a numpy float and the second an integer, because there we already have the explicit cast in place: https://github.com/NNPDF/eko/blob/633fbb48f6a5d85b49757f7417c1c0ca4b5f16ff/src/eko/matchings.py#L188

felixhekhorn: (it seems this was detected by unit testing, but not this new change ...)

felixhekhorn: @AleCandido are you happy with this explanation?

AleCandido: Sorry, I haven't replied before: I have nothing against the explanation, only something against the present situation.

Can we just call `.tolist()` there and normalize it?

felixhekhorn:
> edit: your commit fixed it

indeed - I like the fix in #284 better

niclaurenti:
@felixhekhorn @AleCandido In the apfel benchmarks (and also in banana) we still have `QrefQED` that is no longer present in `eko`. My proposal was to restore `Qedref` to make it compatible with the nnpdf runcards and use it to switch on and off the em_running

felixhekhorn:
> @felixhekhorn @AleCandido In the apfel benchmarks (and also in banana) we still have `QrefQED` that is no longer present in `eko`. My proposal was to restore `Qedref` to make it compatible with the nnpdf runcards and use it to switch on and off the em_running

didn't we have this behaviour already at some point of the 1M iterations? `aem_running = QrefQED == Qref`

niclaurenti:
Last commit shows what I had in mind: I changed `nan` with 0 since em_running will be enabled only if `Qref==Qedref` so 0 could be every number different from `Qref`

src/eko/io/runcards.py
felixhekhorn: Have a look to https://github.com/NNPDF/eko/pull/274/commits/a500ddccd22d5a85acf9db9686a2193c0f96954e - the new added test fails if we're not doing anything

felixhekhorn:
@niclaurenti you need to run `poetry lock` in order to get the lock file correct, since there is a hashing algorithm in the game ...

felixhekhorn:
let's wait for https://github.com/NNPDF/eko/actions/runs/5280648705/jobs/9553070195 ...

PS: can you double check you can run APFEL still (i.e. with new banana)?

niclaurenti:
> let's wait for https://github.com/NNPDF/eko/actions/runs/5280648705/jobs/9553070195 ...
> 
> PS: can you double check you can run APFEL still (i.e. with new banana)?

It works

felixhekhorn (approved):

PR #280, opened 2023-06-05
fix Q2 grid in generate_block

giacomomagni:
This PR is the proper fix of https://github.com/NNPDF/nnpdf/pull/1750

### DIFF
--- a/src/ekobox/genpdf/__init__.py
+++ b/src/ekobox/genpdf/__init__.py
@@ -220,10 +220,11 @@ def generate_block(
     xfxQ2: Callable, xgrid: List[float], evolgrid: List[EPoint], pids: List[int]
 ) -> dict:
     """Generate an LHAPDF data block from a callable."""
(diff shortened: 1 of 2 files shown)


AleCandido (approved):
In principle, you should also take care that all the `nf` values are consistent. But never mind, I hope this will become simpler with `partons`, I'm working on it :)

giacomomagni:
> In principle, you should also take care that all the `nf` values are consistent. 

Let's hope this is not causing any problem for the time being. 😅

felixhekhorn (approved):
I wonder if part of https://github.com/NNPDF/nnpdf/pull/1746 should be done (or repeated) here

AleCandido:
> I wonder if part of [NNPDF/nnpdf#1746](https://github.com/NNPDF/nnpdf/pull/1746) should be done (or repeated) here

Don't waste too much effort on that, at least for a while more.

PR #281, opened 2023-06-06
Write the right `nf` in the LHAPDF `.info` file.

scarlehoff:
Currently the `NumFlavors` written by eko in the `info` file correspond to the number of possible partons, while this field is actually the maximum `nf` used for the evolution of `alpha_s`.

Ref: https://lhapdf.hepforge.org/config.html

### DIFF
--- a/src/ekobox/info_file.py
+++ b/src/ekobox/info_file.py
@@ -39,7 +39,7 @@ def build(
     template_info["Authors"] = ""
     template_info["FlavorScheme"] = "variable"
     template_info.update(info_update)
-    template_info["NumFlavors"] = theory_card.heavy.num_flavs_max_pdf
+    template_info["NumFlavors"] = max(nf for _, nf in op.mugrid)
     template_info["Flavors"] = [-6, -5, -4, -3, -2, -1, 1, 2, 3, 4, 5, 6, 21, 22]
     # TODO actually point to input grid
     template_info["XMin"] = float(operators_card.xgrid.raw[0])
--- a/src/ekobox/info_file.py
+++ b/src/ekobox/info_file.py
@@ -39,7 +39,7 @@ def build(
     template_info["Authors"] = ""
     template_info["FlavorScheme"] = "variable"
     template_info.update(info_update)
-    template_info["NumFlavors"] = theory_card.couplings.max_num_flavs
+    template_info["NumFlavors"] = theory_card.heavy.num_flavs_max_pdf
     template_info["Flavors"] = [-6, -5, -4, -3, -2, -1, 1, 2, 3, 4, 5, 6, 21, 22]
     # TODO actually point to input grid
     template_info["XMin"] = float(operators_card.xgrid.raw[0])
--- a/src/ekobox/info_file.py
+++ b/src/ekobox/info_file.py
@@ -39,7 +39,7 @@ def build(
     template_info["Authors"] = ""
     template_info["FlavorScheme"] = "variable"
     template_info.update(info_update)
-    template_info["NumFlavors"] = theory.couplings.max_num_flavs
+    template_info["NumFlavors"] = theory_card.couplings.max_num_flavs
     template_info["Flavors"] = [-6, -5, -4, -3, -2, -1, 1, 2, 3, 4, 5, 6, 21, 22]
     # TODO actually point to input grid
     template_info["XMin"] = float(operators_card.xgrid.raw[0])
--- a/src/ekobox/info_file.py
+++ b/src/ekobox/info_file.py
@@ -39,7 +39,7 @@ def build(
     template_info["Authors"] = ""
     template_info["FlavorScheme"] = "variable"
     template_info.update(info_update)
-    template_info["NumFlavors"] = 14
+    template_info["NumFlavors"] = theory.couplings.max_num_flavs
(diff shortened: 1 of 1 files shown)


src/ekobox/info_file.py
```diff
@@ -39,7 +39,7 @@ def build(
     template_info["Authors"] = ""
     template_info["FlavorScheme"] = "variable"
     template_info.update(info_update)
-    template_info["NumFlavors"] = 14
+    template_info["NumFlavors"] = theory.couplings.max_num_flavs
```
scarlehoff, line 42:
```suggestion
    template_info["NumFlavors"] = theory_card.couplings.max_num_flavs
```

andreab1997:
Ok for me but what about the `Flavors` list? It should be consistent with `NumFlavors` right?

felixhekhorn:
> Currently the `NumFlavors` written by eko in the `info` file correspond to the number of possible partons, while this field is actually the maximum `nf` used for the evolution of `alpha_s`.
> 
> Ref: https://lhapdf.hepforge.org/config.html

not quite - quoting said HTML page

> Maximum/fixed number of flavors used in the fit.

so this refers to the PDFs in the first place - so it should be  `theory_card.heavy.num_flavs_max_pdf` (of course 14 was just wrong)

> alpha_s evolution will use this value if AlphaS_NumFlavors is not explicitly set.

this is only a fallback

PS: @AleCandido I know you want to give up on that specific variable (did we do it already?), so we would need to provide an alternative

felixhekhorn:
> Ok for me but what about the `Flavors` list? It should be consistent with `NumFlavors` right?

No - again quoting from the officals:

> List of supported (user-visible) parton flavors, using the PDG Monte Carlo numbering scheme (quarks from 1 .. 6, antiquarks from -6 .. -1, and gluon = 21, as well as any "optional extras" such as photon = 22).

so you can stuff what ever there (also BSM) - `NumFlavors` refers to DGLAP

scarlehoff:
> so this refers to the PDFs in the first place

Oh, I missed the `.` I understood "on the fit evolution" as one object.

In any case, it is not 14 :P 

> Ok for me but what about the Flavors list? It should be consistent with NumFlavors right?

Not really since we are just filling with 0s the numbers we don't care about in `Flavors`.

src/ekobox/info_file.py
```diff
@@ -39,7 +39,7 @@ def build(
     template_info["Authors"] = ""
     template_info["FlavorScheme"] = "variable"
     template_info.update(info_update)
-    template_info["NumFlavors"] = 14
+    template_info["NumFlavors"] = theory_card.couplings.max_num_flavs
```
felixhekhorn, line 42:
```suggestion
    template_info["NumFlavors"] = theory_card.heavy.num_flavs_max_pdf
```

src/ekobox/info_file.py
```diff
@@ -39,7 +39,7 @@ def build(
     template_info["Authors"] = ""
     template_info["FlavorScheme"] = "variable"
     template_info.update(info_update)
-    template_info["NumFlavors"] = 14
+    template_info["NumFlavors"] = theory_card.heavy.num_flavs_max_pdf
```
AleCandido, line 42: Same as https://github.com/NNPDF/nnpdf/pull/1746#discussion_r1219959665
```suggestion
    template_info["NumFlavors"] = max(nf for _, nf in op.mugrid)
```

AleCandido: > PS: @AleCandido I know you want to give up on that specific variable (did we do it already?), so we would need to provide an alternative

This is the alternative. If really needed (also outside LHAPDF), we can make a property for this.

AleCandido: I have the feeling LHAPDF metadata are rather redundant, but maybe I'm wrong...

scarlehoff: Why using the loop if we already have the property?

AleCandido: > Why using the loop if we already have the property?

We don't have the property, the two things are not synced: one is a theoretical maximum declared in the `theory_card`, the other one is the one computed in practice, spelled out in the `op`erator card.
Moreover, the theoretical maximum is deprecated, since it is redundant with matching scales, so it is going to be removed in the next minor (i.e. breaking) release of EKO, v0.14.0.

AleCandido: Before applying, I'll wait for @felixhekhorn to take a look and confirm

scarlehoff: We should keep this in mind for the new theory database (whenever we actually do it) since there we do have a maximum set.

AleCandido: I hope that the new database (if ever) will be almost straightforward, after this round of polishing the pineline.

EKO has an almost minimal runcard, the more we refactor the internals, the more we clean it up. Same thing is going to happen to yadism, since I'd like to move the runcard translation to Pineko.
The moment we have minimal theories for each program, the theory database fields will be the union of the various runcards attributes (the only non-straightforward part is in Pinefarm, since we should propagate theory settings even to MadGraph and the other MCs).

andreab1997 (approved):

src/ekobox/info_file.py
felixhekhorn: I guess this is also fine ... this makes the info file slightly more static (you can not add dynamically more Q2 points later on), but I guess that feature might be never needed

PR #284, opened 2023-06-14
Use evolpoint in apply

### LINKED ISSUE #283: Fix `apply_pdf` so that `(Q, nf)` does not overwrite `(Q, nf-1)`.
As seen in https://github.com/NNPDF/nnpdf/pull/1754, currently when evolving a PDF eko will return an output `pdf(Q)`, while instead it should return `pdf(Q, nf)`.

felixhekhorn:
Closes #283 

(I only explicitly tested the LHA bench, but also all the other should be fine ...)

### DIFF
--- a/src/ekomark/benchmark/runner.py
+++ b/src/ekomark/benchmark/runner.py
@@ -252,7 +252,7 @@ class Runner(BenchmarkRunner):
             # find the first hit, regardless of the flavor, since others can not deal with them appearing in parallel
             eps = [ep for ep in pdf_grid.keys() if np.isclose(ep[0], q2)]
             if len(eps) != 1:
-                if len(eps) < 1:
+                if len(eps) == 0:
                     raise KeyError(f"PDF at Q2={q2} not found")
                 raise KeyError(f"More than one evolution point found for Q2={q2}")
             res = pdf_grid[eps[0]]
--- a/src/ekomark/benchmark/runner.py
+++ b/src/ekomark/benchmark/runner.py
@@ -4,6 +4,7 @@ import logging
 import os
 import sys
 
+import numpy as np
 import pandas as pd
 from banana import cfg as banana_cfg
 from banana.benchmark.runner import BenchmarkRunner
@@ -248,7 +249,13 @@ class Runner(BenchmarkRunner):
         for q2 in q2s:
             log_tab = dfdict.DFdict()
             ref_pdfs = ext["values"][q2]
-            res = pdf_grid[q2]
+            # find the first hit, regardless of the flavor, since others can not deal with them appearing in parallel
+            eps = [ep for ep in pdf_grid.keys() if np.isclose(ep[0], q2)]
+            if len(eps) != 1:
+                if len(eps) < 1:
+                    raise KeyError(f"PDF at Q2={q2} not found")
+                raise KeyError(f"More than one evolution point found for Q2={q2}")
+            res = pdf_grid[eps[0]]
             my_pdfs = res["pdfs"]
             my_pdf_errs = res["errors"]
 
--- a/src/ekobox/apply.py
+++ b/src/ekobox/apply.py
@@ -84,18 +84,18 @@ def apply_pdf_flavor(
 
     # build output
     out_grid = {}
-    for (mu2, nf), elem in eko.items():
+    for ep, elem in eko.items():
         pdf_final = np.einsum(CONTRACTION, elem.operator, pdfs, optimize="optimal")
         if elem.error is not None:
             error_final = np.einsum(CONTRACTION, elem.error, pdfs, optimize="optimal")
         else:
             error_final = None
-        out_grid[mu2] = {
+        out_grid[ep] = {
             "pdfs": dict(zip(eko.bases.targetpids, pdf_final)),
             "errors": None,
         }
         if error_final is not None:
-            out_grid[mu2]["errors"] = dict(zip(eko.bases.targetpids, error_final))
+            out_grid[ep]["errors"] = dict(zip(eko.bases.targetpids, error_final))
 
     # rotate to evolution basis
     if flavor_rotation is not None:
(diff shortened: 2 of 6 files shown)


src/ekomark/benchmark/runner.py
```diff
@@ -248,7 +249,13 @@ def log(self, theory, _, pdf, me, ext):
         for q2 in q2s:
             log_tab = dfdict.DFdict()
             ref_pdfs = ext["values"][q2]
-            res = pdf_grid[q2]
+            # find the first hit, regardless of the flavor, since others can not deal with them appearing in parallel
```
AleCandido, line 252: That's a bit sad, but of course correct...
(unless we start tricking them with multiple runs, but it could be a pain, definitely not for this PR)

src/ekomark/benchmark/runner.py
```diff
@@ -248,7 +249,13 @@ def log(self, theory, _, pdf, me, ext):
...
+            eps = [ep for ep in pdf_grid.keys() if np.isclose(ep[0], q2)]
+            if len(eps) != 1:
+                if len(eps) < 1:
```
AleCandido, line 255:
```suggestion
                if len(eps) == 0:
```
As specific as possible.

AleCandido (approved):
It should be alright, thanks

src/ekomark/benchmark/runner.py
felixhekhorn: > definitely not for this PR

agreed

scarlehoff:
Shouldn't this function be fixed as well?

https://github.com/NNPDF/eko/blob/e81f315cdeabcdd7ad4c7ce9201e0ffb00c578db/src/ekobox/genpdf/__init__.py#L219

I appreciate that `xfxQ` tries to follow the lhapdf signature but it feels weird having to use the information that is being passed (but ignored) to create a callable that will take into account the value of `nf`...


That said, I just tested this branch and seems to work as expected :) (fixing the call to `genpdf` in `evolven3fit`).

AleCandido:
> I appreciate that `xfxQ` tries to follow the lhapdf signature but it feels weird having to use the information that is being passed (but ignored) to create a callable that will take into account the value of `nf`...

By now, I start feeling weird anything related to LHAPDF grids. Since I believe EKO could produce more, I'm working on that side. In my mind, once the replacement will be available, LHAPDF grids will come by compression and conversion of potentially wider grids.

felixhekhorn:
> Shouldn't this function be fixed as well?
> 
> https://github.com/NNPDF/eko/blob/e81f315cdeabcdd7ad4c7ce9201e0ffb00c578db/src/ekobox/genpdf/__init__.py#L219
> 
> I appreciate that `xfxQ` tries to follow the lhapdf signature but it feels weird having to use the information that is being passed (but ignored) to create a callable that will take into account the value of `nf`...
> 
> That said, I just tested this branch and seems to work as expected :) (fixing the call to `genpdf` in `evolven3fit`).

Well, that's why I added this line
https://github.com/NNPDF/eko/blob/dc0e959451b665dac4bc235382485f37d2173d9d/src/ekobox/evol_pdf.py#L83
and then inside a single "block" there is (usually) only one nf

so either I hack the eko output or lhapdf 
https://github.com/NNPDF/eko/blob/dc0e959451b665dac4bc235382485f37d2173d9d/src/ekobox/genpdf/__init__.py#L69
and changing ourself is easier

scarlehoff:
Yes, sure, I understand the logic. It is just a bit awkward since right now you pass down (to `generate_block`) an `evolgrid` object which includes `(nf, Q)` but because your function only takes `Q` you need to remove the `nf` information inside:

https://github.com/NNPDF/eko/blob/e81f315cdeabcdd7ad4c7ce9201e0ffb00c578db/src/ekobox/genpdf/__init__.py#L223

but also outside!
https://github.com/NNPDF/eko/blob/dc0e959451b665dac4bc235382485f37d2173d9d/src/ekobox/evol_pdf.py#L83

My suggestion would be (if you want to keep the LHAPDF signature) to modify `generate_block` so that it takes as an input an array of `Q` and not `evolgrid`. That way it is explicit that `generate_block` doesn't know about `nf`. And since it is a block it shouldn't anyway.

(either way is done it can be worked with from the outside so don't bother with it if it makes the changes to the code too complicated, I don't know what other changes would be needed)

AleCandido:
> My suggestion would be (if you want to keep the LHAPDF signature) to modify `generate_block` so that it takes as an input an array of `Q` and not `evolgrid`. That way it is explicit that `generate_block` doesn't know about `nf`. And since it is a block it shouldn't anyway.

I agree with the argument of Juan, since we are thrashing `nf`, it's worth to do it in a single place, it makes it easier to maintain `generate_block`, because you're not in a mixed state (in which you know _part_ of `nf`, but you don't want to).

felixhekhorn:
> For me it's good to go, even before #274

I'm waiting for either @scarlehoff or @andreab1997 ... now #274 is sitting on top, so should be fine

scarlehoff:
> I'm waiting for either @scarlehoff

Let me test it in `evolven3fit_new`

src/ekomark/benchmark/runner.py
felixhekhorn: you mean the python comment? here we are comparing eko output against others (e.g. APFEL); now we can have a single Q2 have appear twice: `(4.92,4)` and `(4.92,5)` ... I have to check I'm not running in this situation and if I'm not there is only one hit and that is the one

AleCandido: Even 4 times :P
(i.e. even `(4.92, 3)` and `(4.92, 6)` are possible, in principle)

andreab1997: I understand, Thanks!

scarlehoff (approved):
This works fine with `evolven3fit_new` and produces the desired output :)

And has the secondary effect or simplifying the code in our side quite a bit actually.
Let me know when a tag is made with this changes in so I can push my changes to nnpdf

felixhekhorn:
> Let me know when a tag is made with this changes in so I can push my changes to nnpdf

tag is on the way: https://github.com/NNPDF/eko/releases/tag/v0.13.5

Issue #248, opened 2023-04-21
Next runcards update


alecandido:
At the moment we want to keep compatibility with the runcard format defined by #220

However, while reworking the internals for numerical FONLL I just defined the concept of the "evolution point", that (apart from the name) is a clean abstraction suitable for the runcard as well.

For this reason, I start here a list of desiderata for the next runcard update, without any idea about its schedule.

- [x] [evolution points](#evolution-points) in the runcard
- [x] drop [intrinsic range](#intrinsic-range)
- [ ] drop [integration cores](#integration-cores)
- [x] drop [max nf](#max-nf) configurations

And related:
- [x] drop [`runcards.update()`](#runcards-update)

### Evolution points

Move `num_flavs_init`:
https://github.com/NNPDF/eko/blob/0224739f3911509670b59b30db31c0bf2de5edcf/src/eko/quantities/heavy_quarks.py#L106-L111
to `mu0`:
https://github.com/NNPDF/eko/blob/0224739f3911509670b59b30db31c0bf2de5edcf/src/eko/io/runcards.py#L95-L96

Also the couplings reference should be an evolution point `(scale, num_flavs_ref)`:
https://github.com/NNPDF/eko/blob/0224739f3911509670b59b30db31c0bf2de5edcf/src/eko/quantities/couplings.py#L23-L25

### Intrinsic range

https://github.com/NNPDF/eko/blob/0224739f3911509670b59b30db31c0bf2de5edcf/src/eko/runner/parts.py#L45-L62
https://github.com/NNPDF/eko/blob/0224739f3911509670b59b30db31c0bf2de5edcf/src/eko/quantities/heavy_quarks.py#L114-L115

### Integration cores

After #242 we should not need any longer, and parallelism in Python has poor support, better avoiding it, and apply at some different level (dispatch multiple jobs, use parallel primitives).

https://github.com/NNPDF/eko/blob/0224739f3911509670b59b30db31c0bf2de5edcf/src/eko/io/runcards.py#L87-L88

### Max nf

Drop both `MaxNfPdf` and `MaxNfAs` because redundant with the matchings mechanism.

Actually, the first one is fully redundant, while the second is only redundant if equal to the first, but there is no evidence that it makes sense using it separately during an evolution run (and since EKO is a library, like APFEL, if you want just to evolve $\alpha_s$ you can pass your matchings to the `Couplings` class)

### Runcards update

It will be up to the user to use the new nested runcards.

[...]

PR #291, opened 2023-07-21
Add mypy

### LINKED ISSUE #178: Check with mypy in the CI
Since I'm transitioning code to type hints (or type hints friendly, at least), we can get more from type checks.
In any case, types are inferred, so there is a benefit in type checking also without any type hint.

I propose to add `mypy` check to our usual CI workflow. In the end, it costs nothing, and it is not even an extra dependency.

felixhekhorn:
Closes #178 

Current state is ~~`Found 63 errors in 17 files (checked 287 source files)`~~ - see CI


src/eko/evolution_operator/__init__.py
```diff
@@ -579,6 +579,9 @@ def quad_ker_qed(
     return ker
 
 
+OPMEMBERS = Dict[Tuple[int, int], OpMember]
```
AleCandido, line 582: This is a type hint, not a constant. The naming convention (i.e. the case) should be the same of classes (sorry to be pedantic...)

felixhekhorn: https://github.com/NNPDF/eko/pull/291/commits/494da3c0ccd3d081cc3e345f39e96457b0fd46a3

felixhekhorn:
some problems (accumulated so far):
1. `src/eko/io/struct.py:283: error: Incompatible return value type (got "tuple[Any, ...]", expected "tuple[float, int] | None")  [return-value]`: this is going through numpy and I don't know how to cast
2. `src/eko/evolution_operator/grid.py:156: error: "object" has no attribute "path"  [attr-defined]`: this is going through a dict and I don't know how to cast
3. `src/eko/runner/operators.py:21: error: Incompatible return value type (got "list[Operator | None]", expected "list[Operator]")  [return-value]` an `Inventory` does not know if it is `contentless` and so `__getitem__` might return `None` - however in practice here this is not allowed to happen
4. unit tests are failing, because I made `num_flavs_ref` mandatory, but `banana` does not provide it - shall we change there?

felixhekhorn:
With this PR we introduce mypy via pre-commit. This would be a further check which developers need to control.
- The changes are mainly backward compatible, BUT (as said above) the change to `num_flavs_ref` which is no longer an optional parameter, but a mandatory one. This reaches beyond eko, as banana would need to provide a value - if we are happy with that change, I can open a new PR there (else here we fail unit tests). (One could consider this even a breaking change - as already said: our no-default-strategy clashes with the public interface.)
- for the most part, I did the minimal required change to make mypy running - so #181 still remains to be done
- for the moment the check is restricted to `src/` at some point we can consider widening ...

benchmarks/__init__.py
AleCandido, line 1: I don't like so much this idea: this folder is a collection of scripts. It is not meant to be a package.

Can we avoid this?

benchmarks/ekobox/__init__.py
AleCandido, line 1: Same of above.

benchmarks/__init__.py
felixhekhorn: This was a left over from before mypy was restricted to `src/` (it was complaining about double names)

src/eko/quantities/couplings.py
```diff
@@ -22,7 +21,7 @@ class CouplingsInfo(DictLike):
     alphaem: Coupling
     scale: LinearScale
     max_num_flavs: FlavorsNumber
```
giacomomagni, line 23: It might be slightly off topic, can we drop `max_num_flavs` since it's not used anymore ?

src/eko/io/legacy.py
```diff
@@ -92,10 +100,16 @@ def from_old(cls, old: RawCard):
         """Load from old metadata."""
         heavy = HeavyInfo(
             num_flavs_init=4,
-            num_flavs_max_pdf=None,
-            intrinsic_flavors=None,
-            masses=HeavyQuarkMasses([1.51, 4.92, 172.5]),
-            masses_scheme=None,
+            num_flavs_max_pdf=5,
```
giacomomagni, line 103: On a similar point can we start dropping this?

src/eko/quantities/couplings.py
AleCandido: It's a breaking change, listed with others in #248

src/eko/io/legacy.py
AleCandido: Same as above

src/eko/quantities/couplings.py
felixhekhorn: yes, but we want to do it - however, maybe not in this PR, but in a new (it was also confusing in Como :upside_down_face: )

AleCandido:
> * for the most part, I did the minimal required change to make mypy running - so #181 still remains to be done

I'm also fine with closing #181, since Mypy will actually run. Or keeping it open, as a reminder to incrementally adopt types.

In any case, this is something that won't finish ever, since types will always be optional, no matter what (unless we ban `Any` in Mypy, but I still feel it's a bit too hardcore...).

src/eko/quantities/couplings.py
felixhekhorn: Actually, since I already introduced a breaking change here, we can even do it here ...

AleCandido: At this point, we are about to break, let's implement the whole #248.

It's not expensive implementation-wide, the problem was just breaking.

felixhekhorn: #295

src/eko/io/legacy.py
felixhekhorn: #295

felixhekhorn:
here
https://github.com/NNPDF/eko/blob/94846667aa6ed95899dacdf8d89edf8c8ad25378/src/eko/io/types.py#L23-L24
we should replace `int` with `Literal[3,4,5,6]`

AleCandido:
> we should replace `int` with `Literal[3,4,5,6]`

Ah, yeah, it could be stricter, but I didn't bother in restricting too much.

However, pay attention that `Literal[3,4,5,6]` might be too restrictive:
```py
from typing import Literal

def subtract(n: int, m: int) -> int:
    return n - m

flavor: Literal[3,4,5,6] = subtract(7, 2)
```
flags the following error:
```
ciao.py:8: error: Incompatible types in assignment (expression has type "int", variable has type "Literal[3, 4, 5, 6]")  [assignment]
```
i.e. `Literal` has to be a `Literal`, it can't be an integer with that value (the main point is that Mypy check does not happen at runtime, if you want to enforce this kind of bounds would be better to start using Pydantic).

src/eko/evolution_operator/grid.py
```diff
@@ -118,7 +127,7 @@ def get_threshold_operators(self, path: List[Segment]) -> List[Operator]:
         is_downward = is_downward_path(path)
         shift = flavor_shift(is_downward)
         for seg in path[:-1]:
-            new_op_key = astuple(seg)
+            new_op_key = seg
```
AleCandido, line 130: Great that you removed the `astuple()` call, but do we need the `new_op_key` variable any longer?

src/eko/io/legacy.py
AleCandido, line 1: I'd be happy to drop this module as well soon.

src/ekobox/apply.py
```diff
@@ -48,6 +52,14 @@ def apply_pdf(
 CONTRACTION = "ajbk,bk"
 
 
+@dataclass
+class PdfResult:
+    """Helper class to collect PDF results."""
+
+    pdfs: Dict[Union[int, str], float]
```
AleCandido, line 59: The union could be a `Label` or `Flavor` type.

src/ekobox/cli/inspect.py
```diff
@@ -27,7 +27,7 @@ def subcommand(ctx, path: pathlib.Path):
 @click.pass_obj
 def sub_mu2(operator: EKO):
     """Check operator's mu2grid."""
-    rich.print_json(data=operator.mu2grid.tolist())
+    rich.print_json(data=operator.mu2grid)
```
AleCandido, line 30: Wait, `mu2grid` is still an array, I'd like to really print as JSON (for post-processing, if ever).

Was there a typing problem?

src/eko/runner/operators.py
```diff
@@ -16,7 +16,9 @@ def retrieve(
     elements = []
     for head in headers:
         inv = parts if isinstance(head, Evolution) else parts_matching
-        elements.append(inv[head])
+        op = inv[head]
+        assert isinstance(op, Operator)
+        elements.append(op)
```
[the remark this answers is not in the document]
felixhekhorn: this is point 3 in [in my list](https://github.com/NNPDF/eko/pull/291#issuecomment-1645978114) (I hope you can read below the strike :upside_down_face: )

src/eko/evolution_operator/__init__.py
```diff
@@ -603,8 +607,8 @@ class Operator(sv.ModeMixin):
 
     log_label = "Evolution"
     # complete list of possible evolution operators labels
-    full_labels = br.full_labels
-    full_labels_qed = br.full_unified_labels
+    full_labels = list(br.full_labels)
```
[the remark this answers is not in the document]
felixhekhorn: because it is not of constant size between here and [it's child class](https://github.com/NNPDF/eko/blob/6e0f434b9c1792dad9c79bed9e013e2a5b5cdd1d/src/eko/evolution_operator/operator_matrix_element.py#L199)

src/eko/msbar_masses.py
```diff
@@ -395,7 +395,7 @@ def sc(thr_masses):
     heavy_quarks = quark_names[3:]
     hq_idxs = np.arange(0, 3)
     if nf_ref > 4:
-        heavy_quarks = reversed(heavy_quarks)
+        heavy_quarks = "".join([e for e in reversed(heavy_quarks)])
```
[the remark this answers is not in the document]
felixhekhorn: because `reversed` returns a special iterator which mypy doesn't accept for join

src/eko/io/legacy.py
felixhekhorn: let's discuss either in #295 or #296

src/ekobox/cli/inspect.py
felixhekhorn: apparently - else I wouldn't have touched it (as already said CLI is never used nor tested)

src/eko/couplings.py
```diff
@@ -474,7 +475,7 @@ def assert_positive(name, var):
                 self.decoupled_running,
             )
         # cache
-        self.cache = {}
+        self.cache: Dict[Tuple[float, float, int, float, float], npt.NDArray] = {}
```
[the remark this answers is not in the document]
felixhekhorn: done in 54d9e534

src/ekobox/apply.py
felixhekhorn: done in 9b0d8048

src/eko/msbar_masses.py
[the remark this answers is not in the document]
felixhekhorn: and `join` because must be a string as before

src/eko/runner/operators.py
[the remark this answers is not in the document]
AleCandido: Ok, you're right. I would have liked to make a generic constant for the `contentless`, but I thought it would have been too much for Python types (I should also derive the type from the const generic...).

Just to make slightly more explicit, check:
```suggestion
        op = inv[head]
        assert op is not None
        elements.append(op)
```
It will be sufficient to infer that is an `Operator`, and passing less information we'll rely more on the type checking.

src/eko/evolution_operator/operator_matrix_element.py
```diff
@@ -208,7 +209,7 @@ class OperatorMatrixElement(Operator):
         (br.matching_hminus_pid, br.matching_hminus_pid),
     ]
     # still valid in QED since Sdelta and Vdelta matchings are diagonal
-    full_labels_qed = copy.deepcopy(full_labels)
+    full_labels_qed: List[Tuple[int, int]] = copy.deepcopy(full_labels)
```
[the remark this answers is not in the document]
felixhekhorn: done in [16e3809](https://github.com/NNPDF/eko/pull/291/commits/16e38092e5eeb42d43348bb07c1d3417e7747a14) - I used `types` though ...

src/ekobox/cli/inspect.py
AleCandido: It seems like we (i.e. I) changed type to `operator.mu2grid`, but I forgot to update here. Thanks Mypy (and @felixhekhorn, of course)!

AleCandido (approved):

felixhekhorn:
@AleCandido it seems pylint does not consider `EKO.read` type safe (see the error before) and I couldn't convince him so I needed the big hammer in https://github.com/NNPDF/eko/pull/291/commits/a49fca63858f9e7101e8dc1a40a63e54606e9e20

AleCandido:
> @AleCandido it seems pylint does not consider `EKO.read` type safe (see the error before) and I couldn't convince him so I needed the big hammer in [a49fca6](https://github.com/NNPDF/eko/commit/a49fca63858f9e7101e8dc1a40a63e54606e9e20)

I'm sorry for it, but it is more of a Pylint problem, since the `EKO.read()` method as even an unneeded assertion just for the sake of the type (`EKO.open()` is the ambiguous one, which is returning a union, but the other one should be alright).

If you wish, you could also add a return hint, by using `-> "EKO"` (that would be `typing.Self` from py3.11 on).

felixhekhorn:
Note that mypy detected the bug of missing `import` in https://github.com/NNPDF/eko/pull/291/commits/4be6d44a2899629a375ee23b87e7d82c95e4f4a8 - so I'd say we really want this live

src/ekobox/evol_pdf.py
```diff
@@ -93,8 +97,6 @@ def evolve_pdfs(
         info_update=info_update,
     )
 
-    # in the eko scales are squared
-    q2block_per_nf = {nf: np.power(q2s, 2) for nf, q2s in q2block_per_nf.items()}
```
felixhekhorn, line 97: this removal is a very suspicious change ... because I introduced it in #391 (d5f63fe7ac995fd9d0d38dba48f2a7969de01c28 to be specific) and [unit tests were passing](https://github.com/NNPDF/eko/actions/runs/9956298042/job/27505888734#step:11:72) - but I have to remove it here to make [the unit tests pass](https://github.com/NNPDF/eko/actions/runs/10301208359/job/28512239152?pr=291#step:11:73) and I can not see how the other changes here would change that ...

PR #292, opened 2023-07-24
Change `manipulate` to act on Operator

### LINKED ISSUE #203: Manipulate iterators
At the moment, it is only possible to apply `manipulate` functions on the disk object, modifying it.

Sometimes, it is definitely useful to be able to get the manipulation in memory, but this of course has to be done one element at a time (otherwise we won't reduce the memory consumed by EKO).

Thus, we want to reshape and yield one by one:

```py
def xgrid_reshape_iter(...):
    for mu2 in mu2grid:
        yield reshape(eko[mu2])
```

felixhekhorn:
Closes #203 

There is no need for additional iterators as `EKO` is already providing all iterators needed. The only necessary change is to let `manipulate` act on `Operator` instead of an whole `EKO`.

I guess, we consider this as a part of our public API, so this will trigger a new version. This might also have an effect on evolven3fit (cc @scarlehoff )

As a nice side effect, we also remove a bug discovered in #291 : when activating `plot_operators` in `ekomark` we were crashing, since there was a `manipulate` and this confused eko afterwards


src/eko/io/manipulate.py
```diff
@@ -18,10 +19,8 @@
 SIMGRID_ROTATION = "ij,ajbk,kl->aibl"
 """Simultaneous grid rotation contraction indices."""
 
-Basis = Union[XGrid, npt.NDArray]
 
-
-def rotation(new: Optional[Basis], old: Basis, check: Callable, compute: Callable):
+def rotation(new: Optional[XGrid], old: XGrid, check: Callable, compute: Callable):
```
giacomomagni, line 23: If I understand correctly this function can be simplified a bit more. Since `newtarget` and `newinput` are not used anymore we can return only the second element. Also the `Optional` for `new` can be dropped.

src/ekomark/plots.py
```diff
@@ -244,9 +249,16 @@ def save_operators_to_pdf(path, theory, ops, me: EKO, skip_pdfs, change_lab=Fals
...
-            results = me[mu2].operator
-            errors = me[mu2].error
+        for mu2, op in me.items():
+            change_lab = False
+            if rotate_to_evolution_basis:
+                if not qed:
+                    op = manipulate.to_evol(op, source=True, target=True)
+                else:
+                    op = manipulate.to_uni_evol(op, source=True, target=True)
+                change_lab = True
```
giacomomagni, line 259: `change_lab ` is activated only if `rotate_to_evolution_basis` is, so can be removed

src/ekomark/plots.py
```diff
@@ -209,7 +210,15 @@ def plot_operator(var_name, op, op_err, log_operator=True, abs_operator=True):
...
 
-def save_operators_to_pdf(path, theory, ops, me: EKO, skip_pdfs, change_lab=False):
+def save_operators_to_pdf(
+    path,
+    theory,
+    ops,
+    me: EKO,
+    skip_pdfs,
+    rotate_to_evolution_basis: bool,
+    qed: bool = False,
```
giacomomagni, line 220: Docstrings are not updated anymore

felixhekhorn: done in https://github.com/NNPDF/eko/pull/292/commits/4699064f9e4c5e368e148f1afb9614e0e9908689

src/ekomark/plots.py
felixhekhorn: done in https://github.com/NNPDF/eko/pull/292/commits/4699064f9e4c5e368e148f1afb9614e0e9908689

src/eko/io/manipulate.py
felixhekhorn: - returning one element is done in https://github.com/NNPDF/eko/pull/292/commits/4699064f9e4c5e368e148f1afb9614e0e9908689
- `Optional` we still need, since we're calling the function always on both sides, but one might me `None`

giacomomagni (approved):

felixhekhorn:
> Since this is already breaking, I decided to put another breaking change on top: [3e2d12a](https://github.com/NNPDF/eko/commit/3e2d12a4ff0a1013ce33cd07b193a2c52c9e4939) drops the `Bases` class, i.e. metadata does not know about different bases (x,pid) in input or output any longer, which is consistent with the main change introduced in this PR

this addresses the first item of #205

PR #295, opened 2023-07-31
Next runcards update

### LINKED ISSUE #248: Next runcards update

felixhekhorn:
Closes #248 

#291 is already big enough on it's own and I expect even more changes here, because we're breaking a lot of stuff so I decided for a PR on its own ...

I won't drop `n_integration_cores` for the moment, because we are still missing the replacement: the actual jets runner

- [x] rerun all noteboooks
- [x] fix unit tests


AleCandido:
> I won't drop `n_integration_cores` for the moment, because we are still missing the replacement: the actual jets runner

I was not planning to write an "actual" jets runner in EKO ever. But if needed I can contribute the utilities for batching to Pineko.

What were your plans about it?

(I'd like, if possible, to drop even `n_integration_cores` before next release, to avoid another breaking one anytime soon)

src/ekomark/data/__init__.py
```diff
@@ -1 +1,29 @@
 """EKO database configuration."""
+from typing import Union
+
+from eko.io.runcards import Legacy, OperatorCard, TheoryCard
+from eko.io.types import RawCard
+
+
+def update_runcards(
```
AleCandido, line 8: That's a good idea. But maybe, at this point, the whole `legacy.py` should be moved here, since it should not be directly used in `eko`.

felixhekhorn: Which one do you mean? `eko/io/legacy.py` or `eko/runner/legacy.py`? The latter I guess we can move, since it is currently only used by ekomark (but there it is used, so we can not completely drop it) and it is deprecated in 0.13. The former, I guess we can't because that we need to convert ekos ... (speaking of I'll comment below)

felixhekhorn:
current status:
- drop max nf configurations: they were already not used, so here the impact is mild
- drop runcards.update(): for us this is just a renaming, so again the impact should be mild, but this might effect the outside world ...
- drop integration cores: as said, I won't do here since we're lacking an alternative at the moment - this is addressed in #296 and subsequent PRs in pineko (cc @andreab1997 )
- evolution points in the runcard: here the impact is stronger, I'd say, especially since this is no doubt public API - we even shuffled stuff from theory to operators ($n_{f0}$)
- drop intrinsic range: this had a huge impact, since I rooted it out at the bottom

all in all, this PR is very big (look at the number of files and we're impacting `eko` at it's core. This will require some extended testing for which I'd like to ask some help. We also need to decide if and what to do for the transition of the files from 0.13 to 0.14, which is a breaking change.

(I also need to fix LHA, but I believe that should be not too complicated)

felixhekhorn:
I believe [the unit test failing in #291](https://github.com/NNPDF/eko/actions/runs/5657890221/job/15327989509?pr=291#step:11:140) is protected here by https://github.com/NNPDF/eko/blob/7cf9e1b30d2f17553094e23db15775aabc8e9545/src/eko/io/runcards.py#L216-L218

src/ekomark/data/__init__.py
AleCandido: Actually, both. Do we really need to keep converting EKOs?

giacomomagni (commented):
Can we also drop this? 

https://github.com/NNPDF/eko/blob/7cf9e1b30d2f17553094e23db15775aabc8e9545/src/eko/matchings.py#L82

src/ekomark/data/__init__.py
felixhekhorn: This I can't answer, but we should ask to @niclaurenti @andreab1997 @giacomomagni

felixhekhorn:
> Can we also drop this?
> 
> https://github.com/NNPDF/eko/blob/7cf9e1b30d2f17553094e23db15775aabc8e9545/src/eko/matchings.py#L82

Why? this is just an alternative constructor - which is handy at times ... (and there is no max nf business around - it just sets up correctly the thresholds ...)

src/ekomark/data/__init__.py
andreab1997: I don't believe so

giacomomagni: Hopefully we won't

felixhekhorn: if needed can be discussed in #328

PR #296, opened 2023-08-01
Improve runner API

alecandido:
The idea is to allow the definition of a suitable runner in Pineko, to make use of the `Q2` granularity, and eventually drop `n_integration_cores` in EKO.
If we can suitably split at the level of `Q2`, there is no need to parallelize at the level of `x`, since our `xgrid` is not going to grow wild. Avoiding parallelization will simplify a bit the content of the `eko` package, and it will solve many problems encountered with Python parallelization.

To allow this, the following steps are required:
- [ ] expose a high level API from `eko.runner`, such that it will be reliable outside the package itself (this will be the public runner API) -> postponed to #399
- [x] load opened EKOs, to avoid the need of merging separately computed `Q2` values, for which we only need
    - to avoid the use of the context manager (such that the folder won't be deleted)
    - [x] to specify a given path for the folder, to be accessed even outside EKO (if `None`, the default, use a temporary one, as it is currently done)
    - [x] a utility wrapper, to create the object from the folder

In a later PR, a merge utility can be added nevertheless, taking into account the scenario in which a single EKO is initialized, and copied over for parts and operators computation.
In particular, this limited `merge` utility will take two EKO tar files, and just copy the content of `parts/` and `operators/` subfolders of the second into the first one (in-place operation, to save disk - to implement the immutable operation just copy the first EKO tar beforehand).


AleCandido:
I'm completing the tests and the review of the runner API. The rest should be already alright (provided that tests are still passing).

To make @felixhekhorn happier: I dropped `open(mode="...")`, now we only have `read`/`write` (and `edit`, which is now just an alias for `read(..., readonly=False)`). But the `Builder` is still there, it's still useful :P

src/eko/io/struct.py
```diff
@@ -327,115 +326,109 @@ def deepcopy(self, path: os.PathLike):
...
     @classmethod
-    def read(cls, path: os.PathLike):
-        """Read the content of an EKO.
+    def read(
+        cls,
+        path: Path,
+        extract: bool = True,
+        dest: Optional[Path] = None,
+        readonly: bool = True,
+    ):
```
felixhekhorn, line 380: This looks to complicated ... can we make `from_tar` and `from_dir` instead (if you wish keep this low-level) and declare `extract` public?

AleCandido: First, I'd like to avoid the `from_` constructors altogether. There is a single EKO format, that's why `read` is useful.

Then, the EKO format is simple, there is nothing so complicate. If you want to extract you do not need `EKO.extract("a.tar", "b/")`, but you can go directly for `tar = tarfile.open("a.tar"); tar.extractall("b/")`. If we are just using the standard library for some operation, there is no need to hide it.
In practice, I expect `EKO.read("a.tar", dest="b/"))` to be used exactly in our use case: you want to open a folder in a controller for many agents to act on it. In principle, it could be even inside the context manager, if you're spawning the process from there, or even just waiting for their completion (and it would be recommended, at least someone will eventually close it).

So, no need for `EKO.extract()` in any case.
The `from_tar` is just `read`, because most of the time you will just read a tar, and nothing else. So it's only a renaming (but a name I don't like, gives the impression of many possible formats, but runtime and disk representation are coupled, so it's not really possible).
The `from_dir` is just an alias for `read(..., extract=False)`, so it's the exact same game of `edit`. However, if we play this game for both of them, we'll have to provide the `from_dir_edit`, and this I won't do. And I'd advocate for `edit`, rather than `from_dir`, since `edit` we are already using plenty of times.

AleCandido: Sorry, I forgot one point.

> Re `extract`: I would keep the function (about the content I'm open) just in case we change the compression at some point (that being said the new function should maybe be called `from_archive`?)

There is no compression: `tar` is only there to make a single blob, but the EKO is the folder inside (that's why we kept the `.tar`).
The compression is at the level of the arrays, and it's internally managed by EKO. So this should never affect the end user directly.

AleCandido:
Since I'm reviewing the runner API: may I drop the `legacy` runner?

@felixhekhorn @giacomomagni @niclaurenti @andreab1997

AleCandido:
I completed the review of the `eko.runner` API as well, now everything is up for discussion.

But before merging I'd like to have a plan for the *legacy* runner removal. This whole PR was meant to provide tools for Pineko to implement a further runner, similar to the *managed* one, but more sophisticated. There should be no reason to keep the *legacy* any further.

AleCandido:
> Looks fine. However, since this is surely part of the v0.14 series we should start merging the related PRs (order to be determined), since we have now several PRs touching these files (e.g. #291 )

I was already on it

Issue #146, opened 2022-09-05
Allow different orders in different parts?


felixhekhorn:
we're currently not able to redo the IC analysis out of the box, since we can not increase the matching order without increasing simultaneously the evolution order (which now is in place). So maybe we should
- allow a different order in matching then evolution (to redo IC)
- allow a different order in beta function (to benchmark CT14lo)

this would introduce 3 orders for the 3 usages we have

What do you think? @giacomomagni @AleCandido

AleCandido:
That things are getting complicated...

Going down this road we're approaching more and more the "new scales to be varied" pattern...

giacomomagni:
I'm wondering if this is really needed or we can just wait someone to ask for these features ...
This will complicate the a bit the situation which is already intricate since we are adding `QED`, 
and give space to run eko with "inconsistent" settings. (For IC we used that option 
because we didn't want to wait for full N3LO, not because we believe this is better)

felixhekhorn:
This is a generalization of #75

felixhekhorn:
Moved to https://github.com/NNPDF/eko/wiki/List-of-Possible-Future-Features

PR #299, opened 2023-08-09
Les Houches benchmark study 23

felixhekhorn:
Let's do the Les Houches stuff here in this branch (which is based on 0.13)

- I started drafting the NNLO benchmark, please have a look - how much do we search for precision? i.e. are we fine with our defaults?
- Commit 36049b323e63e86da8b4d50b919b9e6f61932222 fixes a bug and contains the necessary fix needed here - if we decide to not merge this PR, this should survive (you may call the change breaking, but I wonder how much in practice)


AleCandido:
>  * I started drafting the NNLO benchmark, please have a look - how much do we search for precision? i.e. are we fine with our defaults?

I'd say so: LO is perfect, and the usual accuracy is quite fine. However, we could dump tables as heatmaps (to resemble the same shape of published tables), with a log color-scale, and check if we are more or less fine, attempting a minimal informal quantification of what *fine* does mean (like *most* points below `1e-X`, where most might be 95% and X=3).

>  * Commit [36049b3](https://github.com/NNPDF/eko/commit/36049b323e63e86da8b4d50b919b9e6f61932222) fixes a bug and contains the necessary fix needed here - if we decide to not merge this PR, this should survive (you may call the change breaking, but I wonder how much in practice)

Fine by me, I'm also planning to drop usage of `apply_pdf` in `evolven3fit_new` https://github.com/NNPDF/nnpdf/issues/1786, and there are not many other users out there.

giacomomagni:
Thanks for drafting it. Should I use this branch to include the update on the 2 splitting functions, maybe a different one is neater?

AleCandido:
It's more or less fine both ways: you can do it here, or, if you prefer to keep things separated, do it in another branch, and then we'll rebase this on the new one :)

felixhekhorn:
> Should I use this branch to include the update on the 2 splitting functions, maybe a different one is neater?

For the private stuff we need to use a private fork (e.g. based on this branch)

giacomomagni:
Ah good point that stuff is private...

AleCandido:
Ah, you're right, we even discussed it. But I forgot immediately the privacy issue...

giacomomagni:
Can we merge this?

AleCandido:
Fine by me: the exercise code is useful to have in `master`, since it has already been used to produce results, and everything else is just a small net improvement (fewer `qed` params around make me happier).

Unless @felixhekhorn is planning to modify the code of the scripts any time soon, I'd be in favor of merging.

felixhekhorn:
fine merging it ... but I expect we will at some point need an extension, e.g. PTO is hardcoded here ... how about your scripts @giacomomagni ? (like the "actual" N3LO comparison plot)

giacomomagni:
they are all in the private eko, because of FHMV https://github.com/NNPDF/eko_fhmv_private/pull/1

giacomomagni:
I think this is the best place where to introduce the `PTO_match` which is going to differentiate the order of the matching conditions.

AleCandido:
Fine to introduce the matching order on top of this PR, but could you just split in a separate one? (just pointing to this)

giacomomagni:
> Fine to introduce the matching order on top of this PR, but could you just split in a separate one? (just pointing to this)

I think this feature is actually need also for the LH benchmarks as in the last meeting we agreed in VFNS N3LO with NNLO matching. Moreover all the branches of `eko_private` are based on this and the required changes of this new feature are not that complex.

felixhekhorn:
> I think this is the best place to introduce `PTO_match` which is going to differentiate the order of the matching conditions.

just to tag the corresponding issue: this is the first part of https://github.com/NNPDF/eko/issues/146

AleCandido:
> I think this feature is actually needed also for the LH benchmarks as in the last meeting we agreed in VFNS N3LO with NNLO matching. Moreover all the branches of `eko_private` are based on this and the required changes of this new feature are not that complex.

My point was not that we do not need this feature here, but rather the opposite: we do not need the benchmark for the feature.
So, the moment the feature is completed, we can even merge it before the benchmark (rebasing on `master`, and rebasing this on `master` after the merge).

I'm just trying to keep PRs as small as possible, to simplify the review process and merge faster (such that we also have more atomic changes in `master`).

giacomomagni (approved):
Can we merge this ?

src/eko/io/runcards.py
```diff
@@ -49,6 +49,8 @@ class TheoryCard(DictLike):
     """Ratio between factorization scale and process scale."""
     n3lo_ad_variation: N3LOAdVariation
     """|N3LO| anomalous dimension variation: ``(gg_var, gq_var, qg_var, qq_var)``."""
+    matching_order: Order
```
AleCandido, line 52: Maybe here it would even be fine to have a default
```suggestion
    matching_order: Optional[Order] = None
```

Such that by default we consider it equal to the evolution order.
(instead of doing it only for the old-style runcards)

src/eko/runner/parts.py
```diff
@@ -88,6 +88,7 @@ def evolve_configs(eko: EKO) -> dict:
         n_integration_cores=ocard.configs.n_integration_cores,
         ModSV=ocard.configs.scvar_method,
         n3lo_ad_variation=tcard.n3lo_ad_variation,
+        matching_order=tcard.matching_order,
```
AleCandido, line 91:
```suggestion
        matching_order=tcard.matching_order if tcard.matching_order is not None else tcard.order,
```

I always wondered if there is an easier way for the pattern `x if x is not None else y`. But the only thing it comes to my mind is `x or y`, and it is bad because for `x = False` it would return `y`...

src/eko/runner/legacy.py
```diff
@@ -77,6 +77,7 @@ def __init__(
             couplings=cs,
             interpol_dispatcher=bfd,
             n3lo_ad_variation=new_theory.n3lo_ad_variation,
+            matching_order=new_theory.matching_order,
```
AleCandido, line 80: Same of the comment for the new runner.

AleCandido (approved):
Just one comment, but the PR is practically ready to be merged.

src/eko/io/runcards.py
giacomomagni: I'm fine with and without default. @felixhekhorn are you happy with the dafult?

felixhekhorn: I'm fine with either ... as I always say eventually it may turn out impossible to keep "no defaults" and stable API - this is a clear example why: it is backward compatible, but we introduce defaults or it is a breaking change with no defaults

giacomomagni: well but here the default can also make sense, let's go for it.

AleCandido:
Actually @felixhekhorn is the author of the PR, so he can not approve. And I already approved (@giacomomagni as well).

Shall we just merge this?

src/eko/runner/parts.py
giacomomagni: Actually here I think we have introduced a bug.
`matching_order`  follows `PTO` (ie is 2 for NNLO) so here we should subtract one, once doing the default, no?

felixhekhorn: just to have easier tracking: moved to #322

PR #301, opened 2023-08-16
fix large N NS  N3LO ad limit

giacomomagni:
This PR introduce an (hopefully) minor fix in: 

* [x] The large N limit of the Non Singlet approximation. 
   In eq 2.17 of [ref](https://arxiv.org/pdf/1707.08315.pdf) we are replacing the large-N expansion of the harmonic sum with 
   the actual harmonic:

   $$\ln(N) + \gamma_E + \frac{1}{2N} \to S_1(N)$$

   thus the large-N reads: 

   $$\gamma_{ns} \approx A_4 S_1(N) - B_4 + C_4 \frac{S_1(N)}{N} - D_4 \frac{1}{N} $$

   This should not have any major impact at the PDF level (to be verified..), but might be relevant for the benchmark study.

* [x] The $n_f^2$ part of the Non Singlet and Pure Singlet is now parametrised according to the analytic result [ref](https://arxiv.org/abs/2308.07958)(constructed with large number of moments). 
The latter not being used directly for simplicity and due to the presence of weight 5 harmonic sums (NS) and unphysical pole at `N=2` in the PS which was overcomplicating units tests.


doc/source/theory/Mellin.rst
```diff
@@ -56,7 +56,7 @@ Plus Distributions
 ^^^^^^^^^^^^^^^^^^
 
 .. math ::
-    \mathcal M[1/(1-x)_+](N) = S_1(N)
+    \mathcal M[1/(1-x)_+](N) = - S_1(N - 1)
```
AleCandido, line 59: The minus sign is fine. I was not sure about the `N - 1`, but after checking I realized that is coming from the `N - 1` in the Mellin transform (`N - 2` after division, again `N - 1` after integration).

It should be correct, well-spotted!

felixhekhorn:
If we want this to be part of the v0.13 series we need to merge soon ...

doc/source/theory/Mellin.rst
felixhekhorn: This is a typo indeed

AleCandido:
> If we want this to be part of the v0.13 series we need to merge soon ...

Agreed, if we wait enough it won't make any sense any longer (but I hope this could be sorted out at most in September).

felixhekhorn (approved):
@giacomomagni if you're happy please merge (if not I guess we can keep a bit longer)

giacomomagni:
Let me try to add some results from https://arxiv.org/abs/2308.07958 as well. 
i.e. fix the nf^2 parts for NS and PS.

src/ekore/anomalous_dimensions/unpolarized/space_like/as4/gnsm.py
```diff
@@ -55,6 +55,8 @@ def gamma_ns_nf3(n, cache):
 def gamma_nsm_nf2(n, cache):
     r"""Return the parametrized valence-like non-singlet part proportional to :math:`nf^2`.
 
+    From :cite:`Moch:2017uml` ancillary files.
```
felixhekhorn, line 58: Is this the new stuff (then the cite is wrong) or catchup from old?

felixhekhorn (approved):

PR #309, opened 2023-10-05
propagate inverse inside Matching

giacomomagni:
This PR fixes the inverse matching inside the `Matching` dataclass.

### DIFF
--- a/src/eko/io/items.py
+++ b/src/eko/io/items.py
@@ -86,17 +86,17 @@ class Matching(Header):
 
     scale: SquaredScale
     hq: FlavorIndex
-    inverse: bool = False
+    inverse: bool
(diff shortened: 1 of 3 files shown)


giacomomagni:
[00937e1](https://github.com/NNPDF/eko/pull/309/commits/00937e12405437582cc68f5f7d9ffc11e7644286) should provide a working fix.

felixhekhorn (approved):
2f403e33 fails in master, but works here.

I'll leave the final say to @AleCandido since he's more familiar with this

AleCandido (approved):
It is exactly how I would have done it (though the test is only at recipe level - even sufficient, but not the strongest).

PR #311, opened 2023-10-06
Interface QED and N3LO

niclaurenti:
This pull request aim is to make QED and N3LO work together.


AleCandido:
> This pull request aim is to make QED and N3LO work together.

Definitely a nice target :P

AleCandido:
If on your way you happen to find some `qed` branching that could be reconciled with the main flow, even better!

PR #312, opened 2023-10-09
Fix `nf` in inverse matching

giacomomagni:
Somehow the `nf` passed during the inverse matching is not correct: 

see: 

https://github.com/NNPDF/eko/blob/50f108f05bbebef263e2d301f2c4fbdd5bbdf3d5/src/eko/runner/parts.py#L152

```
[2] > /data/theorie/gmagni/N3PDF/eko_fhmv_private/src/eko/runner/parts.py(155)match()
-> nf_match = op.nf - 1 if recipe.inverse else op.nf
(Pdb++) recipe
Matching(scale=2.2801, hq=4, inverse=True)
(Pdb++) op.nf
3
```

This can be fixed by setting always: 

```
 nf_match = op.nf
```


AleCandido:
Could you also provide a unit test?

AleCandido:
> This can be fixed by setting always:
> 
> ```
>  nf_match = op.nf
> ```

Btw, at this point it should not be worth to define `nf_match` at all. Just use `op.nf`.

benchmarks/eko/benchmark_iveverse_matching.py
felixhekhorn, line 1: Please fix the name also here :upside_down_face:

src/eko/runner/parts.py
```diff
@@ -150,9 +150,8 @@ def match(eko: EKO, recipe: Matching) -> Operator:
     op.compute()
 
     binfo = blowup_info(eko)
-    nf_match = op.nf - 1 if recipe.inverse else op.nf
     res, err = matching_condition.MatchingCondition.split_ad_to_evol_map(
-        op.op_members, nf_match, recipe.scale, **binfo
+        op.op_members, op.nf, recipe.scale, **binfo
```
giacomomagni, line 154: Actually, I was wandering if this could be a symptom of a deeper bug, something as the operator not multiplied with the correct order for backward evolution...

AleCandido (approved):
If there is no brilliant idea for any unit test in the short term, the benchmark is definitely good enough.

We can merge.

giacomomagni:
> If there is no brilliant idea for any unit test in the short term, the benchmark is definitely good enough.
> 
Thanks, indeed I had not a brilliant idea for a unit test, as this bug was mostly visible at PDF level.

felixhekhorn:
> > If there is no brilliant idea for any unit test in the short term, the benchmark is definitely good enough.
> 
> Thanks, indeed I had not a brilliant idea for a unit test, as this bug was mostly visible at PDF level.

I do have some idea (call `match` with mocked `ome.OME` and check the outcome), but I can do in a separate PR

PR #315, opened 2023-10-11
Add new result for ` nf^2` N3LO `P_gq` and improve `P_qqps` 

giacomomagni:
This PR is to include results from https://arxiv.org/pdf/2310.01245.pdf


felixhekhorn (approved):

giacomomagni:
I attach here the comparison with the analytic result in x-space:

Linear-x, relative difference: 
[pgq_rel_diff.pdf](https://github.com/NNPDF/eko/files/12869694/pgq_rel_diff.pdf)
[pqq_ps_rel_diff.pdf](https://github.com/NNPDF/eko/files/12869695/pqq_ps_rel_diff.pdf)

Log-x, absolute plot, in yellow the old approximation based on 5/10 moments.
For `P_qq_ps` the approximated result (in green) is not matching exactly with the analytic result
(as it happens for `P_gq`) because the small-x limit can't be imposed directly during the fit 
as it will screw up the large-x suppression (1-x).
However the difference between green an blue is negligible to any practical purpose and
green mostly solves the discrepancy pointed out in https://arxiv.org/pdf/2308.07958.pdf

[pgq_logx.pdf](https://github.com/NNPDF/eko/files/12869701/pgq_logx.pdf)
[pqq_ps_logx.pdf](https://github.com/NNPDF/eko/files/12869702/pqq_ps_logx.pdf)

felixhekhorn:
I'm a bit confused ... can you please say again which color refers to what approximation? (in either case)

giacomomagni:
> I'm a bit confused ... can you please say again which color refers to what approximation? (in either case)

For the log plot:

green: new approximation based on the analytic result
blue: analytic result
yellow: approximation based on 5/10 moments

Not that for P_gq blue and green overlaps.

In the linear plot there is only the relative difference "green-blue/green"

giacomomagni:
I realised there was an easy fix also for this last tiny detail of `P_qqps` at small-x: keeping the exact LL power and suppress the NLL only subtracting `1/n`. 
Now it matches spot on with the analytic. 


[pqq_ps_logx_new.pdf](https://github.com/NNPDF/eko/files/12884318/pqq_ps_logx_new.pdf)

I'm happy to merge.

felixhekhorn:
good! do it! :upside_down_face:

PR #316, opened 2023-10-12
Interface QED and N3LO, n.2

niclaurenti:
reopening #311 after rebasing master

### DIFF
--- a/src/eko/basis_rotation.py
+++ b/src/eko/basis_rotation.py
@@ -275,7 +275,7 @@ map_ad_to_unified_evolution = {
 }
 
 
-def ad_projector(ad_lab, nf, qed=False):
+def ad_projector(ad_lab, nf, qed):
     """
     Build a projector (as a numpy array) for the given anomalous dimension sector.
 
@@ -355,7 +355,7 @@ def select_light_flavors_uni_ev(ad_lab, nf):
         return map_ad_to_evolution[ad_lab]
 
 
-def ad_projectors(nf, qed=False):
+def ad_projectors(nf, qed):
     """
     Build projectors tensor (as a numpy array), collecting all the individual sector projectors.
 
--- a/src/ekobox/apply.py
+++ b/src/ekobox/apply.py
@@ -12,7 +12,6 @@ def apply_pdf(
     lhapdf_like,
     targetgrid=None,
     rotate_to_evolution_basis=False,
-    qed=False,
 ):
     """
     Apply all available operators to the input PDFs.
@@ -34,6 +33,7 @@ def apply_pdf(
         out_grid : dict
             output PDFs and their associated errors for the computed mu2grid
     """
+    qed = eko.theory_card.order[1] > 0
     if rotate_to_evolution_basis:
         if not qed:
             rotate_flavor_to_evolution = br.rotate_flavor_to_evolution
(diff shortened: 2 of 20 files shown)


niclaurenti:
Now this is more or less ready. @AleCandido I wasn't able to reconcile the qed branches as you asked. I think that the uglier branch is the one happening in https://github.com/NNPDF/eko/blob/f39b012e6436d8bf5e1321b5f5e20c398e5191f6/src/eko/evolution_operator/__init__.py#L273
but it is quite complicated to change it since qcd and qed evolution require different ingredients and also the structure is a bit different(in one case we have singlet + non_singlet, in the other also the valence). If you have proposals we can try to restructure it a bit.

Then, whenever we will tag the next eko version, we have to pay attention to the fact that i remover the argument `qed` from `apply_pdf` (that information now is contained in the `EKO` that is passed to the function). The only part that is affected should be
https://github.com/NNPDF/nnpdf/blob/db8b790c1bacf52f71cc3a7fabf0fd9674cbf476/n3fit/src/evolven3fit_new/evolve.py#L178
but maybe some other parts of the `NNPDF` code are using this function (maybe @andreab1997 knows better)

src/ekobox/apply.py
```diff
@@ -34,6 +33,7 @@ def apply_pdf(
         out_grid : dict
             output PDFs and their associated errors for the computed mu2grid
     """
+    qed = eko.theory_card.order[1] > 0
```
giacomomagni, line 36: here and in the following don't you need to cast to bool? 
```suggestion
    qed = bool(eko.theory_card.order[1] > 0)
```

src/eko/basis_rotation.py
```diff
@@ -355,7 +355,7 @@ def select_light_flavors_uni_ev(ad_lab, nf):
         return map_ad_to_evolution[ad_lab]
 
 
-def ad_projectors(nf, qed=False):
+def ad_projectors(nf, qed):
```
giacomomagni, line 358: Are these functions still used somewhere?

giacomomagni (commented):
Looks good to me: 

* `to_flavor_basis_tensor` still contains a `qed=False` as default which can be dropped?

src/eko/basis_rotation.py
niclaurenti: I don't think so

src/ekobox/apply.py
niclaurenti: I can do it but I think it will not change anything: `order[1] > 0` is already a bool

giacomomagni: yes true, stupid me

niclaurenti:
> Looks good to me. `to_flavor_basis_tensor` still contains a `qed=False` as default which can be dropped?

Thanks I missed it. I will fix it asap

src/eko/basis_rotation.py
AleCandido: We always said that EKO is a library in the first place, and not an application. So, not all the functions have to be used.

For as long as they are correct, and they do not introduce much extra burden, I would keep them. It was non-trivial to get the normalization right (maybe because we're not especially clever, but to us, it was confusing).

They are just part of the *basis rotation toolbox*.

giacomomagni: okay, it make sense, let's keep them.

felixhekhorn (approved):
Did you check that the apfel benchmark is still fine? else this seems alright!

niclaurenti:
> Did you check that the apfel benchmark is still fine? else this seems alright!

@felixhekhorn The apfel bench at order = (3, 2) is still good. However, I cannot test order=(4,2) since apfel doesn't implement N3LO (I think)

PR #317, opened 2023-10-12
N3LO P_gg P_gq 5th moment

giacomomagni:
Update N3LO P_gg and P_gq using 5th moment from https://arxiv.org/abs/2310.05744

See also: https://github.com/NNPDF/eko_fhmv_private/pull/2

I checked that the provided scripts matches 100% with the ones shared with us by Moch.


felixhekhorn (approved):
We know already that this is working, right?

giacomomagni:
> We know already that this is working, right?

It should.

 Both approximation are based on the interpolated fifth moment, I want to use the analytic one
then we can merge.

Issue #322, opened 2023-11-07
Is `matching_order` correctly determined?


felixhekhorn:
Actually here I think we have introduced a bug.
`matching_order`  follows `PTO` (ie is 2 for NNLO) so here we should subtract one, once doing the default, no?

_Originally posted by @giacomomagni in https://github.com/NNPDF/eko/pull/299#discussion_r1384385630_

felixhekhorn:
hmm? to me it seems [the LHA benchmarks have passed](https://github.com/NNPDF/eko/actions/runs/6458211610/job/17531433642), which should test the default - and also the N3LO benchmarks seemed alright, which should test the overwriting ... or do you mean we can no longer access the N3LO matching?

niclaurenti:
Hi @giacomomagni and @felixhekhorn , but looking at 

https://github.com/NNPDF/eko/blob/10f0733635c012ce7e1f0fd4cdfc497619292830/src/eko/evolution_operator/operator_matrix_element.py#L66

`matching_order` must be 2 at NNLO:
It should be $delta_{ij}$ at LO, $a_s A^{(1)}$ at NLO and $a_s^2 A^{(2)}$ at NNLO.

giacomomagni:
Indeed, we `matching_oder` should have a shift wrt to `order`.
@felixhekhorn  I think this bug affects only the "new runcards", while in the LH benchmarks run through the old format no ?

felixhekhorn:
Okay, so you are worried about this line specifically:
https://github.com/NNPDF/eko/blob/10f0733635c012ce7e1f0fd4cdfc497619292830/src/eko/runner/parts.py#L93
then maybe I agree with you, there is a bug - do you agree as well @niclaurenti ? When fixing this, we should provide also a unit test

niclaurenti:
Ok @giacomomagni you are right. Sorry for doubting.
Previously I did a test with the "old runcards" (via apfel_bench.py) and it was ok.
Now I did another one with the new ones and in `build_ome` we have `matching_order=(3,0)` at NNLO.

The point in which it must be changed is:

https://github.com/NNPDF/eko/blob/10f0733635c012ce7e1f0fd4cdfc497619292830/src/eko/evolution_operator/operator_matrix_element.py#L167C17-L167C17

giacomomagni:
[...]

PR #323, opened 2023-11-07
address issue #322

giacomomagni:
Address issue #322 .

### DIFF
--- a/src/eko/runner/parts.py
+++ b/src/eko/runner/parts.py
@@ -90,7 +90,7 @@ def evolve_configs(eko: EKO) -> dict:
         n3lo_ad_variation=tcard.n3lo_ad_variation,
         matching_order=tcard.matching_order
         if tcard.matching_order is not None
-        else tcard.order,
+        else (tcard.order[0] - 1, tcard.order[1]),
     )
 
 
--- a/src/eko/runner/parts.py
+++ b/src/eko/runner/parts.py
@@ -88,8 +88,9 @@ def evolve_configs(eko: EKO) -> dict:
         n_integration_cores=ocard.configs.n_integration_cores,
         ModSV=ocard.configs.scvar_method,
         n3lo_ad_variation=tcard.n3lo_ad_variation,
-        matching_order=tcard.matching_order if tcard.matching_order is not None
         # Here order is shifted by one, no QED matching is available so far.
+        matching_order=tcard.matching_order
+        if tcard.matching_order is not None
         else (tcard.order[0] - 1, 0),
     )
 
--- a/src/eko/runner/parts.py
+++ b/src/eko/runner/parts.py
@@ -90,7 +90,7 @@ def evolve_configs(eko: EKO) -> dict:
         n3lo_ad_variation=tcard.n3lo_ad_variation,
         matching_order=tcard.matching_order
         if tcard.matching_order is not None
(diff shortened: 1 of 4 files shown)


giacomomagni:
@felixhekhorn as usual tests of this pieces are quite involved  :( Any idea?

src/eko/runner/parts.py
```diff
@@ -90,7 +90,7 @@ def evolve_configs(eko: EKO) -> dict:
         n3lo_ad_variation=tcard.n3lo_ad_variation,
         matching_order=tcard.matching_order
         if tcard.matching_order is not None
-        else tcard.order,
+        else (tcard.order[0] - 1, tcard.order[1]),
```
felixhekhorn, line 93:
```suggestion
        else (tcard.order[0] - 1, tcard.order[1] - 1),
```
I know QED matchings are beyond our current scope, let's spend 3 characters now if ever we get there ...

giacomomagni: true and also in the legacy...

felixhekhorn:
> @felixhekhorn as usual tests of this pieces are quite involved :( Any idea?

it was not that complicated :upside_down_face: 2cf291d6

src/eko/runner/parts.py
felixhekhorn: in the old style `QED` was a boolean, which translated naturally into the new counting - in the new style we still need to translate - do you agree @niclaurenti ? (actually, are we putting this in the fits for the paper to 2?)

giacomomagni: Okay, then it should be fine.
But on the other hand we should add a minus sign here: 
https://github.com/NNPDF/eko/blob/10f0733635c012ce7e1f0fd4cdfc497619292830/src/eko/io/runcards.py#L216

felixhekhorn: why? again in the old style PTO=0 means LO, no?

giacomomagni: no I meant on the the QED

felixhekhorn: okay, yes - just remember to put also a `min` maybe

niclaurenti: > in the old style `QED` was a boolean, which translated naturally into the new counting - in the new style we still need to translate - do you agree @niclaurenti ? (actually, are we putting this in the fits for the paper to 2?)

In the old style `QED` was the order in $alpha$ of the splitting kernels, so if we had QED corrections to matching conditions it should have a `-1` as for the QCD order (In the fit I'm using `QED=2`).

niclaurenti (approved):

src/eko/runner/parts.py
```diff
@@ -90,7 +90,7 @@ def evolve_configs(eko: EKO) -> dict:
         n3lo_ad_variation=tcard.n3lo_ad_variation,
         matching_order=tcard.matching_order
         if tcard.matching_order is not None
-        else tcard.order,
+        else (tcard.order[0] - 1, tcard.order[1] - 1),
```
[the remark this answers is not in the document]
giacomomagni: Okay I agree, let's set it to 0 everywhere, so it's clear what we are doind.

giacomomagni:
@felixhekhorn if you agree with this solution I'll merge asap.

felixhekhorn (approved):

PR #326, opened 2023-11-21
Enforce default for matching order at post-init time

AleCandido:
Hot fix for matching order setting.

It should be properly tested, but if merging this is urgent we can postpone it.

### DIFF
--- a/src/eko/io/runcards.py
+++ b/src/eko/io/runcards.py
@@ -54,7 +54,7 @@ class TheoryCard(DictLike):
     def __post_init__(self):
         """Enforce defaults."""
         if self.matching_order is None:
-            self.matching_order = self.order
+            self.matching_order = (self.order[0] - 1, 0)
 
 
 @dataclass
--- a/src/eko/io/runcards.py
+++ b/src/eko/io/runcards.py
@@ -46,8 +46,7 @@ class TheoryCard(DictLike):
     xif: float
     """Ratio between factorization scale and process scale."""
     n3lo_ad_variation: N3LOAdVariation
-    """|N3LO| anomalous dimension variation: ``(gg_var, gq_var, qg_var,
-    qq_var)``."""
+    """|N3LO| anomalous dimension variation: ``(gg, gq, qg,qq)``."""
     matching_order: Optional[Order] = None
     """Matching conditions perturbative order tuple, ``(QCD, QED)``."""
 
--- a/src/eko/io/runcards.py
+++ b/src/eko/io/runcards.py
@@ -1,9 +1,7 @@
 """Structures to hold runcards information.
 
-All energy scales in the runcards should be saved linearly, not the squared
-value, for consistency.
-Squares are consistenly taken inside.
-
+All energy scales in the runcards should be saved linearly, not the
+squared value, for consistency. Squares are consistenly taken inside.
 """
 from dataclasses import dataclass
 from math import nan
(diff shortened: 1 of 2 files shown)


src/eko/io/runcards.py
```diff
@@ -48,10 +46,16 @@ class TheoryCard(DictLike):
...
-    """|N3LO| anomalous dimension variation: ``(gg_var, gq_var, qg_var, qq_var)``."""
+    """|N3LO| anomalous dimension variation: ``(gg_var, gq_var, qg_var,
+    qq_var)``."""
     matching_order: Optional[Order] = None
     """Matching conditions perturbative order tuple, ``(QCD, QED)``."""
 
+    def __post_init__(self):
+        """Enforce defaults."""
+        if self.matching_order is None:
+            self.matching_order = self.order
```
AleCandido, line 57: This is the only relevant part. The additional noise has been added by the formatter (I usually try to avoid this, but I authored the PR too quickly...)

AleCandido:
> It should be properly tested, but if merging this is urgent we can postpone it.

I take it back: thanks to @felixhekhorn is already tested. And the behavior was different from the one I imagined.

Actually, this would not work with the polarized settings @RoyStegeman forwarded (the default would not reproduce those settings, but you should keep manually specifying it).
However, before fine-tuning for polarized, I'd like to understand why polarized should be different from unpolarized. @Radonirinaunimi

felixhekhorn:
- I'm slightly worried about unintended side effects (i.e. please wait and check the benchmarks)
- we should drop these lines https://github.com/NNPDF/eko/blob/0c4bf11088354da85090e7990103d3e6478c3cd4/src/eko/runner/parts.py#L93-L94 to avoid double acounting
- it is fine to keep this one https://github.com/NNPDF/eko/blob/0c4bf11088354da85090e7990103d3e6478c3cd4/src/eko/io/runcards.py#L223 since we may have `PTO_matching` (although in real life, I think, this never happens ...)

giacomomagni:
> * we should drop these lines https://github.com/NNPDF/eko/blob/0c4bf11088354da85090e7990103d3e6478c3cd4/src/eko/runner/parts.py#L93-L94
>    to avoid double acounting

I thought the same, but one should be to evolve ekos without that key and the other one for reading ekos without that key no?

felixhekhorn:
> I thought the same, but one should be to evolve ekos without that key and the other one for reading ekos without that key no?

Mmm not really, I think - either case would create a `TheoryCard` object and after calling the constructor you're done

AleCandido:
Indeed, @felixhekhorn is right: they are acting both at the level of the `TheoryCard`, so they are doubled.
The new one is for reading and writing, since acting directly on `TheoryCard`, the old one only for writing, since enforced during computation.

src/eko/io/runcards.py
```diff
@@ -48,10 +46,16 @@ class TheoryCard(DictLike):
     xif: float
     """Ratio between factorization scale and process scale."""
     n3lo_ad_variation: N3LOAdVariation
-    """|N3LO| anomalous dimension variation: ``(gg_var, gq_var, qg_var, qq_var)``."""
+    """|N3LO| anomalous dimension variation: ``(gg_var, gq_var, qg_var,
+    qq_var)``."""
```
felixhekhorn, line 50:
```suggestion
    """|N3LO| anomalous dimension variation: ``(gg, gq, qg,qq)``."""
```

I seem to remember someone (pylint or black?) was complaining if the short description was not in one line (but maybe I misremember ...)

felixhekhorn (approved):

src/eko/io/runcards.py
AleCandido: No one of them, but `pydocstyle`.

This is the formatter that it is instead adding the newline, because the line was too long...

AleCandido:
For me, it was ready. If you and @giacomomagni agree, then I'll merge.

PR #335, opened 2023-12-27
Adding FHMRUVV N3LO splitting functions

giacomomagni:
This PR is imported from: https://github.com/NNPDF/eko_fhmv_private/pull/1

PR #336, opened 2024-01-09
Pass nl as an argument in qed beta functions

niclaurenti:
Implement APFEL running of alpha_em: below $m_{\tau}$, $\alpha$ runs with $n_l=2$. 
Since EKO is never (or at least almost never) used below 1 GeV we don't care about switching off muon and electron running, nor setting $n_f=0$ below 0.5 GeV, as APFEL does.


src/eko/constants.py
```diff
@@ -114,3 +117,22 @@ def charge_combinations(nf):
     vde2m = nd / nf * (eu2 - ed2)
     e2delta = vde2m - vue2m + e2avg
     return e2avg, vue2m, vde2m, e2delta
+
+
+@nb.njit(cache=True)
+def lepton_number(q2):
```
felixhekhorn, line 123: maybe this is better located in `matchings.py` (since this is almost another atlas)

src/eko/constants.py
```diff
@@ -114,3 +117,22 @@ def charge_combinations(nf):
...
+
+    Parameters
+    ----------
+    q : float
+        energy
+
+    Returns
+    -------
+    nl : int
+
```
felixhekhorn, line 136:
```suggestion
    """Compute the number of leptons.

    Note: at the moment we never go below 1 GeV so we don't need muon and electron.

    Parameters
    ----------
    q2 : float
        scale

    Returns
    -------
    int :
       Number of leptons

```

felixhekhorn (commented):
please also adjust the documentation everywhere

src/eko/constants.py
niclaurenti: I put it there since there was already `uplike_flavors`. But if you want I can move it

felixhekhorn: I would still move it - as said it acts as an atlas (I was even briefly suggesting to do a full atlas, but then I thought this is too much); `uplike_flavors` instead is not an atlas and in fact unaware of any scale, but just does the "trivial" `nf // 2` - that is group theory and not physics

src/eko/constants.py
```diff
@@ -84,7 +87,7 @@ def uplike_flavors(nf):
     nu : int
 
     """
-    if nf not in range(2, 6 + 1):
+    if nf > 6:
```
giacomomagni, line 90: why do you need to change here?

src/eko/couplings.py
```diff
@@ -737,7 +738,20 @@ def a(
...
-                new_a = self.compute(final_a, seg.nf, seg.origin, seg.target)
+                nli = constants.lepton_number(seg.origin)
+                nlf = constants.lepton_number(seg.target)
+                if nli != nlf:
+                    # it means that MTAU is between origin and target:
+                    # first we evolve from origin to MTAU with nli leptons
+                    a_tmp = self.compute(
+                        final_a, seg.nf, nli, seg.origin, constants.MTAU**2
+                    )
+                    # then from MTAU to target with nlf leptons
```
giacomomagni, line 749: Just for curiosity: At NLO there are no matching conditions to add here ?

benchmarks/eko/benchmark_strong_coupling.py
```diff
@@ -322,7 +322,7 @@ def benchmark_APFEL_vfns(self):
                 # print(apfel_vals_cur)
                 np.testing.assert_allclose(apfel_vals, np.array(apfel_vals_cur))
             # check myself to APFEL
-            np.testing.assert_allclose(apfel_vals, np.array(my_vals))
+            np.testing.assert_allclose(apfel_vals, np.array(my_vals), rtol=3e-3)
```
giacomomagni, line 325: why this is changed?

src/eko/constants.py
niclaurenti: because it is used in validphys in this line https://github.com/NNPDF/nnpdf/blob/970cd427415fd6139991c2da1cb5b5ffecfa8df8/validphys2/src/validphys/photon/compute.py#L413 , in which now nf can be 0

giacomomagni: I see okau

benchmarks/eko/benchmark_strong_coupling.py
niclaurenti: fixed it with https://github.com/NNPDF/eko/blob/1999d92e71a8ba687f1f611af237fcd88f96e5b4/src/eko/couplings.py#L749

src/eko/matchings.py
```diff
@@ -209,3 +211,23 @@ def is_downward_path(path: Path) -> bool:
 def flavor_shift(is_downward: bool) -> int:
     """Determine the shift to number of light flavors."""
     return 4 if is_downward else 3
+
+
+@nb.njit(cache=True)
+def lepton_number(q2):
+    """Compute the number of leptons.
+
+    Note: at the moment we never go below 1 GeV so we don't need muon and electron.
```
giacomomagni, line 220:
```suggestion
    Note: muons and electrons are always massless as for up, down and strange.
```

src/eko/couplings.py
niclaurenti: that's a good question...

felixhekhorn: I guess (!) the whole derivation goes through for any theory of a scalar boson with fermions (with the appropriate changes on couplings etc.) ... however, quoting SF from a mail on Monday "This said, it seems to me that using a VFN for leptons corresponds to using a scheme that differs from MSbar in the QED sector, which is not a very good idea!"

src/eko/beta.py
```diff
@@ -32,7 +32,7 @@ def beta_qcd_as2(nf):
 
 
 @nb.njit(cache=True)
-def beta_qed_aem2(nf):
+def beta_qed_aem2(nf, nl):
```
felixhekhorn, line 35: some more docs copying please :upside_down_face:

tests/eko/test_beta.py
felixhekhorn, line 1: please add some more unit tests here (and maybe even higher? I mean on $a(Q^2)$-level?)

niclaurenti:
@felixhekhorn @giacomomagni regarding the conversation about the VFNS in the lepton number and the matching:
the reason I am doing this PR is that this week I talked with SF and turned out that the photon I obtained was different from the one in NNPDF3.1QED since the $\alpha_{qed}$ running was different.
He wanted this running to be as close as possible to the one coming from the package alphaQED (http://www-com.physik.hu-berlin.de/~fjeger/software.html), that we should take as God-given truth.
Implementing the $\alpha_{qed}$ running as it is done in APFEL seems to mimic this running very well: this was probably the reason why they switched off $n_f$ at a certain point (moreover $\alpha_{qed}$ must exhibit a plateau for very small $Q$ and the only way to do it is to send the beta function to zero, i.e. $n_f=n_l=0$).
This is why I implemented this running first of all in the photon module in validphys and then here in EKO (in order to be consistent). I attach the plots to show what I'm saying: in the first one there is the wrong running, while in the second the running has been fixed
[alphaQED23_vs_pineline4.pdf](https://github.com/NNPDF/eko/files/13893936/alphaQED23_vs_pineline4.pdf)
[alphaQED23_vs_pineline_mq05.pdf](https://github.com/NNPDF/eko/files/13893938/alphaQED23_vs_pineline_mq05.pdf)

In the end a test in which $m_q$ (the scale at which we pass to $n_f=0$) is varied
[alphaQED23_vs_pineline_mq1.pdf](https://github.com/NNPDF/eko/files/13893942/alphaQED23_vs_pineline_mq1.pdf)

src/eko/evolution_operator/__init__.py
```diff
@@ -704,40 +704,20 @@ def compute_aem_list(self):
...
             for step, mu2_h in enumerate(mu2_steps[1:]):
                 mu2_half = (mu2_h + mu2_l) / 2.0
-                a_s, aem = couplings.compute(
-                    a_ref=a_start, nf=self.nf, scale_from=mu2_start, scale_to=mu2_half
-                )
+                a_s, aem = couplings.a(scale_to=mu2_half, nf_to=self.nf)
                 a_half[step] = [a_s, aem]
                 mu2_l = mu2_h
+        print(as_list.tolist())
+        print(a_half.tolist())
```
felixhekhorn, line 720:
```suggestion
```
do you want to have this PR in draft mode? :innocent:

tests/eko/test_beta.py
felixhekhorn: `alpha23` can be a good candidate for a benchmark as it seems we want to replicate that one

src/eko/evolution_operator/__init__.py
niclaurenti: It is ready. I'm doing benchmarks with master but I missed that print :)

niclaurenti:
I have performed some benchmarks and the agreement with APFEL is the same of the master branch (as it should be since it is a tiny effect)

tests/eko/test_beta.py
felixhekhorn: I like the new benchmark :+1: , but please add still some unit tests in `test_beta.py`

niclaurenti: Done!

felixhekhorn: you can also add another good physics case: $\beta^{(2n_l)} = 2\beta^{(n_l)}\, \forall n_l$

niclaurenti: Both done!

benchmarks/eko/benchmark_alphaem.py
```diff
@@ -0,0 +1,144 @@
+"""This module benchmarks alpha_em against alphaQED23."""
```
felixhekhorn, line 1:
```suggestion
"""This module benchmarks alpha_em against alphaQED23.

alphaQED23 can be obtained from http://www-com.physik.hu-berlin.de/~fjeger/software.html .
"""
```

felixhekhorn (approved):

felixhekhorn:
This will be in conflict with #335 as both touch QED  kernels (should be simple to solve - however, but be aware)

niclaurenti:
> This will be in conflict with #335 as both touch QED kernels (should be simple to solve - however, but be aware)

There was no conflict

felixhekhorn:
> There was no conflict

even better (because you are touching `sv.exponentiated.gamma_variation_qed` and #335 `ad_us.gamma_singlet_qed`)

felixhekhorn:
let's wait for https://github.com/NNPDF/eko/actions/runs/7501124078 and then merge

Issue #263, opened 2023-05-09
pole structure NNLO time-like ad


felixhekhorn:
@t7phy I know you copied the time-like anomalous dimension from A. Vogt - but I wonder how useful is this condition:
https://github.com/NNPDF/eko/blob/fb46d8b6a7588465de581ad7e31270f36696a7ee/src/ekore/anomalous_dimensions/unpolarized/time_like/as3.py#L268
are we ever hitting that case? i.e. can we remove this branching?

Moreover, the function also contains explicit $1/N$ (which already makes $N=0$ unstable) and even $1/(N-1)$ ... speaking of, can you remind me about the pole structure of the time-like anomalous dimensions? because in the space-like case `nsv` does neither have a pole at $N=1$ nor at $N=0$ (because number conservation is defined there)

cc @enocera

t7phy:
Hi @felixhekhorn 

For the 1st question, the answer is that we do not really know whether we will need that condition or not, that is why it's preserved as was given by Vogt.

For the second question, it's my understanding that Vogt's functions use simple harmonic sums like s1, s2, s3 but the other harmonic sums are made with the combination of the required terms. It is in these combinations where the terms like 1/N or 1/N-1 are used.

felixhekhorn:
Ciao @t7phy 

> For the 1st question, the answer is that we do not really know whether we will need that condition or not, that is why it's preserved as was given by Vogt.

well, the question whether you need this or not is directly related to the pole structure; to be specific: if there is already a pole at $N=0$ there is no need to regularize that point ...

> For the second question, it's my understanding that Vogt's functions use simple harmonic sums like s1, s2, s3 but the other harmonic sums are made with the combination of the required terms. It is in these combinations where the terms like 1/N or 1/N-1 are used.

of course they are made by those terms, because they are the natural building blocks. The exact NNLO expressions might be more complicated, but the Vogt strategy is to approximate them with s-es and polynomials - however, he always preserves the *physics features*, i.e. the large-N behaviour and the pole structure of the leading pole. Please check the relevant papers for the pole structure - they should be discussed there, because, as said, they correspond to actual physics. And then double check the pole structure on the actual expression e.g. with Mathematica.

giacomomagni:
Just for my curiosity, do you know why is this term replaced by `zeta2` in the small N limit? 

```
NI = 1/n;
NM = n - 1;
NMI = 1/NM;
S1M = S[1,n] - NI;
B1M = -S1M * NMI;
[...]

PR #344, opened 2024-02-07
fixing the condition from #263

t7phy:
It appears I had made a mistake when reimplementing this code from https://github.com/vbertone/MELA/blob/master/src/evolution/andim_nnlo_tl.f#L117-118. It is now fixed. 

The ref in Vogt's code has this eqn which matches our definition of Mellin transform: 
image

so with this, I think everything is consistent.

### DIFF
--- a/src/ekore/anomalous_dimensions/unpolarized/time_like/as3.py
+++ b/src/ekore/anomalous_dimensions/unpolarized/time_like/as3.py
@@ -265,7 +265,7 @@ def gamma_nsv(N, nf, cache):
     S11 = S1 + N1I
     S12 = S11 + N2I
     B1 = -S1 * NI
-    if abs(N.imag) < 0.00001 and abs(N.real) < 0.00001:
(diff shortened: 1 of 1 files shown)


giacomomagni:
Okay, good that you sorted out. 

For my understanding, since you only shifted the condition, but not N in the whole function, 
does it mean that the Mela implementation is bugged ?

t7phy:
> Okay, good that you sorted out.
> 
> For my understanding, since you only shifted the condition, but not N in the whole function, does it mean that the Mela implementation is bugged ?

Actually, the change made now corresponds with MELA, i.e. I had made a mistake earlier when copying the condition from MELA, so MELA implementation should be fine.

giacomomagni:
Ah ok, got it, there was only a typo. :)

t7phy:
> Ah ok, got it, there was only a typo. :)

Indeed :)

giacomomagni (approved):

Issue #201, opened 2023-01-20
Test scale variations expanded against exponentiated


andreab1997:
At the moment the test `scheme_diff` in `[/scale_variations/test_expanded.py](https://github.com/NNPDF/eko/pull/191/files/591dfbb78381a5aa043b7103c8605767bc074a2d#diff-69e13373b89b52fc6ae55e1044c7029153d1724257ae27f74352b9cb8311ebef) is not really checking the difference between scheme A and B (i.e exponentiated vs expanded). However we know how to do the proper test because we have all the terms in a mathematica notebbok in pineko ([here](https://github.com/NNPDF/pineko/pull/68)). While the actual expressions for scale variations have been checked against this notebook, the test has not been updated consistently yet (for lack of time). We should
do the proper thing as soon as possible. @AleCandido @felixhekhorn @giacomomagni

felixhekhorn:
Note that in #222 the SV behaviour was re-derived and changed (and, e.g., in effect for `exponentiated` $\xi_F$ flipped on it's head)

PR #346, opened 2024-02-13
Test scale variation schemes difference. 

giacomomagni:
This PR aims to close issue #201 and provide a reasonable test for the difference between the 
scale variations kernels, at least for FFNS. 

A further test at PDF level would be possible, but even more complicated.


felixhekhorn:
- I reshuffled the test a bit: lowering L and dropping thus the 3. order
- now this is a hard math statement: the difference is proportional $a_0 * L$, which we see is not well behaved
- in fact, even for $Q_1^2=99^2$ (and xif2 = 0.9) I can not drop the second order

felixhekhorn:
I added another stupid test, that checks `expanded` is linear in $L$

giacomomagni:
I extended the test to NLO. 
Given that numbers are so close to `1`, not sure we are actually testing anything but still...
I think we can merge @felixhekhorn

felixhekhorn (approved):

PR #349, opened 2024-02-27
Allow py 3.12

### LINKED ISSUE #348: Release for py 3.12
Numba 0.59 supports python 3.12, would there be anything left in eko that would not work with it?

felixhekhorn:
Closes #348

cc @scarlehoff


felixhekhorn:
waiting for https://github.com/N3PDF/banana/pull/61

felixhekhorn:
@alecandido if you have time, take a look - else I use you as trigger for LHA :upside_down_face:

alecandido (approved):

felixhekhorn:
For future reference: we needed to port the changes from https://github.com/N3PDF/banana/pull/61 to here - which are:
- drop support for python 3.8
- use pandas >2
- use pylint >3.1

in addition, I set numba > 0.59 (which is out since 4 weeks) as I failed [here](https://github.com/NNPDF/eko/actions/runs/8078156542/job/22069866292) and a simple update seemed to solve the issue ...

alecandido:
This is reasonable, because `ekomark` not only depends on `banana`, but uses those tools directly (to play with `banana` objects of course, but the dependency is explicitly imported).

So, it should be sensible, since with `ekomark` you really have `banana` around. We should just stop the `ekomark` propagation downstream.

PR #355, opened 2024-03-11
Add a nicer parametrization of aggQ3

giacomomagni:
in this PR I'd like to make the $a_{gg,Q}^{(3)}$ parametrization to have a easier and compact form in `x-space`, 
such that can be easily digest alo in that case. 

As visible in the two plots below, the new parametrization is equivalent to the previous one.
 
Relative difference in linear scale:
[aggQ3_rel_diff.pdf](https://github.com/NNPDF/eko/files/14555988/aggQ3_rel_diff.pdf)

Comparison in log scale:
[aggQ3_diff.pdf](https://github.com/NNPDF/eko/files/14555989/aggQ3_diff.pdf)

This parametrization can be shared easily with the MSHT people.

### DIFF
--- a/src/ekore/operator_matrix_elements/unpolarized/space_like/as3/agg.py
+++ b/src/ekore/operator_matrix_elements/unpolarized/space_like/as3/agg.py
@@ -5,15 +5,33 @@ import numba as nb
 import numpy as np
 
 from .....harmonics import cache as c
+from .....harmonics.log_functions import (
+    lm11,
+    lm11m1,
+    lm11m2,
+    lm12,
+    lm12m1,
+    lm12m2,
+    lm13,
+    lm13m1,
+    lm13m2,
+    lm14m1,
+    lm14m2,
+)
 
 
 @nb.njit(cache=True)
 def a_gg3(n, cache, nf):
-    r"""Compute the approximate part of :math:`a_{gg}^{(3)}(N)`.
+    r"""Compute :math:`a_{gg}^{(3)}(N)`.
 
-    This is the part of :math:`A_{gg}^{S,(3)}(N)` proportional to :math:`\mathcal{O}(\epsilon^0)`,
-    the expression is presented in  :cite:`Ablinger:2022wbb`.
-    It contains binomial factors which are approximated.
+    The expression is presented in  :cite:`Ablinger:2022wbb`.
+
+    The :math:`n_f^0` piece is parametrized from:
+    the known small-x (eq. 4.10) and large-x (eq. 4.11)
+    limits, the expansion of the local and singular parts
+    in eq. 4.6, 4.7 and the first 15 Mellin moments up to N=30.
+    The analytical expression contains binomial factors
+    which are not handy to use.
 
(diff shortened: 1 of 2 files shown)


felixhekhorn:
Mmm ...
- isn't the also the old one simple to translate? it just contains $S_k$ s (which are straight to translate) and $S_1/n$, which maybe we also know (I would need to think a moment, though)
- the two don't seem to have the same limits: i.e.
  - the old is $S_1^4$ but the new $S_1^1$ for $N\to \infty$
  - the old and new poles at $N=1$ differ by ~10%, is that ok?

giacomomagni:
> * isn't the also the old one simple to translate? it just contains Sk s (which are straight to translate) and S1/n, which maybe we also know (I would need to think a moment, though)

Sure `S[k,n]` translates in to `PolyLog[k, x]`. 
This paramterization, avoids even that problem, as only Logs appear now.

The main point is that new one reproduces exactly the limits given in the paper, 
thus is more consistent with the procedure we do for `A_Qg`,
while for instance before the $\Delta(1-x)$ was different.

> * the two don't seem to have the same limits: i.e.
>   
>   * the old is S14 but the new S11 for N→∞

The new one is more respectful of the `x \to 1` limit provided in the paper, 
so if anything it's even better.

>   * the old and new poles at N=1 differ by ~10%, is that ok?

The LL term should be the same, there is a difference in the NLL ($\propto \frac{1}{x}$), 
as visible in the plot.

giacomomagni:
So regarding the small-x, the difference comes form:

* Before I reconstructed the pole at $N=1$ and check that was in reasonable agreement with the limit of the N-space result:

```
643.592/(-1. + n)^2 - 4276.31/(-1. + n)
```

* Now I used the ancillary file containing the $x \to 1 $ limit provided in the paper, which gives:

```
701.199/(-1 + n)^2 - 4954.44/(-1 + n) 
```

The funny part is that if I code [eq 4.10](https://arxiv.org/pdf/2211.05462.pdf) (at least for the term $\propto 1/x \ln(x) $), now I get: 

```
-643.588
```
which matches with old implementation. So I guess one of the provided file, is not consistent...

felixhekhorn:
> The funny part is that if I code [eq 4.10](https://arxiv.org/pdf/2211.05462.pdf) (at least for the term

that one I could still type :see_no_evil: - and no I don't can confirm your disagreement, but I find agreement with the ancillary file:

```python
from scipy.special import zeta
ca = 3
tf = .5
cf=4./3.
z2 = zeta(2)
z3 = zeta(3)

ll = ca*tf*(ca*(-11488/81+224*z2/27+256*z3/3) + cf*(-15040/243-1408*z2/27-1088*z3/9))
print(ll)
```
yields `-701.1986854426287`

felixhekhorn (approved):
I conclude: for both large and small x we're closer to the truth. Then I think this is fine.

giacomomagni:
Okay, I'd wait for MHST feedbacks, if any, and then eventually merge.

PR #362, opened 2024-04-23
Fix large-x wiggles 

giacomomagni:
Here we change the `r` parameter regulating the Talboth path, to make the large-x integration smoother. 
@scarlehoff

### DIFF
--- a/src/eko/mellin.py
+++ b/src/eko/mellin.py
@@ -223,7 +223,7 @@ class Path:
     def __init__(self, t, logx, axis_offset):
         self.t = t
         # TODO: shall we use: 0.4 * 16.0 / ( - logx) ?
-        self.r = 0.4 * 16.0 / (1.0 - logx)
+        self.r = 0.4 * 16.0 / (0.1 - logx)
         if axis_offset:
             self.o = 1.0
         else:
--- a/src/eko/mellin.py
+++ b/src/eko/mellin.py
@@ -222,7 +222,10 @@ class Path:
 
     def __init__(self, t, logx, axis_offset):
         self.t = t
-        # TODO: shall we use: 0.4 * 16.0 / ( - logx) ?
+        # The prescription suggested by :cite:`Abate` for r is 0.4 * M / ( - logx)
+        # with M the number of accurate digits; Maria Ubiali suggested in her thesis M = 16.
+        # However, this seems to yield unstable results for the OME in the large x region
+        # so we add an additional regularization, which makes the path less "edgy" there
         self.r = 0.4 * 16.0 / (0.1 - logx)
         if axis_offset:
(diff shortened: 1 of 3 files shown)


src/eko/mellin.py
```diff
@@ -223,7 +223,7 @@ class Path:
     def __init__(self, t, logx, axis_offset):
         self.t = t
         # TODO: shall we use: 0.4 * 16.0 / ( - logx) ?
```
giacomomagni, line 225:
```suggestion
        #original prescription is with  0.4 * 16.0 / ( - logx) ?
```

src/ekobox/evol_pdf.py
```diff
@@ -56,7 +57,7 @@ def evolve_pdfs(
     else:
         if store_path is None:
             raise ValueError("'store_path' required if 'path' is not provided.")
-        eko.solve(theory_card, operators_card, path=store_path)
+        managed.solve(theory_card, operators_card, path=store_path)
```
alecandido, line 60: Actually, this is a separate feature from the one described by the PR.

Make sure to advertise it properly.

(Also because the `managed` runner was not intended to be always used, but just for quick calculations -> if it becomes the default, this information should be propagated)

src/eko/mellin.py
felixhekhorn: take a look to my suggestion

src/ekobox/evol_pdf.py
felixhekhorn: @alecandido is still right :upside_down_face: you need to document this

felixhekhorn: actually, let's do the proper thing and just make two PRs

felixhekhorn: #363

src/eko/mellin.py
giacomomagni: excellent! Thanks

felixhekhorn:
Just as a curious side note: this has almost no impact on our LHA benchmarks (which ran on the first commit) - meaning the last point $x=0.9$ still has the worst relative accuracy of up to a few percent (of course the absolute error is still just fine as before)

scarlehoff:
I think this should be merged (and a version tagged if possible so that the MC PDFs are released using a tagged version of the code).

niclaurenti (approved):

niclaurenti:
@giacomomagni @felixhekhorn can we merge it?

giacomomagni:
Green light from my side!

felixhekhorn:
I think if this work in an actual fit, this is an actual fix

Please double check the changes since the last release before a new tag (I think we should still be compatible, but better double check)

PR #369, opened 2024-05-13
Rustify ekore part 2: ad.u.s.{as2,as3}

felixhekhorn:
Actually, we should push for the continued rustification before @giacomomagni receives his halo and someone else has to translate all those endless N3LO expression. I will not attempt the full rustification here but still proceed in parts (i.e. this is part 2 (after #189 ) out of X): currently my only goal here is to be ready for QCD N3LO ad. To be specific:

ekore/anomalous_dimensions/unpolarized/space_like/
- [x] as2.py
- [x] as3.py


giacomomagni:
Thanks @felixhekhorn 
I'm not sure if I'll be able to contribute to this endeavour at some point

felixhekhorn:
> for the time being we go for qcd only

indeed - let's proceed in parts and first we do QCD only

One simple comment up front: please run pre-commit

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as2.rs
```diff
@@ -0,0 +1,236 @@
...
+    let gamma_qq = gamma_nsp(c, nf) + gamma_ps(c, nf);
+    [
+        [gamma_qq, gamma_qg(c, nf)],
+        [gamma_gq(c, nf), gamma_gg(c, nf)],
+    ]
+}
+
+pub fn test_const() -> f64 {
+    CA
+}
```
felixhekhorn, line 116:
```suggestion

```

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as2.rs
```diff
@@ -0,0 +1,236 @@
...
+    use crate::{anomalous_dimensions::unpolarized::spacelike::as2::*, harmonics::cache::Cache};
+    use float_cmp::assert_approx_eq;
+    use num::complex::Complex;
+    use std::f64::consts::PI;
+    use num::traits::Pow;
+
+    const NF: u8 = 5;
+
+    #[test]
+    fn gamma_1() {
```
felixhekhorn, line 130: I would split the tests apart:
- physical constraints
- N=2
- N=3
- N=4

felixhekhorn (commented):
- I think the `NC=4` case is artificial and we simply allow for this hypothetical feature any longer
- of course we eventually have to hook the function in on the higher level

felixhekhorn:
> mm ok. I guess by `rustify` you mean running the script `rustify.sh`?

exactly - this will hook the Rust backend into the main eko package + remember to actually compile via `poe compile`; this way we can run the usual eko workflows (with the limited available orders of course) e.g. to check against the LHA benchmark number which we will have to reproduce ...

tgiani:
ok I see. Other question: isn t `pytest` installed automatically in the poetry venv when I run `poetry install`?
If I do `poetry install`, rustify, `poe compile` and then `poe lha -m "ffns and nlo and not sv"` it doesn t find `pytest`

felixhekhorn:
> ok I see. Other question: isn t `pytest` installed automatically in the poetry venv when I run `poetry install`? If I do `poetry install`, rustify, `poe compile` and then `poe lha -m "ffns and nlo and not sv"` it doesn t find `pytest`

no, since it is a developer dependency - you can do `poetry install --with test` (Poetry has this [dependency groups](https://python-poetry.org/docs/managing-dependencies/#dependency-groups))

felixhekhorn:
> now somehow `cargo fmt` is broken for me and I don't know why since it was working fine the whole afternoon, so I might have broken pre-commit myself, but I will try to fix this tomorrow

okay, so it turns out that it was just a veeery long run for `cargo fmt`, but eventually it does terminate - as a fix I inserted a bunch of [`#[rustfmt::skip]`](https://github.com/rust-lang/rustfmt#tips) - not all of them are actually required, but mostly the ones for the long expressions. However, then for consistency I put them all. In the end the expressions are just math and there is no big insight in having them formatted (although they are easier to check this way). What do you think @tgiani? should we keep the markers?

tgiani:
yea I think so else every time we commit we have to wait half an hour

tgiani:
I m trying to run the benchmark `poe lha -m "ffns and nlo and not sv"` to investigate the problem you mentioned yesterday but that also has been run for several minutes now. Do you know how long it is supposed to take?

felixhekhorn:
> I m trying to run the benchmark `poe lha -m "ffns and nlo and not sv"` to investigate the problem you mentioned yesterday but that also has been run for several minutes now. Do you know how long it is supposed to take?

Ehm, yes this is a feature :upside_down_face: (one that we are trying to partly tackle here) for the first run eko will take a while, because it needs to compile all the numba stuff, which is a lot. Normally it can take up to 15 min if the OME are involved, but actually here it should be quicker, since we're running FFNS and you `poe compile` already did part of the work. You can check if numba is the problem by setting [NUMBA_DEBUG](https://numba.readthedocs.io/en/stable/reference/envvars.html#envvar-NUMBA_DEBUG) or you can set[ EKO_LOG_STDOUT](https://github.com/NNPDF/eko/blob/2348778fa7a9688a0a482f7ed0e3baefbd816722/benchmarks/lha_paper_bench.py#L145) to check what EKO is doing.

felixhekhorn:
> ok after 30 min I ve stopped it and I m trying again setting the two env variables you suggested. Is the singlet sector alright from the benchmark? If so the problem should be something in `gamma_nsm` no?

Mmm not sure ... in the singlet sector we match my 0.4% and I'm not sure if that is good enough (LO we match by 0.02%) and we should check with master. I also see that T8 and T15 are kind of okay, but T3 is not - and all should be NS+ ... Of course T3 is small in absolute value, so maybe there is some precision loss somewhere?

tgiani:
mm i don t know... In the expressions I m doing the following replacements:
- I turn all integers except those in exponents into floats (else algebra with complex is not defined)
- np.power(N,integer)->N.powu(integer) 
- np.pi**2 -> PI.pow(2)
- NF -> (NF as f64)

tgiani:
anyway I m still stuck on this `poe lha -m "ffns and nlo and not sv"`. Since about 1hr 30 min the last log I get from numba is 
```
================================================================================
-------------------------OPTIMIZED DUMP cb_quad_ker_qcd-------------------------
; ModuleID = 'cb_quad_ker_qcd'
source_filename = "<string>"
target datalayout = "e-m:o-i64:64-i128:128-n32:64-S128"
target triple = "arm64-apple-darwin23.4.0"

================================================================================
================================================================================
----------------------------ASSEMBLY cb_quad_ker_qcd----------------------------
        .section        __TEXT,__text,regular,pure_instructions
        .build_version macos, 14, 0
.subsections_via_symbols

================================================================================
```
not sure what the problem here

tgiani:
@felixhekhorn ah ok good. Yes also after your last commit I still have the same problem on mac

felixhekhorn:
Continuing from here we can have several parallel parts:
1. ome.u.s.as1 and ome.u.s.as2
1. recover SV
1. ad.p.s.*
1. recover solution methods
1. ad.u.s.as4 and ad.u.s.as4.fhmruvv
1. ad.u.s.\*aem*
1. ad.u.t.*

The order of the list is a slight indication of priority. Not every item needs a separate PR although that might help in the review. In case you are not tired yet @tgiani pick your favourite!

tgiani:
sure, I guess I ll start from 1. What s 2?

PR #368, opened 2024-05-13
Fix NNLO polarized

felixhekhorn:
thanks @igborsa!

## anomalous dimensions

```
While comparing with the results from EKO and APFEL, I noticed that there were some differences between the EKO and polPEGASUS for the distributions related to the singlet combination (for the valence distributions, on the other hand, there was complete agreement). I traced that back to a couple of bugs in the "\Delta P_{gq}^{(2)}" and "\Delta P_{PS}^{(2)}" implementations in EKO. Specifically:

*) In lines 180-182, the harmonic sums S_2, S_3 and S_4 are being called using the value of S_1.
*) There is a similar bug in lines 263 and 264.

I have checked that correcting those calls leads to similar results as those of polPEGASUS (with relative differences that are typically between 10^{-6}-10^{-5} for most of the range in x). 
```

these are obvious typos and fixed in 5538c2e2

## matching conditions

```
We have now also implemented the matching coefficients from 2211.15337. I turned those factors off to produce the tables in the repo, but, in case you want, I can easily upload similar results including the OMEs. Just in case, I compared numerically the values of the OMEs against those in EKO. For A_{Hq}^{PS,(2)}", "A_{gq,H}^{S,(2)}", "A_{gg,H}^{S,(2)}" and "A^{S,(2)}(N)" I obtained the exact same results. However, there was a difference for "A_{Hg}^{S,(2)}". I think the difference is originating in the following lines:

*) Line 176: The factor should be a 6.
*) Line 191: The correct term should be (2 + 3 * n + 3 * n**2).
*) Line 210: The correct power of the factor (1+n) should be 0.
*) Lines 242-252: It seems that the Mellin transform of [Li_2(x)/(1+x)] is being evaluated in N, instead of N+1. 
```
@giacomomagni can you please have a look? I remember you fixed a (significant) number of typos in the original

PR #373, opened 2024-05-16
Add ome.ps.as2 benchmark vs. Pegasus

felixhekhorn:
@igborsa is it okay if I copy your implementation of the polarized NNLO OME here?

As will be visible in the CI the benchmark fails in the last line - commenting that line out, we pass

The error will be
```
E           AssertionError: 
E           Not equal to tolerance rtol=1e-07, atol=0
E           hg,N=2
E           Mismatched elements: 1 / 1 (100%)
E           Max absolute difference: 0.2584379
E           Max relative difference: 0.02809631
E            x: array(-9.456725+0.j)
E            y: array(-9.198287)
```

### DIFF
--- a/benchmarks/ekore/benchmark_pegasus_ad_us_as2.py
+++ b/benchmarks/ekore/benchmark_pegasus_ad_us_as2.py
@@ -1,4 +1,7 @@
-"""Benchmark the NLO anomalous dimensions against PEGASUS"""
+"""Benchmark the unpolarized NLO anomalous dimensions against PEGASUS.
+
+Recall that we obtained our representation not from PEGASUS, but derived it
+ourselves (see comment there)."""
 
 import numpy as np
 import pytest
@@ -21,7 +24,6 @@ def check_gamma_1_pegasus(N, NF):
     ZETA2 = zeta2
     ZETA3 = zeta3
 
-    # N = np.random.rand(1) + np.random.rand(1) * 1j
     S1 = h.S1(N)
     S2 = h.S2(N)
 
--- a/benchmarks/ekore/benchmark_pegasus_ad_us_as2.py
+++ b/benchmarks/ekore/benchmark_pegasus_ad_us_as2.py
@@ -9,39 +9,6 @@ from eko.constants import CA, CF, TR, zeta2, zeta3
 
 
 @pytest.mark.isolated
-def benchmark_melling_g3_pegasus():
-    for N in [1, 2, 3, 4, 1 + 1j, 1 - 1j, 2 + 1j, 3 + 1j]:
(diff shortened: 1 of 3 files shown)


igborsa:
@felixhekhorn Yes, no problem at all.

felixhekhorn:
@igborsa as the CI shows with the last commit we are now passing the benchmark - so I think you should do the same on your side.

Just to rephrase what @giacomomagni [already said](https://github.com/NNPDF/eko/pull/368#issuecomment-2111236336) on #368 : we are using two different representation of the same object and [your fix](https://github.com/NNPDF/benchmark_pol_timelike_evol/commit/2a37eea7b45dd5c20efc53e9766a64011f94a7a7) for `ACG3` is not needed. You are using the lhs of Eq. 131 of 2211.15337, specifically Eq. 113 _as written_, which requires all objects, i.e. $g_3$ and $\beta$, to be evaluated at $N+1$ (this is what my fix is doing here and the bug you introduced changed) - we are using the rhs of Eq. 131 (and do _not_ use Eq. 113 as written), but call the linear independent $S_{-2,1}$ eventually at $N$.

In practice you just need to revert that line.

igborsa:
Hi @felixhekhorn. 

Actually, the fix is needed. The routine ACG3(N) returns the Mellin transform already evaluated in N+1, so using ACG3(N1) would return the result evaluated in N+2 (that's what confused me when I listed the potential bugs). The correct "translation" to the eko routines is indeed

ACG3(N) = h.g_functions.mellin_g3(N1, S11)

and that's why your last modification worked.

felixhekhorn:
> Hi @felixhekhorn.
> 
> Actually, the fix is needed. The routine ACG3(N) returns the Mellin transform already evaluated in N+1, so using ACG3(N1) would return the result evaluated in N+2 (that's what confused me when I listed the potential bugs). The correct "translation" to the eko routines is indeed
> 
> ACG3(N) = h.g_functions.mellin_g3(N1, S11)
> 
> and that's why your last modification worked.

I see ... but this means the remaining difference we have between us must come from somewhere else then $A_{hg},A_{hq},A_{gg},A_{gq}$, which would have been the prime candidates ...

igborsa:
It seems that (at least part of) the difference we have is actually coming from the term \beta''(N+1) in A_Hg.
I was using the implementation from the code ANCONT (BET2) that apparently is not precise enough. 

Replacing BET2(N+1) -> (DPSI((N1+1)/2.d0,2) - DPSI(N1/2.d0,2))/8.d0,

where  DPSI(N,2) corresponds to the second derivative of the Digamma, seems to improve the agreement between eco and pol-pegasus

Issue #389, opened 2024-07-15
Low Q2 EXPANDED alpha_s is broken.


giacomomagni:
I seems that alphas evolution with EXPANDED  solution is broken at very low Q2 from NLO and higher.

```python
import eko    
from ekobox.cards import  example
from eko.io.types import EvolutionMethod

tcard = example.theory()
tcard.order = (2,0)
for evmod in [
    EvolutionMethod.ITERATE_EXACT,
    EvolutionMethod.TRUNCATED,
]:
    print("Evolution method:",  evmod.value)
    print("Evolution order:",  str(tcard.order))
    evmod = eko.couplings.couplings_mod_ev(evmod)
    sc = eko.couplings.Couplings(
            tcard.couplings,
            tcard.order,
            evmod,
            masses=[(x.value) ** 2 for x in tcard.heavy.masses],
            hqm_scheme=tcard.heavy.masses_scheme,
            thresholds_ratios=[1,1,1],
        )

    for q2 in [0.03, 0.04]:
        print("q2: ", q2, "as: ",sc.a_s(q2))
```
Returns:

```
Evolution method: iterate-exact
Evolution order: (2, 0)
q2:  0.03 as: 386447.3190965184
q2:  0.04 as: 386447.3190965184
Evolution method: truncated
Evolution order: (2, 0)
q2:  0.03 as: nan
q2:  0.04 as: 23.626254462242528
```

Although this is might sound irrelevant, because Q2 is really low, if by chance an FKTable has such value you get all the values to be `nan`. 

We should at least return an error if this happens.

giacomomagni:
The point which break is exactly here:

https://github.com/NNPDF/eko/blob/38cc28c5c6c646118b674a51023b7a816d1d080e/src/eko/couplings.py#L105

for too low values of Q2 this become negative. 
So the solution is to return and error and crash, otherwise we might fill ekos and fktables with nans.

felixhekhorn:
Mmm ... not sure ... surely I would not classify this as bug: you give non-sense, you expect non-sense, you get non-sense.

> for too low values of Q2 this become negative.

This precisely defines $\Lambda_{QCD}$.

I'd say this is a feature and I'm tempted to close this as "will not be fixed".

Also from what I learned from @alecandido I think this is a pythonic attitude: a library accepts almost anything as input (and hopes it casts itself to the representation it needs) and just uses it as. It is the callers responsibility and not the libraries to do the sanity checks.

alecandido:
[...]

PR #390, opened 2024-07-15
Limit Q2 range for expanded as evolution.

giacomomagni:
address issue #389

### DIFF
--- a/src/eko/couplings.py
+++ b/src/eko/couplings.py
@@ -58,6 +58,14 @@ def couplings_mod_ev(mod_ev: EvolutionMethod) -> CouplingEvolutionMethod:
 
 
 @nb.njit(cache=True)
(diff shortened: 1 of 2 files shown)


felixhekhorn:
The current implementation is way too specific - there are infinite many more reasons why we might fail.

If we want to do something - [and I'm not sure we want to as said here](https://github.com/NNPDF/eko/issues/389#issuecomment-2230532768) - we can add a [`np.isfinite`](https://numpy.org/doc/stable/reference/generated/numpy.isfinite.html#numpy-isfinite) to the end of `a`

felixhekhorn:
eko is behaving as expected - the problem which raised the question was addressed in https://github.com/NNPDF/pineappl/commit/9ca3022ef8288b9360584ced5d78f6701f7667e9

PR #391, opened 2024-07-16
Use nf to generate alpha_s

scarlehoff:
The information about nf is not utilized when generating the info files. This PR fixes that.
evolven3fit relies on eko for the info file so once this is merged into a release of eko it should work fine.

Apparently LHAPDF doesn't have a block mechanism for `alpha_s` so at the moment we just provide two values for two nfs for the same Q at the thresholds.

(As far as I can see, the same problem was there in NNPDF4.0 -apfel + evolven3fit- and NNPDF3.1 -apfel + evolvennfit-, so nobody ever noticed...)

### DIFF
--- a/src/ekobox/info_file.py
+++ b/src/ekobox/info_file.py
@@ -9,6 +9,7 @@ from eko import couplings
 from eko.io.runcards import OperatorCard, TheoryCard
 
 from .genpdf import load
+from .utils import regroup_evolgrid
 
 
 def build(
@@ -16,19 +17,19 @@ def build(
     operators_card: OperatorCard,
     num_members: int,
     info_update: dict,
-):
-    """Generate a lhapdf info file from theory and operators card.
+) -> dict:
+    """Generate a lhapdf info file.
 
     Parameters
     ----------
-    theory_card : dict
+    theory_card :
         theory card
-    operators_card : dict
-        operators_card
-    num_members : int
+    operators_card :
+        operators card
+    num_members :
         number of pdf set members
-    info_update : dict
-        info to update
+    info_update :
+        additional info to update
 
     Returns
     -------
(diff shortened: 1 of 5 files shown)


src/ekobox/info_file.py
```diff
@@ -68,19 +68,13 @@ def build(
...
-        dtype=float,
-    )
-    template_info["AlphaS_Vals"] = alphas_values.tolist()
-    template_info["AlphaS_Qs"] = np.array(
-        [mu for mu, _ in operators_card.mugrid]
-    ).tolist()
+
+    alphas_values = []
+    alphas_qs = []
+    for mu, nf in operators_card.mugrid:
```
felixhekhorn, line 74: there might be lurking another bug around:
- see the counter part for the EKOs [here](https://github.com/NNPDF/eko/blob/38cc28c5c6c646118b674a51023b7a816d1d080e/src/ekobox/evol_pdf.py#L98)
- since we are having tuples there is no natural sorting any longer implied - moreover, here you are relying on the opcard which may have an arbitrary order (as it comes from the user)

thus I suggest a stronger check:
- we first sort the `mugrid` by `nf`and then `mu` (as done for the EKOs)
- we check that we can actually dump to LHAPDF, i.e. there are no conflicting (=overlapping) mu ranges
- the edge case of one overlapping mu point needs to be checked with LHAPDF, i.e. it allows such a thing and does not crash on reading - then we need to decide something, e.g. keep them as suggested by @scarlehoff

felixhekhorn (changes requested):

src/ekobox/info_file.py
scarlehoff: > since we are having tuples there is no natural sorting any longer implied - moreover, here you are relying on the opcard which may have an arbitrary order (as it comes from the user)

This version is also more likely to have the correct order than the previous one (which was taking the Qs and the values of alpha_s independently)

> we check that we can actually dump to LHAPDF, i.e. there are no conflicting (=overlapping) mu ranges
> the edge case of one overlapping mu point needs to be checked with LHAPDF

All our PDFs from 3.1 (at least) had that overlapping point so it should be fine.

That said, feel free to change it around. I've used this for `NNPDF40MC_nnlo_as_01180` if you want to have a look a PDF with different values of alpha_s for the same Q:

```
 [4]: from lhapdf import mkPDF
   ...: aa = mkPDF("NNPDF40MC_nnlo_as_01180")
   ...: print(aa.alphasQ(1.51))
   ...: print(aa.alphasQ(1.51+1e-13))
   ...: print(aa.alphasQ(1.51-1e-13))

0.34671367444940654
0.3467136744493946
0.3467136744494207
```

felixhekhorn: @scarlehoff can you please have a look to cb6953df if that makes sense?

scarlehoff:
> I'm now relying on the passed opcard and we better hope that the opcard which is given as argument and the EKO are compatible (i.e. they have the same scales), something we don't explicitly check ...

I guess this is fine unless you give a wrapper that does the full thing. But given that the info file generation is its own thing it makes sense to allow for extra flexibility

(sorry for commenting in the commit, I just followed the link in the email and didn't pay attention :___ )

scarlehoff:
I just rerun this after your changes (with the same eko from before) and seems to work the same

felixhekhorn (approved):

PR #385, opened 2024-05-31
Rustify ome.u.s.{as1, as2}

tgiani:
addressing #383 (ome.u.s.as1 and ome.u.s.as2)

PR #394, opened 2024-07-19
Recover SV, ev method, backwards method

felixhekhorn:
1. recover SV: a "major" change is to use always `Lsv` in the callbacks
2. recover evolution method (iterate-exact, etc.) - trivial thanks to #395
3. recover backwards method - trivial thanks to #395


tgiani:
@felixhekhorn indeed let me first complete the unit tests in #385, it is well likely that there is some little bug in ```as2.rs``` that will be spotted by doing that

tgiani:
@felixhekhorn there was indeed a bug in ```as2```. Still need to finish the unit test, maybe there will be more

felixhekhorn:
> @felixhekhorn there was indeed a bug in `as2`. Still need to finish the unit test, maybe there will be more

let's complete the unit tests, but the benchmark numbers seem now one on one to be the same: https://github.com/NNPDF/eko/actions/runs/10829978519/job/30048696220 :tada:

Issue #383, opened 2024-05-29
Oxidation roadmap


felixhekhorn:
- [x] ad.u.s.as1 #189 
- [ ] ad.u.s.{as2,as3} #369 
- [ ] ome.u.s.as1 and ome.u.s.as2 + VFNS
- [ ] recover SV
- [ ] ad.p.s.*
- [ ] recover solution methods
- [ ] ad.u.s.as4 and ad.u.s.as4.fhmruvv
- [ ] ad.u.s.*aem* + QED
- [ ] ad.u.t.*
- [ ] ome.u.s.as4
- [ ] anything left

PR #395, opened 2024-07-19
Pass evolution + backwards method internally via int

felixhekhorn:
SV mode is cast to integer internally - we should do the same for evolution and backwards instead of using strings. We know this is better, meaning faster, for numba as numba prefers integer over strings (for obvious reasons). This also makes it trivial for #383 to pass it through rust as is proven by #394 .

This is a purely internal change and the outside, i.e. op cards, is unaffected.

This is touching the core of eko so a solid review is appreciated; explicitly this interacts with QED, N3LO and backwards


src/eko/kernels/singlet_qed.py
```diff
@@ -96,7 +97,7 @@ def dispatcher(
     e_s : numpy.ndarray
         singlet EKO
     """
-    if method in ["iterate-exact", "iterate-expanded"]:
+    if method in [EvoMethods.ITERATE_EXACT, EvoMethods.ITERATE_EXPANDED]:
```
felixhekhorn, line 100: @niclaurenti I think `iterate-expanded` should be dropped here (same for V)

src/eko/evolution_operator/operator_matrix_element.py
```diff
@@ -20,6 +21,33 @@
 logger = logging.getLogger(__name__)
 
 
+class BackwardsMethods(enum.IntEnum):
```
giacomomagni, line 24: maybe `MatchingMethods` would be better as `BackwardsMethods.FORWRD` is a bit of a contraddciton

src/eko/evolution_operator/operator_matrix_element.py
```diff
@@ -241,7 +271,10 @@ def labels(self):
             logger.warning("%s: skipping non-singlet sector", self.log_label)
         else:
             labels.append((200, 200))
-            if self.is_intrinsic or self.backward_method is not None:
+            if (
+                self.is_intrinsic
```
giacomomagni, line 275: Do we already have a PR to drop this `intrinsic_range` once for all?

src/eko/evolution_operator/operator_matrix_element.py
felixhekhorn: agreed - done in c88627f7

src/eko/evolution_operator/operator_matrix_element.py
felixhekhorn: #295

src/eko/scale_variations/__init__.py
```diff
@@ -6,40 +6,41 @@
...
+        int representation
 
     """
     if s is not None:
         return Modes[s.value]
     return Modes.unvaried
 
 
-class ModeMixin:
+class ScaleVariationModeMixin:
```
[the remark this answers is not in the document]
felixhekhorn: as you can see I did it as a first thing, because I expected a name clash between "SV mode" and "ev mode" - that didn't happen in the end, but I prefer the explicit name anyway - so I suggest to keep the change unless there are reasons against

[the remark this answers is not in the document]
giacomomagni: well it's already scoped in the proper module, but fine to keep it

src/eko/kernels/singlet_qed.py
felixhekhorn: done in e40acaec

Issue #379, opened 2024-05-22
Define MSRV


felixhekhorn:
- expose information
- propagate to CI
- test in CI

alecandido:
https://github.com/foresterre/cargo-msrv

felixhekhorn:
https://doc.rust-lang.org/cargo/guide/continuous-integration.html#verifying-rust-version

felixhekhorn:
Fixed with #545

PR #400, opened 2024-08-12
EKO Rust reader v2

felixhekhorn:
Second attempt of #97 and restart of #260 

- comparison to #260:
   - I started from the operator loading and not the metadata loading, so the former is now available the latter not
   - [`serde_yaml`](https://github.com/dtolnay/serde-yaml) used there and [in `pineappl`](https://github.com/NNPDF/pineappl/blob/046a59a02a9ed1fbdfc080a044ad25f43a26acfd/pineappl_cli/Cargo.toml#L39) is dead, the maybe-successor [`serde_yml`](https://github.com/sebastienrousseau/serde_yml) is still 0.0 which I didn't dare to use and so I opted of [`yaml-rust2`](https://github.com/Ethiraric/yaml-rust2)
 - relation to #296: this is kept in mind, i.e. we allow to read an already opened eko from dir
 - relation to #97:
   - tools:
     - tar :heavy_check_mark: 
     - yaml: see comment above
     - npz files: same as #260
   - implementation goals:
     - operator loading :heavy_check_mark: 
     - metadata loading: TODO - if we want to do small PRs can even be done later ...
     - "operators headers syncer": I don't know what is meant
- as said by @alecandido it's not wise to maintain two implementation (both in Rust and Python). The structure using `inventories` is a direct mirror of the Python side to keep this idea in mind. The actual Python interface will not be in this PR, but possibly in a later.
- that being said: I intend to have `dekoder` as a pure Rust package and put the Python interface somewhere else (e.g. into `eko`). This follows another idea of @alecandido to keep a dedicated Python interface layer to not clutter the actual library (see also `pineappl` and `pineappl_py`) (or any additional interfaces)
- a dedicated memory management (as was the target of #242) is not necessary here, as we can rely on the Rust features: i.e. it is the users responsibility to explicitly keep the 4D object in scope as long as he needs (or else Rust will clean it up). ~~This idea leads to the following structure:~~ Rust passes ownership on return (see e283553e)
- The short term goal is to expose sufficient features for PineAPPL (not reached yet here and may be done in a second PR). One consequence is that we will of course preserve our idea of `EvolutionPoint` and PineAPPL will have to deal with that


crates/dekoder/src/inventory.rs
```diff
@@ -14,7 +14,9 @@ const HEADER_EXT: &'static str = "*.yaml";
 /// Header type in an inventory.
 pub(crate) trait HeaderT {
     /// Load from yaml.
-    fn load_from_yaml(yml: &Yaml) -> Result<Box<Self>>;
+    fn load_from_yaml(yml: &Yaml) -> Result<Self>
+    where
+        Self: Sized;
```
alecandido, line 19: It should not be needed, as `Sized` is assumed by default. Instead, you should use `?Sized` if it is possible to be unsized.

But I may be missing something

felixhekhorn: this is what the compiler suggested me and the only way it works

alecandido: Ok, it seems that a trait `Self` is the only exception to the general rule

> All type parameters have an implicit bound of Sized. The special syntax ?Sized can be used to remove this bound if it’s not appropriate.

> The one exception is the implicit Self type of a trait. A trait does not have an implicit Sized bound as this is incompatible with [trait object](https://doc.rust-lang.org/book/ch17-02-trait-objects.html)s where, by definition, the trait needs to work with all possible implementors, and thus could be any size.

https://doc.rust-lang.org/std/marker/trait.Sized.html

alecandido: With a caveat

> Although Rust will let you bind Sized to a trait, you won’t be able to use it to form a trait object later:

(same page).

But if you're not planning to use dynamic dispatch, then there is no problem :)

felixhekhorn:
I'm now ready for some feedback - someone please?. While @alecandido would be an ideal candidate we can not assume something on his availability - so @cschwan could you please take a look? e.g. on my usage of Rust :upside_down_face:

crates/dekoder/tests/data/v0.15.tar
cschwan, line 1: I strongly suggest not to put any binary files into the repository and instead load them in the CI, for instance like that: https://github.com/NNPDF/pineappl/blob/046a59a02a9ed1fbdfc080a044ad25f43a26acfd/.github/workflows/rust.yml#L25-L60.

crates/dekoder/tests/test_load.rs
cschwan, line 1: I suggest you run `cargo clippy`, which points out a couple of warnings. Fixing them will improve the code quality a bit.

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+/// EKO output
+pub struct EKO {
+    /// Working directory
+    path: PathBuf,
+    /// Associated archive path
+    tar_path: Option<PathBuf>,
+    /// allow content modifications?
+    read_only: bool,
```
cschwan, line 115: This probably mirrors how the implementation choices from Python (?), but I'd drop both `tar_path` and `read_only` and instead always allow changes and require an explicit path when writing out the modified EKO. To me it seems like an unnessarily tight coupling between the EKO object and its representation as a file. What was the reasoning behind this?

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,87 @@
+//! Assets manager.
+use glob::glob;
+use hashbrown::HashMap;
```
cschwan, line 3: Why are you using `hashbrown::HashMap` instead of `std::collections::HashMap`? I believe the latter should even link to the former, so to keep the number of depenencies down you should use the standard modules.

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,87 @@
...
+use std::path::PathBuf;
+use yaml_rust2::{Yaml, YamlLoader};
+
+use crate::{EKOError, Result};
+
+/// Headers are in yaml files.
+const HEADER_EXT: &'static str = "*.yaml";
+
+/// Header type in an inventory.
+pub(crate) trait HeaderT {
```
cschwan, line 15: I'd remove this `trait` completely and its use with `From<&Yaml>` and `Eq` which are the standard traits. For ULPS I'd use a `const` that's reasonably large, for instance `64`.

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,87 @@
...
+    /// Load from yaml.
+    fn load_from_yaml(yml: &Yaml) -> Result<Self>
+    where
+        Self: Sized;
+    /// Comparator.
+    fn eq(&self, other: &Self, ulps: i64) -> bool;
+}
+
+/// Value type in an inventory.
+pub(crate) trait ValueT {
```
cschwan, line 25: Similar to `HeaderT` you could replace `ValueT` with `From<&Path>` or even more generally `From<Read>`.

crates/dekoder/tests/test_load.rs
felixhekhorn: sorry, not running `clippy` was a clear oversight on my part - I should add it to pre-commit ... done mostly in 814e2567 ; the remaining warnings concern the MSRV, requiring 1.70 - I guess I can just lift it, right? it is out since [more than a year](https://releases.rs/docs/1.70.0/), so no fancy stuff ...

felixhekhorn: > I should add it to pre-commit ... 

 bb9edd18

> the remaining warnings concern the MSRV, requiring 1.70 - I guess I can just lift it, right? it is out since [more than a year](https://releases.rs/docs/1.70.0/), so no fancy stuff ...

 [d9fe346](https://github.com/NNPDF/eko/pull/400/commits/d9fe346f3b1d949fcaef04a2c9609c1a6040b442)

crates/dekoder/src/inventory.rs
felixhekhorn: ehm @alecandido suggested me to prefer the first over the second, but I can also change - I've no idea :see_no_evil: ... done in d888c339; this should then also be reflected on the `ekore` side ...

crates/dekoder/src/inventory.rs
felixhekhorn: should I then hardcode the file suffix `npz.lz4`? for our scope here is fine I think, but this would basically lift everything into `Inventory`; Pushing this idea, it would make our "values" (in the `Inventory`) always of what is currently called `Operator`, again I think this is fine here as they will always be 4D arrays (possibly with errors). It is only the "keys" which need to be of different types. Do you agree with this plan? (lift everything)

crates/dekoder/tests/test_load.rs
cschwan: Yes, 1.70 isn't problematic. PineAPPL's MSRV is also 1.70.

crates/dekoder/src/inventory.rs
cschwan: Yes!

crates/dekoder/src/inventory.rs
felixhekhorn: I made an attempt in [f0d6a27](https://github.com/NNPDF/eko/pull/400/commits/f0d6a276360d1b61736e4e259f77036148b9dbd8), but it is not working since I don't fully understand Rust yet :innocent: in particular there is a problem with lifetimes: in particular here
https://github.com/NNPDF/eko/blob/f0d6a276360d1b61736e4e259f77036148b9dbd8/crates/dekoder/src/inventory.rs#L50
Rust doesn't know that the error has the correct type and in particular he can not know. The exact error is
```
error[E0277]: `?` couldn't convert the error to `EKOError`
  --> crates/dekoder/src/inventory.rs:50:37
   |
50 |                 K::try_from(&cnt[0])?,
   |                                     ^ the trait `std::convert::From<<K as std::convert::TryFrom<&yaml_rust2::Yaml>>::Error>` is not implemented for `EKOError`, which is required by `std::result::Result<(), EKOError>: std::ops::FromResidual<std::result::Result<std::convert::Infallible, <K as std::convert::TryFrom<&yaml_rust2::Yaml>>::Error>>`
   |
   = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
   = note: required for `std::result::Result<(), EKOError>` to implement `std::ops::FromResidual<std::result::Result<std::convert::Infallible, <K as std::convert::TryFrom<&yaml_rust2::Yaml>>::Error>>`
help: consider introducing a `where` clause, but there might be an alternative better way to express this requirement
   |
30 | impl<K: Eq + for<'a> TryFrom<&'a Yaml>> Inventory<K> where EKOError: std::convert::From<<K as std::convert::TryFrom<&yaml_rust2::Yaml>>::Error> {
   |                                                      ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++

For more information about this error, try `rustc --explain E0277`.
error: could not compile `dekoder` (lib test) due to 1 previous error
```
however the suggested fix doesn't work, because there is again a reference appearing in the where clause and I don't know how to fix that. (I understand the general idea of lifetimes, i.e. `Inventory::load_keys` is promising to `EvolutionPoint::try_from` that it will hold the yaml data the whole time.) Note that we need `TryFrom` instead of `From` since reading yaml is a complicated action.

Can you give me some hint?

cschwan: Try changing `TryFrom<&'a Yaml>` to `TryFrom<&'a Yaml, Error = EKOError>`.

felixhekhorn: yes! done in 599fb855 - such an easy fix :upside_down_face: a) why didn't the compiler already suggested that and b) I did look to [the documentation](https://doc.rust-lang.org/std/convert/trait.TryFrom.html) but there I read it as it only takes one generic argument; can I just specify "the member" `Error` as well?

crates/dekoder/src/inventory.rs
felixhekhorn: done in 1a696a18

crates/dekoder/src/inventory.rs
cschwan: `Error` is called an ['associated type'](https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#specifying-placeholder-types-in-trait-definitions-with-associated-types) of `TryFrom` and may be specified if you need to restrict it, otherwise you can leave it out and it is inferred by the compiler. In this case you had to specify it because otherwise the error type of the return type of `try_from` is unknown and it must be at least convertible into an `EKOError` when it is called in `Inventory::load_keys`; that's what the compiler message said. The suggested `where` clause of the compiler is much less strict (`Error` convertible into `EKOError`) than my suggestion (`Error` IS `EKOError`), so that's fine.

crates/dekoder/src/inventory.rs
alecandido: Well, there were many reasons, mainly listed on the repo README itself: https://github.com/rust-lang/hashbrown (memory, performances, compatibility with `no_std`).

It is maintained by the https://github.com/rust-lang/ org itself, and it is compatible with PyO3 (since a long time).
https://pyo3.rs/v0.22.2/conversions/tables

crates/dekoder/src/lib.rs
felixhekhorn: - done in 8d0d759a
- this was mostly because me/we thinking of an EKO object associated with one thing; and a file is more of "a thing" to me then a directory, but I guess it works just the same
- this basically lifts the file management to the user (from the library) which is just as fine
- the "clean up function" is currently called `destroy` - any suggestions for a better name? just `remove`?

crates/dekoder/src/inventory.rs
felixhekhorn: please tell me if I should revert that commit or not :innocent: (in wise prediction I already put it in a single commit :see_no_evil: )

cschwan: No, the README confirm what I said above:

> Since Rust 1.36, this is now the HashMap implementation for the Rust standard library.

crates/dekoder/src/lib.rs
felixhekhorn: but since in the end we will deliver files, also to PineAPPL, we need to have the `extract` also in Rust. As said above it is now simply the users responsibility to keep track of the file

alecandido:
About YAML, after using for a while I discovered that it is not only "human-friendly", but also plenty of frequently unused features, which make it complex to parse, and thus much slower than JSON.

While you should not care that much about performances (the YAML content is not the big deal), the complexity is still relevant: maintaining a YAML parser is much more complicate, so they tend to be less reliable than JSON counterparts.

Maybe it's not worth to do in this PR (because it would break compatibility, and @scarlehoff would get angry :P), but I'd schedule replacement with JSON for the next breaking change (or another simple, Serde-top-tier format). It will make the dependency more reliable, and eventually your life easier.

crates/dekoder/src/lib.rs
cschwan: OK that makes sense.

felixhekhorn:
> Maybe it's not worth to do in this PR (because it would break compatibility, and @scarlehoff would get pretty angry :P), but I'd schedule replacement with JSON for the next breaking change (or another simple, Serde-top-tier format). It will make the dependency more reliable, and eventually your life easier.

no, not in this PR, but for later I don't mind ...

just to say that at the end of this saga (the rust reader) I expect a major change anyways. E.g. eventually I want to go back to using hashes as file names - as we did in Python. I didn't even attempt it here since I expect the Rust implementation and Python implementation of hashing an object are not the same. Instead here I just rely on scanning the folder and finding the original Python hash on the fly.

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,84 @@
...
+}
+
+impl<K: Eq + for<'a> TryFrom<&'a Yaml, Error = EKOError>> Inventory<K> {
+    /// Load all available entries.
+    pub fn load_keys(&mut self) -> Result<()> {
+        let path = self.path.join(HEADER_EXT);
+        let path = path
+            .to_str()
+            .ok_or(EKOError::KeyError("due to invalid path".to_owned()))?;
+        for entry in glob(path)
```
alecandido, line 33: Since you're only using it here, and with a simple pattern, you could easily drop one dependency (though maintained by @rust_lang org as well), by using https://doc.rust-lang.org/std/fs/fn.read_dir.html and just filtering based on the extension.

I'm pretty sure you did this because it's exactly how it is written in Python. But in Python globbing is part of the standard library :P
(Python comes batteries-included, Rust tends to be minimal - even more minimal with `no_std`, but also the full standard library is not as extended as Python's one)

cschwan:
> About YAML, after using for a while I discovered that it is not only "human-friendly", but also plenty of frequently unused features, which make it complex to parse, and thus much slower than JSON.
> 
> While you should not care that much about performances (the YAML content is not the big deal), the complexity is still relevant: maintaining a YAML parser is much more complicate, so they tend to be less reliable than JSON counterparts.
> 
> Maybe it's not worth to do in this PR (because it would break compatibility, and @scarlehoff would get pretty angry :P), but I'd schedule replacement with JSON for the next breaking change (or another simple, Serde-top-tier format). It will make the dependency more reliable, and eventually your life easier.

Did you actually discover any problem in the past related to YAML (if not in EKO, then in any NNPDF project)? Otherwise this sounds like [bikeshedding](https://news.ycombinator.com/item?id=29019361).

crates/dekoder/Cargo.toml
```diff
@@ -0,0 +1,30 @@
...
+rust-version.workspace = true
+version.workspace = true
+
+[package.metadata.docs.rs]
+rustdoc-args = ["--html-in-header", "doc-header.html"]
+
+[dependencies]
+tar = "0.4.41"
+glob = "0.3.1"
+float-cmp = "0.9.0"
```
alecandido, line 21: I'm becoming more and more careful with dependencies, and `float-cmp` is only used for the `approx_eq!` macro. Maybe is worth to just replace it with a simple function, even defined here.

ULPS are fancy, but even a reasonable relative and absolute accuracies (in AND or OR) should be able to get the job done.
Even the library docs is saying:
> When I was a kid in the ’80s, the programming rule was “Never compare floating point numbers”. If you can follow that rule and still get the outcome you desire, then more power to you. However, ***if you really do need to compare them***, this crate provides a reasonable way to do so.
> https://docs.rs/float-cmp/latest/float_cmp/

(emphasis mine)

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /// List available keys.
+    pub fn keys(&self) -> Vec<&K> {
+        let mut ks = Vec::new();
+        for k in self.keys.values() {
+            ks.push(k);
+        }
+        ks
+    }
```
alecandido, line 59: https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.keys
(referring to `std` API, but I checked that also `hashbrown` has it, in case)

https://doc.rust-lang.org/std/vec/struct.Vec.html#impl-FromIterator%3CT%3E-for-Vec%3CT%3E

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    /// Check if `k` is available (with given precision).
+    pub fn has(&self, k: &K) -> bool {
+        self.keys.iter().any(|it| (it.1).eq(k))
+    }
```
alecandido, line 64: For as long as the "given precision" is not there, you could even use:
https://doc.rust-lang.org/stable/std/collections/struct.HashMap.html#method.contains_key

crates/dekoder/src/inventory.rs
alecandido: Don't get me wrong: it's also fine to use the `std` one, to have one dependency less. I was just explaining that there might be motivations to take that into account (most likely not that relevant here, since you have few big elements, rather than many small ones).

scarlehoff:
Just to make my position clear, I don't mind how much everything changes internally as long as eko is able to say "oh, an older operator, I'll use this compability parser".

Otherwise I'll need to keep two (or more) copies of the ekos in the server which would make me sad. See [Fig. 1](https://static.wikia.nocookie.net/muppet/images/e/ef/Marvinandroid.jpg)

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Load `k` from disk.
+    pub fn load(&self, k: &K, v: &mut Operator) -> Result<()> {
+        // Find key
+        let k = self.keys.iter().find(|it| (it.1).eq(k));
+        let k = k.ok_or(EKOError::KeyError("because it was not found".to_owned()))?;
+        let p = self.path.join(k.0).with_extension("npz.lz4");
+        // Read npz.lz4
+        let mut reader = BufReader::new(FrameDecoder::new(BufReader::new(File::open(&p)?)));
+        let mut buffer = Vec::new();
+        std::io::copy(&mut reader, &mut buffer)?;
```
alecandido, line 75: As for the one above, most likely you don't even need this: just pass the variable returned by `FrameDecoder::new()` to `NpzReader::new()`.

The idea is that readers are not reading immediately, they are a promise that they will read when needed, i.e. they are lazy - only when you will trigger the read with an operation they will actually do the job.
Here you're triggering it twice: with `copy()`, and with `.by_name()`. But you only need the second, and that's why you're essentially undoing `copy()` by wrapping it in a `Cursor` (that will make it similar again to a reader, i.e. a `Read` implementor, by wrapping a memory buffer).

alecandido (commented):
First chunk about dependencies and reading process.

I still have to read `lib.rs`.

alecandido:
> Just to make my position clear, I don't mind how much everything changes internally as long as eko is able to say "oh, an older operator, I'll use this compability parser".
> 
> Otherwise I'll need to keep two (or more) copies of the ekos in the server which would make me sad. See [Fig. 1](https://static.wikia.nocookie.net/muppet/images/e/ef/Marvinandroid.jpg)

@scarlehoff I was partially joking, of course.

What it is possible is to have a deprecation period, in which you keep the legacy behavior, and convert whenever possible. As soon as you don't have that many around any longer, you could switch, and drop the legacy support (still available as a previous version, but just to be used for conversions).

However, in this PR I'd do nothing, because you're already introducing a big new feature, so one thing at a time.

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Evolution scale.
+    pub scale: f64,
+    /// Number of flavors
+    pub nf: i64,
+}
+
+/// 4D evolution operator.
+pub struct Operator {
+    pub op: Option<Array4<f64>>,
+}
```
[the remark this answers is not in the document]
alecandido: You would even get the `Default` implementation for free :P

https://doc.rust-lang.org/std/option/enum.Option.html#impl-Default-for-Option%3CT%3E

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+            ))?
+        } else {
+            yml["scale"].as_i64().ok_or(EKOError::KeyError(
+                "because failed to read scale as float from int".to_owned(),
+            ))? as f64
+        };
+        let nf = yml["nf"]
+            .as_i64()
+            .ok_or(EKOError::KeyError("because failed to read nf".to_owned()))?;
+        Ok(Self { scale, nf })
```
alecandido, line 78: I would consider keeping using `serde-yaml` (death means it's completely stable :P), and then maybe transitioning to its successor (or directly `serde-json`) instead of going through manual parsing...

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+}
+
+/// Operators directory.
+const DIR_OPERATORS: &str = "operators/";
+
+impl EKO {
+    /// Check our working directory is safe.
+    fn check(&self) -> Result<()> {
+        let path_exists = self.path.try_exists().is_ok_and(|x| x);
+        if !path_exists {
```
alecandido, line 106:
```suggestion
        if !self.path.try_exists().is_ok_and(|x| x) {
```
if you're not reusing it, I'd not be scared of inlining it (unless the expression is exaggeratedly complex, and assigning it a name would serve as "documentation").

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    operators: inventory::Inventory<EvolutionPoint>,
+}
+
+/// Operators directory.
+const DIR_OPERATORS: &str = "operators/";
+
+impl EKO {
+    /// Check our working directory is safe.
+    fn check(&self) -> Result<()> {
+        let path_exists = self.path.try_exists().is_ok_and(|x| x);
```
alecandido, line 105: I believe this is equivalent to:
```suggestion
        let path_exists = self.path.exists();
```
> #### See Also
> This is a convenience function that coerces errors to false. If you want to check errors, call [Path::try_exists](https://doc.rust-lang.org/std/path/struct.Path.html#method.try_exists).
> https://doc.rust-lang.org/std/path/struct.Path.html#method.exists

Otherwise, if you want to receive the errors as errors, it could be better to use:
```suggestion
        let path_exists = self.path.try_exists()?;
```
since:
> [Path::exists()](https://doc.rust-lang.org/std/path/struct.Path.html#method.exists) only checks whether or not a path was both found and readable. By contrast, try_exists will return Ok(true) or Ok(false), respectively, if the path was verified to exist or not exist.
> https://doc.rust-lang.org/std/path/struct.Path.html#method.try_exists

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        if !path_exists {
+            return Err(EKOError::NoWorkingDir);
+        }
+        Ok(())
+    }
```
alecandido, line 110: Btw, since it's just about giving back errors, you could chain the whole function:
```suggestion
    fn check(&self) -> Result<()> {
        self.path.exists().then_some(()).ok_or(EKOError::NoWorkingDir)
    }
```

alecandido: Though, you could even consider stopping before, and just returning the boolean (and use the boolean in all the other functions).
But this depends on whether you really want to enforce this error always. In which case, having it in a single place is not a bad idea.

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub fn has_operator(&self, ep: &EvolutionPoint) -> bool {
+        self.operators.has(ep)
+    }
+
+    /// Load the operator at the evolution point `ep` from disk.
+    pub fn load_operator(&self, ep: &EvolutionPoint, op: &mut Operator) -> Result<()> {
+        self.check()?;
+        self.operators.load(ep, op)?;
+        Ok(())
+    }
```
alecandido, line 171: Which is the advantage of loading into the `Operator`?

If you are not trying to save memory allocations (the expensive operation here is loading, more than allocating memory), I'd suggest you to give up on return-by-argument (that's really Fortran style...) and just return the operator as a return value.

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub fn destroy(&self) -> Result<()> {
+        self.check()?;
+        Ok(remove_dir_all(&self.path)?)
+    }
+
+    /// Write content back to an archive and destroy working directory.
+    pub fn write_and_destroy(&self, dst: PathBuf) -> Result<()> {
+        self.write(dst)?;
+        self.destroy()?;
+        Ok(())
```
alecandido, line 122:
```suggestion
        self.destroy().into()
```

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        if !path_exists {
+            return Err(EKOError::NoWorkingDir);
+        }
+        Ok(())
+    }
+
+    /// Remove the working directory.
+    pub fn destroy(&self) -> Result<()> {
+        self.check()?;
+        Ok(remove_dir_all(&self.path)?)
```
alecandido, line 115:
```suggestion
        remove_dir_all(&self.path)
```

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+    /// Write content back to an archive.
+    pub fn write(&self, dst: PathBuf) -> Result<()> {
+        self.check()?;
+        // create writer
+        let dst_file = File::create(&dst)?;
+        let dst_file = BufWriter::with_capacity(128 * 1024, dst_file);
+        let mut ar = tar::Builder::new(dst_file);
+        // do it!
+        ar.append_dir_all(".", &self.path)?;
+        Ok(())
```
alecandido, line 134:
```suggestion
        ar.append_dir_all(".", &self.path).into()
```

Here and in `.destroy()` you're actually receiving a different `Result` from yours. But the `thiserror` declaration above should implement somehow `From` for your `Result`, and I'm pretty sure there is no magic in `?` (other than early returning, and automated `.into()` invocation). Though I didn't tested (but the compiler will tell you immediately).

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+        self.destroy()?;
+        Ok(())
+    }
+
+    /// Write content back to an archive.
+    pub fn write(&self, dst: PathBuf) -> Result<()> {
+        self.check()?;
+        // create writer
+        let dst_file = File::create(&dst)?;
+        let dst_file = BufWriter::with_capacity(128 * 1024, dst_file);
```
alecandido, line 130: Top-level constants?

alecandido:
I didn't tested, but I trust you that that part was already covered by you :)

crates/dekoder/src/inventory.rs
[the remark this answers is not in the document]
felixhekhorn: done in 6c7c470c - no `.iter()` needed

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,84 @@
...
+    /// Check if `k` is available (with given precision).
+    pub fn has(&self, k: &K) -> bool {
+        self.keys.iter().any(|it| (it.1).eq(k))
+    }
+
+    /// Load `k` from disk.
+    pub fn load(&self, k: &K, v: &mut Operator) -> Result<()> {
+        // Find key
+        let k = self.keys.iter().find(|it| (it.1).eq(k));
+        let k = k.ok_or(EKOError::KeyError("because it was not found".to_owned()))?;
```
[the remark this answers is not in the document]
felixhekhorn: done in 5ca8848b

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,84 @@
...
+        // Read npz.lz4
+        let mut reader = BufReader::new(FrameDecoder::new(BufReader::new(File::open(&p)?)));
+        let mut buffer = Vec::new();
+        std::io::copy(&mut reader, &mut buffer)?;
+        let mut npz = NpzReader::new(Cursor::new(buffer))
+            .map_err(|_| EKOError::OperatorLoadError(p.to_owned()))?;
+        let operator: Array4<f64> = npz
+            .by_name("operator.npy")
+            .map_err(|_| EKOError::OperatorLoadError(p.to_owned()))?;
+        v.op = Some(operator);
```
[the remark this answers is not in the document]
felixhekhorn: done in dff75814

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    /// Load `k` from disk.
+    pub fn load(&self, k: &K, v: &mut Operator) -> Result<()> {
+        // Find key
+        let k = self.keys.iter().find(|it| (it.1).eq(k));
+        let k = k.ok_or(EKOError::KeyError("because it was not found".to_owned()))?;
+        let p = self.path.join(k.0).with_extension("npz.lz4");
+        // Read npz.lz4
+        let mut reader = BufReader::new(FrameDecoder::new(BufReader::new(File::open(&p)?)));
```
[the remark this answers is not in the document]
felixhekhorn: This is/was a 1:1 copy from pineappl
https://github.com/NNPDF/pineappl/blob/046a59a02a9ed1fbdfc080a044ad25f43a26acfd/pineappl_cli/src/evolve.rs#L400-L405
and I trusted in @cschwan he knows better Rust then me :upside_down_face:  I did remove `BufReader` here: 787ca42f

crates/dekoder/src/inventory.rs
felixhekhorn: ... while I could remove `BufRead`, I can not drop the copy to `Vec` - there is a type mismatch (`FrameDecoder` has no `Seek`)

crates/dekoder/src/lib.rs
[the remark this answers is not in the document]
felixhekhorn: well you know this `struct` will eventually acquire a second member `err`; and surely `op` has to be public because that's what users want

crates/dekoder/src/lib.rs
```diff
@@ -0,0 +1,172 @@
...
+/// 4D evolution operator.
+pub struct Operator {
+    pub op: Option<Array4<f64>>,
+}
+
+impl Operator {
+    /// Empty initializer.
+    pub fn new() -> Self {
+        Self { op: None }
+    }
```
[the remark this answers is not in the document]
felixhekhorn: done in 1656cfc4 - just to say: I started with the `new` method and then clippy said I should add a `default`; but yes they are strict aliases

crates/dekoder/src/lib.rs
felixhekhorn: done in feca19af

crates/dekoder/src/lib.rs
felixhekhorn: I opted for this way, because it is precisely suggested there

> Warning: this method may be error-prone, consider using [try_exists()](https://doc.rust-lang.org/std/path/struct.Path.html#method.try_exists) instead! It also has a risk of introducing time-of-check to time-of-use (TOCTOU) bugs.

and since we may eventually do want to run in parallel, this could be an issue

crates/dekoder/src/lib.rs
felixhekhorn: yes, I do want that error early - to better reflect this I renamed the function into `assert_working_dir` here 410dcfa1

crates/dekoder/src/lib.rs
felixhekhorn: Not possible:
```
error[E0308]: mismatched types
   --> crates/dekoder/src/lib.rs:108:9
    |
106 |     pub fn destroy(&self) -> Result<()> {
    |                              ---------- expected `std::result::Result<(), EKOError>` because of return type
107 |         self.assert_working_dir()?;
108 |         remove_dir_all(&self.path)
    |         ^^^^^^^^^^^^^^^^^^^^^^^^^^ expected `Result<(), EKOError>`, found `Result<(), Error>`
    |
    = note: expected enum `std::result::Result<_, EKOError>`
               found enum `std::result::Result<_, std::io::Error>`
help: use `?` to coerce and return an appropriate `Err`, and wrap the resulting value in `Ok` so the expression remains of type `Result`
    |
108 |         Ok(remove_dir_all(&self.path)?)
    |         +++                          ++
```

crates/dekoder/src/lib.rs
felixhekhorn: done in 8eb8775c - no `.into()` needed

crates/dekoder/src/lib.rs
felixhekhorn: `.into()` doesn't work either (same as ["plain" return](https://github.com/NNPDF/eko/pull/400#discussion_r1724616312) doesn't) - I changed to the shortest form `Ok(blub?)` here 8eb8775c

crates/dekoder/src/lib.rs
felixhekhorn: done in 229670a3

crates/dekoder/src/lib.rs
felixhekhorn: as said [on top](https://github.com/NNPDF/eko/pull/400#issue-2461037962) I'd like to keep it this way, because then we don't need to deal with the memory problem here, but it is the users responsibility. And the crucial thing is that I don't hold the object and so I have to take it by reference. Yes, this is Fortran style and that's the price we have to pay to get the memory management "for free". I don't know yet how this will play out in the Python interface - maybe we have to copy it to the other side and delegate again it's management to the user

Cargo.toml
```diff
@@ -18,5 +18,5 @@ edition = "2021"
 keywords = ["physics"]
 license = "GPL-3.0-or-later"
 repository = "https://github.com/NNPDF/eko"
-rust-version = "1.60.0"
+rust-version = "1.70.0"
```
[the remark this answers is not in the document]
felixhekhorn: yes, this version was suggested to me by [clippy](https://github.com/NNPDF/eko/pull/400#discussion_r1723062234) so I'm pretty sure it is fine - we can always add that check later. I added the link to the relevant issue #379

crates/dekoder/src/inventory.rs
alecandido: It makes sense for `FrameDecoder` to have only `Read`, because it's implementing a stream out (while `Seek` would be really unnatural).

However, in principle `.npy` files don't need `Seek`, it's only for `.npz`:
https://docs.rs/ndarray-npy/latest/ndarray_npy/trait.ReadNpyExt.html

crates/dekoder/src/lib.rs
alecandido: > Note that while this avoids some pitfalls of the exists() method, it still can not prevent time-of-check to time-of-use (TOCTOU) bugs. You should only use it in scenarios where those bugs are not an issue.

crates/dekoder/src/lib.rs
alecandido: You don't need a separate object to do that. Even without references, Rust has move semantics by default (the C++ move), so if you return the object, the control is flowing to the caller. And if you pass the object by value to another function, the control is passed inside.
The ownership rule is made exactly for this: move memory without copying, avoiding multiple references to cause memory issues.

You can just return, and stay safe and efficient.

> I don't know yet how this will play out in the Python interface - maybe we have to copy it to the other side and delegate again it's management to the user

This is more complicated. For as long as the manipulation is in Rust, you can return an object with the hidden reference to Python, and just call the Rust API to manipulate memory. This is exactly what NumPy is doing in C (it's never giving back a Python list, unless forced with `.tolist()` - but it's constructing a rich API to manipulate its own objects).
However, if you want a NumPy array you will to "translate" it somehow, from a Rust ndarray one. This could be done efficiently, as Python has standardized the concept of memory buffers (essentially, you describe the structure of your buffer, and pass a reference to it to another object). But you don't want to do this manually - just hope that the Rust `numpy` crate is doing it for you.

crates/dekoder/src/lib.rs
alecandido:
```suggestion
        remove_dir_all(&self.path).into()
```
?

alecandido: Sorry, I just saw the other one https://github.com/NNPDF/eko/pull/400#discussion_r1724625978

alecandido:
I will review again the `Ok(another_result?)` pattern, since there should also be another way to convert between results.

As the amount of characters, that's clearly one of the easiest. But it's still a bit less explicit, because conceptually they're two operations instead of a single one (though I hope they're getting optimized to the same binary).

In any case, it's really a detail. It's just for better understanding the mechanics of `?`, rather than for a clear advantage.

crates/dekoder/src/inventory.rs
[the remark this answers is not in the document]
cschwan: Using two `BufReader` seems strange indeed. I checked with `strace` that in PineAPPL only the inner `BufReader` is really necessary to minimize the number of `read` syscalls. Of course, you can always leave them out completely.

crates/dekoder/src/inventory.rs
felixhekhorn: done in c6784d9b

crates/dekoder/src/inventory.rs
felixhekhorn: mh? but we are reading npz
https://github.com/NNPDF/eko/blob/c6784d9bd9274af843d4f4ae384d3f40d0c8cbe3/crates/dekoder/src/inventory.rs#L71

crates/dekoder/src/lib.rs
felixhekhorn: ok  - done in 5be3f186 (with the compressed version from below)

felixhekhorn:
> I didn't tested, but I trust you that that part was already covered by you :)

the " :heavy_check_mark: "* does (in case I forget) :upside_down_face: (* in the inline preview the thing is properly green, but not in the preview tab ...)

alecandido:
> (* in the inline preview the thing is properly green, but not in the preview tab ...)

This just depends on the font rendering the emoji xD

crates/dekoder/Cargo.toml
felixhekhorn: done in c81bb08b

crates/dekoder/src/inventory.rs
felixhekhorn: ehm - I just use an eko which I most likely created when merging the many PR some time ago as I running the LO LHA several times, so I believe my test object is v0.15 (i.e. master); again I copied the code blindly from PineAPPL and it works - I believe
https://github.com/NNPDF/eko/blob/c81bb08b1f219cf010cf1764b8257925b0eab891/crates/dekoder/tests/test_load.rs#L83
note that I have a single test object and with that I'm happy for the moment :see_no_evil:

crates/dekoder/src/lib.rs
felixhekhorn: Mmm not sure ...
a) as you can see above I already had to do some magic to make it work (because for an unknown reason in my test case the scale is saved as `10000` (note the missing `.`) and `as_f64` doesn't like that, so I don't know what `serde` would do
b) for the time being it is rather simple, so not sure we want to trade it for a dead crate
c) since we already planning (in the long/mid term) to replace it, I wonder if we can live with a non-optimal version for the moment

alecandido: > so I don't know what `serde` would do

The right thing, because `serde` is aware that the final type will be `f64` (is based on a schema, built by the `struct` definitions), so it will attempt a cast (or better, loading in the correct way from the beginning - I hope, though the specific loading is up to the deserializer implementation).

Instead, `yaml-rust2` is free-form, so it's inferring the type of the structure to create from the content itself. It could be handy for some applications, but not that much when the structure is known and fixed.

> b) for the time being it is rather simple, so not sure we want to trade it for a dead crate

The crate alive has potential to become quickly more dead than the previous one: it seems to be a fork of another one (https://github.com/chyh1990/yaml-rust) that died because of `serde-yaml` (the second contributor was the author of `serde-yaml`), and it has just been resurrected recently.
image

Being maintained is relevant for bug fixes, but being popular and used is telling you that there are fewer bugs around (and, if dead, no new one).

> c) since we already planning (in the long/mid term) to replace it, I wonder if we can live with a non-optimal version for the moment

Definitely. Mine is just an opinion, whose summary is something like: "they are both suboptimal, but one of the two is more painful than the other" xD

(I guess that in the end you'll want `serde` - but for the time being you can do whatever)

crates/dekoder/src/inventory.rs
alecandido: Can you check the content of the EKO you're testing? Just run `tar -t myeko.tar`.

felixhekhorn: I was about to say, that the problem is with `.npz` and I don't know if I can read that - but yes I can
grafik
and the thing is really called `operator.npy` - and it has the errors

crates/dekoder/src/lib.rs
felixhekhorn: > but for the time being you can do whatever

I would say let's reconsider the moment I start parsing the metadata (which I would not attempt in this PR)

crates/dekoder/src/inventory.rs
alecandido: I was only interested in the previous file, the `EFM...ks=.npz`. The one inside should be an arbitrary string, so it doesn't have to have the `.npy` extension at all...

Impressive that the GUI Gnome utility is aware of `.npz` files.

alecandido: It seems like it is an internal implementation detail:

> The .npz file format is a zipped archive of files named after the variables they contain. The archive is not compressed and each file in the archive contains one variable in .npy format.

Being a zip, it explains why Gnome can open it.

I wonder whether we should stop using `.npz` files, and just use the `<hash>.npy` and `<hash>-err.npy` to avoid one archive layer...

This would also be a breaking change, so it should be scheduled for later.

In any case, pay attention that your implementation is making use of a (documented) implementation detail. And it would fail if there is no error around, because in that case the EKO would just contain `<hash>.npy` directly.
Sorry for the confusion, I had no idea of how an `.npz` was implemented at the time (I could read the note, most likely it was already there...), and it seemed a good idea to keep them together. Not that good, in the hindsight...

felixhekhorn: > so it should be scheduled for later.

yes, as said already: I'm expecting at the end of this journey a breaking change and we can break even further

> In any case, pay attention that your implementation is making use of a (documented) implementation detail. And it would fail if there is no error around, because in that case the EKO would just contain `<hash>.npy` directly.

yes, but for the moment I'm happy with my single test case - this is just the first step.

felixhekhorn: in the end do you want me to change something?

crates/dekoder/src/lib.rs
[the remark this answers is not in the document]
felixhekhorn: > well you know this `struct` will eventually acquire a second member `err`; 

done in 1177a578

alecandido:
Ah, just for some more fun, the `:white_check_mark:` is actually a green box with GitHub's webfont :white_check_mark:

alecandido:
> Did you discover any problems in the past related to YAML - if not in EKO, then in any NNPDF project? Otherwise this sounds like [bikeshedding](https://news.ycombinator.com/item?id=29019361).

@cschwan I hit myself [the Norway problem](https://www.bram.us/2022/01/11/yaml-the-norway-problem/) in a project within NNPDF (don't remember exactly which and when).
And PyYAML caused a lot of problems with object serialization (it tries to serialize unknown objects as custom types - unluckily, NumPy scalars were also custom objects, as 0-rank arrays, so we needed a lot of `float(array[idx])`) - while now is causing problem for lack of reliable libraries in Rust.

Outside NNPDF, we got a huge speed-up by dropping YAML in favor of JSON (parsers are much simpler and much more optimized).

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,86 @@
...
+    pub(crate) keys: HashMap<OsString, K>,
+}
+
+impl<K: Eq + for<'a> TryFrom<&'a Yaml, Error = EKOError>> Inventory<K> {
+    /// Load all available entries.
+    pub fn load_keys(&mut self) -> Result<()> {
+        for entry in read_dir(&self.path)? {
+            // is header file?
+            let entry = entry?.path();
+            if !entry.extension().is_some_and(|ext| ext.eq(HEADER_EXT)) {
```
cschwan, line 30: Use `ext == HEADER_EXT` instead of `ext.eq(HEADER_EXT)`.

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /// List available keys.
+    pub fn keys(&self) -> Vec<&K> {
+        self.keys.values().collect()
+    }
+
+    /// Check if `k` is available (with given precision).
+    pub fn has(&self, k: &K) -> bool {
+        self.keys.iter().any(|it| (it.1).eq(k))
```
cschwan, line 57: Same comment for `eq`.

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    /// Load `k` from disk.
+    pub fn load(&self, k: &K) -> Result<Operator> {
+        // Find key
+        let k = self
+            .keys
+            .iter()
+            .find(|it| (it.1).eq(k))
```
cschwan, line 66: Also here.

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,86 @@
...
+
+use crate::{EKOError, Operator, Result};
+
+/// Headers are in yaml files.
+const HEADER_EXT: &str = "yaml";
+
+/// Assets manager.
+pub(crate) struct Inventory<K: Eq + for<'a> TryFrom<&'a Yaml, Error = EKOError>> {
+    /// Working directory
+    pub(crate) path: PathBuf,
```
cschwan, line 19: Remove `pub(crate)` and instead write a `pub(crate)` constructor.

crates/dekoder/src/inventory.rs
```diff
@@ -0,0 +1,86 @@
...
+    /// Available keys
+    pub(crate) keys: HashMap<OsString, K>,
```
cschwan, line 21: Remove `pub(crate)`.

alecandido (approved):
I agree with the last @cschwan review, but they are all tiny improvements (that I'd recommend applying).

Other than that, the PR should be fine, for EKOs containing integration errors.
As I said, most of the EKO (if not all those produced for FK tables) should not contain the errors, since we switched the default to optimize disk and memory consumption (these errors are never used in the pineline). So, in practice, I expect `dekoder` not to be ready for production yet.

In any case, the npz/npy problem could be faced in another PR, possibly changing at the same time the format, getting rid of the npz.
This would be a breaking data format change, so, in principle, it should be scheduled for a major release, and even a data format version bump. However, as said above, the EKOs that are going to be broken I expect to be unused, and you can check it quickly on existing EKOs with something like `tar -t myeko.tar | grep npz`. I'd suggest doing this, before breaking something to @scarlehoff :P

crates/dekoder/src/inventory.rs
felixhekhorn: d394978d

crates/dekoder/src/inventory.rs
felixhekhorn: d394978d

crates/dekoder/src/inventory.rs
felixhekhorn: d394978d

crates/dekoder/src/inventory.rs
felixhekhorn: bf132ea3

crates/dekoder/src/inventory.rs
felixhekhorn: bf132ea3

crates/dekoder/src/eko.rs
```diff
@@ -0,0 +1,140 @@
...
+        let scale = yml["scale"].as_f64();
+        let scale = if scale.is_some() {
+            scale.ok_or(EKOError::KeyError(
+                "because failed to read scale as float".to_owned(),
+            ))?
+        } else {
+            yml["scale"].as_i64().ok_or(EKOError::KeyError(
+                "because failed to read scale as float from int".to_owned(),
+            ))? as f64
+        };
```
cschwan, line 38: Does this make sense? In the branch `if scale.is_some()` the variable `scale` is always `Some(...)` and thus `ok_or` never fails. In the else branch you're trying again with an integer, so it seems to me you want

```suggestion
        let scale = scale.map_or_else(
            || {
                let int = yml["scale"].as_i64().ok_or(EKOError::KeyError(
                    "because failed to read scale as float from int".to_owned(),
                ))?;
                Ok(i16::try_from(int)
                    // UNWRAP: integers unrepresentable as floating-point numbers are wrong inputs
                    .unwrap()
                    .into())
            },
            Ok::<_, EKOError>,
        )?;
```

felixhekhorn: Sorry for merging too earlier. Yes, the `is_some` and `ok_or` was redundant, but this was my fault, because I wanted to ban all `unwrap` :upside_down_face: Your code works the same, but to me it looks more complicated ... would it be ok to just replace the first `ok_or` with `unwrap`? I'm also concerned about the integer range: `i16` might be too small (we need to read $Q^2$) - and I hope `as` does the same job as `into` ...

cschwan: You shouldn't ban `unwrap`, it's the right tool for [unrecoverable errors](https://doc.rust-lang.org/book/ch09-03-to-panic-or-not-to-panic.html). As for the integer case: 1) have you verified that this is necessary? I agree that `i16::MAX` is probably too close to the representable values that we're interested in. 2) `as` doesn't do the same as `into`, the former is approximate, the latter is exact. I believe we don't want approximate conversions.

alecandido: > and I hope `as` does the same job as `into` ...

It's slightly different:

> `as` is most commonly used to turn primitive types into other primitive types, but it has other uses that include turning pointers into addresses, addresses into pointers, and pointers into other pointers.
> https://doc.rust-lang.org/std/keyword.as.html

is more for primitives (as `f64`). While `.into()` is a method of a structured object, even represented by a trait (`Into`, paired with `From`).

> `as` can be seen as the primitive for `From` and `Into`: `as` only works with primitives (`u8`, `bool`, `str`, pointers, …) whereas `From` and `Into` also works with types like `String` or `Vec`.

(same page as above)

So, it is correct to use `as` for casting to `f64` (taking into account all the limitations of casting)

cschwan: @alecandido Either case is better than my solution. However, I'm not sure if reading integers is really necessary. The `... as f64` is possibly problematic as it's approximate.

alecandido: I was not aware of the lossless conversion with `.into()`. However, notice that it is not implemented from `i64` -> `f64` (and it makes sense, you definitely need more byte to represent an integer as a float without losing anything).
image

felixhekhorn: - I like @alecandido 's suggestion from above as it is clear and concise - done in c9bb4ac0
- as can be seen in the list above there is no `From<i64>` so we need `as`; and I'm willing to accept an approximate conversion (which is _always_ there for int->float except a finite number of cases) as we don't have any physical precision to that numerical accuracy
- the casting from int branch is definitely needed and in fact the only tested branch with my test EKO which I got from running the LHA benchmarks
- in fact, this exposes another problematic implementation detail of yaml in Python (as for disadvantages of yaml) : the number in the yaml file is really `10000` - note the missing `.` - despite being originally a float
- as for the int range, stupid me: actually the calculation is easy :upside_down_face: we surely have scales large then $Q = 2^8\text{GeV}$

cschwan: This integer branch could be avoided if you'd used `serde` instead of manually parsing using `yaml`. In that case the Rust structs have the full type information and most of the YAML problems disappear.

felixhekhorn: yes, yes @alecandido was already saying the same above :see_no_evil: - [I will try to do it later](https://github.com/NNPDF/eko/pull/400#discussion_r1724843377)

cschwan: Oh I missed that part of the discussion, but then you know that you should do it :smile: .

PR #405, opened 2024-09-13
Introduce docformatter and ruff

felixhekhorn:
I apologize for this gigantic PR, but I believe we want this and hopefully nothing should have happened.

1. By stalking @alecandido I found out about [docformatter](https://github.com/PyCQA/docformatter). We had so far only a soft rule that docstrings should follow the numpy convention, with this PR it is becoming a strict requirement and docformatter will do the job. Thus ~95% of the changes are a) removing new lines in docstrings or b) adding a `.` here and there. I did not go through the whole documentation to see, if everything went well (and for sure there are some hickups here and there), but I trust in the docformatter developers.
2. while doing that I also found out that [pydocstyle](https://github.com/pycqa/pydocstyle) got deprecated (and which I thought would do the job) in favor of [ruff](https://docs.astral.sh/ruff/), which is then introduced along side in this PR. This has triggered another few changes as the new linter complained about new things and which I had to adjust by hand.

Not sure on how much of a review is required (given the amount of files), but maybe tell me if you agree with the idea.


alecandido:
Ruff is also doing `pydocstyle` job, but it has been developed as a Pylint replacement. You could really consider getting rid of Pylint, and adopt a unique tool.

In principle, the workflows are Pylint-agnostic, as they rely on `poe lint` (so, you should just replace the script). There is only an irrelevant name of a step mentioning explicitly Pylint
https://github.com/search?q=repo%3ANNPDF%2Fworkflows%20pylint&type=code
(in case, you can always upgrade even that, but it's also not required)

felixhekhorn:
> > Ruff can be used to replace [Flake8](https://pypi.org/project/flake8/) (plus dozens of plugins), [Black](https://github.com/psf/black), [isort](https://pypi.org/project/isort/), [pydocstyle](https://pypi.org/project/pydocstyle/), [pyupgrade](https://pypi.org/project/pyupgrade/), [autoflake](https://pypi.org/project/autoflake/), and more, all while executing tens or hundreds of times faster than any individual tool.
> 
> https://docs.astral.sh/ruff/ https://docs.astral.sh/ruff/integrations/#pre-commit

While pydocstyle and pyupgrade were already dropped, dfe7c791 additionally dropps isort and black (I will need to get used to the fact that black will no longer be around after such a long time :upside_down_face: ) - note that there a few changes in the format, but as long as someone is doing the job I don't care about the actual format

felixhekhorn:
Note that 8bcee93 revealed a naming problem for which I don't have better solution then the one proposed ... other ideas are welcome

giacomomagni (commented):
I had a quick look and indeed the changes are mostly only in docs, with some few variables `l` renamed (?)...
I see you are still keeping `blacken-docs`.

felixhekhorn:
> I see you are still keeping `blacken-docs`.

1149f182 drops `blacken-docs` in favor of the ruff equivalent.

> I had a quick look and indeed the changes are mostly only in docs, with some few variables `l` renamed (?)... 

yes, the new linter complained (also pylint would issue a warning) as it considers `l` an ambiguous variable name (i.e. can be confused with `I` in some fonts); however having a slightly more verbose variable name was an easy fix and even a reasonable one

doc/source/conf.py
```diff
@@ -298,24 +295,6 @@
...
 
 
-# I don't know where and when, but at some point sphinx stopped to detect the documentation
-# hidden below numba. This issue is discussed here https://github.com/sphinx-doc/sphinx/issues/3783
-# pointing to this conf.py:
-# https://github.com/duetosymmetry/qnm/blob/d286cad616a4abe5ff3b4e05adbfb4b0e305583e/docs/conf.py#L71-L93
-# However, it doesn't do the trick truly, but the idea is take from there ...
-# see also
-# https://www.sphinx-doc.org/en/master/usage/extensions/autodoc.html#docstring-preprocessing
-def process_numba_docstring(
```
felixhekhorn, line 308: just for future reference: this PR also drops this hack as it does not seem necessary any longer

src/eko/__init__.py
```diff
@@ -5,3 +5,9 @@
 from .runner import solve
 
 __version__ = version.__version__
+
+__all__ = [
+    "io",
+    "EKO",
+    "solve",
+]
```
felixhekhorn, line 13: for future reference: the linter also required me to add these lines (and similar in other places, e.g. `ekobox`) - which are a step towards #145

giacomomagni (approved):
Okay LGTM.

src/eko/__init__.py
alecandido: Ok, PEP8 is pretty much succinct about the topic...

https://peps.python.org/pep-0008/#public-and-internal-interfaces

PR #416, opened 2024-10-16
Rustify QED

tgiani:
- [x] implement `aem1.rs`, `as1aem1.rs`, `aem2.rs`
- [x] add in `anomalous_dimension::unpolarized::spacelike.rs` missing QED ingredients 
- [x] hook rust kernels into python
- [ ] benchmark against apfel


crates/ekore/src/constants.rs
```diff
@@ -41,3 +52,21 @@ pub const PID_NSM: u16 = 10201;
...
+}
+
+pub fn charge_combinations(nf: u8) -> Vec<f64> {
+    let nu = uplike_flavors(nf) as f64;
+    let nd = (nf as f64) - nu;
+    let e2avg = (nu * eu2 + nd * ed2) / (nf as f64);
+    let vue2m = nu / (nf as f64) * (eu2 - ed2);
+    let vde2m = nd / (nf as f64) * (eu2 - ed2);
+    let e2delta = vde2m - vue2m + e2avg;
+    vec![e2avg, vue2m, vde2m, e2delta]
```
felixhekhorn, line 71: Should we return a dedicate `struct`? We could make `nu` and `nd` part of that ...

tgiani: yes for sure. I was going to add a comment asking for input on how to code this bit properly. I ll add a dedicate `struct`

crates/ekore/src/constants.rs
```diff
@@ -61,12 +61,25 @@ pub fn uplike_flavors(nf: u8) -> u8 {
     nf / 2
 }
 
-pub fn charge_combinations(nf: u8) -> Vec<f64> {
+pub struct ChargeCombinations {
```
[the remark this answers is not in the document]
felixhekhorn: ehm, it is one possibility - but I would have done something more strict: a struct with a single field `nf` and a bunch of methods; `pub fn nu() -> u8`,`pub fn nd() -> u8`,`pub fn e2avg() -> f64`, etc.

crates/ekore/src/harmonics/cache.rs
```diff
@@ -46,6 +46,12 @@ pub enum K {
     Sm21e,
     /// $S_{-2,1}(N)$ odd moments
     Sm21o,
+    /// recursive harmonics
+    S1ph,
+    S2ph,
+    S3ph,
+    S1p2,
+    G3p2,
```
felixhekhorn, line 54: I'm not sure we want to keep those - can we inline them?

tgiani: what do you mean? Could you please provide an example?

felixhekhorn: instead of
```py
    let g3N = c.get(K::G3);
    let g3Np2 = c.get(K::G3p2);
```
we do
```py
    let g3N = c.get(K::G3);
    let g3Np2 = g3N + f;
```
where $f = g_3(N+2)-g_3(N)$ is computed analytically beforehand

tgiani: I am afraid I don t understand. How do you compute f analytically beforehand? Do you mean to move `g_functions::g3(self.n + 2., self.get(K::S1p2))` out of the cache?

crates/ekore/src/constants.rs
```diff
@@ -71,24 +71,31 @@ pub fn uplike_flavors(nf: u8) -> u8 {
...
-    let e2delta = vde2m - vue2m + e2avg;
-
-    ChargeCombinations {
-        e2avg,
-        vue2m,
-        vde2m,
-        e2delta,
+impl ChargeCombinations {
+    pub fn nu(&self) -> u8 {
+        uplike_flavors(self.nf)
```
felixhekhorn, line 79: let's move the function here
```suggestion
        self.nf / 2
```
I think we can drop the exception ...

crates/ekore/src/harmonics/cache.rs
felixhekhorn: > out of the cache?

yes

> How do you compute f analytically beforehand?

Mathematica :upside_down_face: let's try to use a short cut: @giacomomagni can you compute quickly this $f$? (also feel free to give your opinion if we should keep that cache entry or not)

felixhekhorn: so I think
$$f = \frac{6 (n+1) (2 n+1) H_n+n \left(-\pi ^2 (n+1)^2-6 n\right)}{6 n^2 (n+1)^3}$$
which I computed half with pen and paper and half Mathematica and checked for $N=1$, $N=2$

to be more precise: for $g_3$ we could use the above equation and for the others we inline the expressions

tgiani: ok thank you! I ll first try to get things working as they are and then I will add this

felixhekhorn:
We are getting non-sense, but we are running
```
[...]
Evolution: order: (3, 2), solution strategy: iterate-exact, use fhmruvv: True
Exception ignored in: '<numba.core.cpu.CPUContext object at 0x7facc6476db0>'
  File "/home/felix/.cache/pypoetry/virtualenvs/eko-KkPVjVhh-py3.12/lib/python3.12/site-packages/numba/np/linalg.py", line 827, in _check_finite_matrix
    raise np.linalg.LinAlgError(
numpy.linalg.LinAlgError: Array must not contain infs or NaNs.
Exception ignored in: '<numba.core.cpu.CPUContext object at 0x7facc6476db0>'
  File "/home/felix/.cache/pypoetry/virtualenvs/eko-KkPVjVhh-py3.12/lib/python3.12/site-packages/numba/np/linalg.py", line 827, in _check_finite_matrix
    raise np.linalg.LinAlgError(
numpy.linalg.LinAlgError: Array must not contain infs or NaNs.
Evolution: computing operators - 1/5 took: 6.413625 s
[...]
 ─── 
  g  
 ─── 
       x     Q2  eko  eko_error      apfel  percent_error
0  0.001  10000  NaN        NaN  29.559749            NaN
1  0.010  10000  NaN        NaN   7.916446            NaN
2  0.100  10000  NaN        NaN   0.834144            NaN
3  0.500  10000  NaN        NaN   0.016857            NaN
4  1.000  10000  NaN        NaN   0.000000            NaN
```

felixhekhorn:
4507b753 fixed a bug for VFNS and the [LHA benchmarks pass with that](https://github.com/NNPDF/eko/actions/runs/13110399107/job/36572843691), but can you please also check against apfel explicitly?

tgiani:
uhm now if I run the apfel benchmark I m getting 
``` Exception ignored in: '<numba.core.cpu.CPUContext object at 0x7f1cbf34bf50>'
  File "/data/theorie/tgiani/miniconda3/envs/eko/lib/python3.11/site-packages/numba/cpython/numbers.py", line 1100, in complex_div
    raise ZeroDivisionError("complex division by zero")
ZeroDivisionError: complex division by zero
Exception ignored in: '<numba.core.cpu.CPUContext object at 0x7f1cbf34bf50>'```
can you please check if that s the case for you as well?

tgiani:
can you please check if that s the case for you as well?

felixhekhorn (approved):

Issue #328, opened 2023-11-28
On the backwards-compatibilities or future-proofing of ekos


scarlehoff:
As you know, I'm trying now to get rid of all the legacy code from nnpdf for which we no longer have a maintainer. For that one of the things we need to do is to replace apfel with eko (i.e., remove the _new modifier from `evolven3fit_new`). By now I think eko has been time-tested more than enough for this change... and I really would like to be able to kick `evolven3fit` out of the default installation.

however....

There is the issue of backwards compatibility, while eko is faster and more developed than apfel the ekos often cannot be used from one version to the next. This is a big problem since nnpdf releases usually contain O(20) theories (and with N3LO and MHOU we are moving to O(100)).
I understand the rapid development associated with the qed/n3lo/mhou releases meant compromising on some other things but now that that's finished... would it be possible to offer some kind of guarantee that an eko.tar _for the evolution of the PDFs_ computed with a version of eko will be compatible with future versions?

Of course, there might come a time when it is necessary to break backwards compatibility for whatever the reason. Right now NNPDF releases happen more or less every ~ 4 years, so I would like that the theories that are prepared for a release are utilizazble until the next one (without having to update all ekos in all theories).

I'm not asking for bug-by-bug compatibility, but to include a compatibility layer every time the format of the eko.tar would change. I would prefer a compatibility-layer in eko that is able to recognize "oh, this is an eko using the 0.15 format, let's use the 0.15 compatibility layer" rather than a conversion mechanism but I would accept a conversion mechanism as well if it is well tested.

In practical terms (I talked about guarantees before, but what does that mean) I would like a test in the CI in this repository (I can write that) that goes through a set of representative theories and checks that the ekos are still usable...

AleCandido:
#106

scarlehoff:
#214

However, if you think that the current version of the ekos/conversor is already robust enough I can write the test starting from 0.14.

I just want to make sure that people can, at the same time, use the latest version of nnpdf and eko.

AleCandido:
#214 is definitely outdated, a different solution has been applied in the meanwhile (i.e. recomputing when actually needed).

The only possible instability is now related to the runcards content, since it keeps being modified.
However, this should be easy enough to translate every time, by a combination of applying defaults and removing fields.

@scarlehoff write a test for version 2, after bumping this:
https://github.com/NNPDF/eko/blob/8372ec376ff502a8599eefe333783f44fd5bf676/src/eko/version.py#L8
[...]

PR #417, opened 2024-10-17
Support reading v0.13 and v0.14 files

### LINKED ISSUE #398: Drop legacy runner
Left-over from #296

> But before merging I'd like to have a plan for the *legacy* runner removal. This whole PR was meant to provide tools for Pineko to implement a further runner, similar to the *managed* one, but more sophisticated. There should be no reason to keep the *legacy* any further.

_Originally posted by @alecandido in https://github.com/NNPDF/eko/issues/296#issuecomment-1682415864_

there are two legacy modules: `eko.io.legacy` and `eko.runner.legacy`

directly related to #145

felixhekhorn:
Okay, this will need some thinking, as we maybe want to come up with a more future-proof strategy (see e.g. #145 #328)

- this drops the Legacy runner, i.e. closes #398 - this will break many things potentially again, i.e. this is surely another major break to master. `eko.runner.managed.solve` which is the new default does only accept new runcards, i.e. if you need to convert the user has to do that before calling the function (see also below)
- this drops the old file converter which was for files before v0.11 (I think)
- then continuing from #175:
  - we still have `eko.io.runcards.Legacy`: 1. we should rename that to `NnpdfDb` or similar (and maybe move it?) and 2. I guess we need to maintain that also midterm (or even forever)
  - we haven't increased the `__data_version__` since we introduced it, but I guess we should start doing so to disentangle a change in API (`__version__`) from a change in the file format (`__data_version__`). The latter also refers _any_ file in the tar, i.e. including all `.yaml` files. We should increase it to `3` and rely on versions beyond `v0.15` on that (`1` ~ `v0.13`, `2` ~ `v0.14` although the latter was never applied).
- I don't know how to recover from #292 : old ekos might be rotated (on any of the four axes) and we would need to counteract this everywhere ... any ideas? It is true that the old format and the new format are technical the same, but the content might not

cc @scarlehoff 

---

TODO:
- [x] unify loading of unit test assets; i.e. move https://github.com/NNPDF/eko/blob/master/crates/dekoder/tests/data/download.sh somewhere more global and use it for both Rust and Python unit tests
- [x] move all test assets to that script, i.e. to the server
- [x] test v0.13:
  - [x] First create EKO object (the v0.13 test case) in v0.13, use LHA LO FFNS
Do the following things for this EKO object in the newest EKO version from master branch:
  - [x] reading of theory card: check it reads the theory card as a theory card
  - [x] reading of operator card: check it reads the operator card as an operator card
  - [x] reading of the 4D operator:
    - [x] Check the dimensions of the operator (i.e. flavour & inputted x-grid)
    - [x] Use the EKO operator to evolve a PDF using the toy pdf (https://eko.readthedocs.io/en/latest/overview/tutorials/pdf.html#Method-1:-Using-apply_pdf)
    - [x] Check if the numbers you get from the last step are the same as the numbers in the LHA benchmark (can find these in EKO/src/ekomark/benchmark/external/LHA.yaml for different cases)
- [x] test v0.14 (same as for v0.13)
   - [x] First create EKO object (the v0.13 test case) in v0.13, use LHA LO FFNS
Do the following things for this EKO object in the newest EKO version from master branch:
  - [x] reading of theory card: check it reads the theory card as a theory card
  - [x] reading of operator card: check it reads the operator card as an operator card
  - [x] reading of the 4D operator:
    - [x] Check the dimensions of the operator (i.e. flavour & inputted x-grid)
    - [x] Use the EKO operator to evolve a PDF using the toy pdf (https://eko.readthedocs.io/en/latest/overview/tutorials/pdf.html#Method-1:-Using-apply_pdf)
    - [x] Check if the numbers you get from the last step are the same as the numbers in the LHA benchmark (can find these in EKO/src/ekomark/benchmark/external/LHA.yaml for different cases)
- [x] test v0.15 (same as for v0.13, ~~but now reuse the existing v0.15 test case instead of making it again~~ the existing one was not a sane object)
  Do the following things for this EKO object in the newest EKO version from master branch:
  - [x] reading of theory card: check it reads the theory card as a theory card
  - [x] reading of operator card: check it reads the operator card as an operator card
  - [x] reading of the 4D operator:
    - [x] Check the dimensions of the operator (i.e. flavour & inputted x-grid)
    - [x] Use the EKO operator to evolve a PDF using the toy pdf (https://eko.readthedocs.io/en/latest/overview/tutorials/pdf.html#Method-1:-Using-apply_pdf)
    - [x] Check if the numbers you get from the last step are the same as the numbers in the LHA benchmark (can find these in EKO/src/ekomark/benchmark/external/LHA.yaml for different cases)
- [x] check and rerun all tutorials (in the documentation)


scarlehoff:
> (or even forever)

4ever

But I'm happy with moving it to the theory parser in NNPDF (i.e., having an attribute in the `Theory` file which is `return_eko_theory` or something).
But I would say let's do things incrementally, for now let's keep it here, let's make sure things are working and then we move it.

> since we introduced it, but I guess we should start doing so to disentangle a change in API

I agree. The ones in 0.15 are surely a data version different from what we have

>  I don't know how to recover from

_from the nnpdf side_ I don't mind not having this. In practice it would mean that the x-grid is completely fixed. If I understand correctly this feature has been more a convenience for people that could create more custom ekos and then got the 192x LHAPDF grid as the output? (did I understood it correctly @giacomomagni ?)

giacomomagni:
> * this drops the Legacy runner, i.e. closes [Drop legacy runner #398](https://github.com/NNPDF/eko/issues/398) - this will break many things potentially again, i.e. this is surely another major break to master. `eko.runner.managed.solve` which is the new default does only accept new runcards, i.e. if you need to convert the user has to do that before calling the function (see also below)

since we are updating `evolven3fit` we might consider to update as well `pineko`... but it will require a bit of work. 

> * this drops the old file converter which was for files before v0.11 (I think)
 
 here we are safe I think, it's already impossible to read 0.12 (?)
 
> * I don't know how to recover from [Change `manipulate` to act on Operator #292](https://github.com/NNPDF/eko/pull/292) : old ekos might be rotated (on any of the four axes) and we would need to counteract this everywhere ... any ideas? It is true that the old format and the new format are technical the same, but the content might not

so in order to use the old ekos (0.13) you'd need to convert them, can't we add some checks while applying the conversion and warn the user?

scarlehoff:
> so in order to use the old ekos (0.13) you'd need to convert them, can't we add some checks while applying the conversion and warn the user?

Wait, I thought the _pdf_ ekos that we have were already ok without rotation? Or do you mean the _fktables_ ekos?

felixhekhorn:
> But I would say let's do things incrementally, for now let's keep it here, let's make sure things are working and then we move it.

definitely. I'm not sure, if that is better here or in NNPDF - we may want to move it e.g. into `ekobox` though

> since we are updating `evolven3fit` we might consider to update as well `pineko`... but it will require a bit of work.

yes, but it will require anyway work since this is a major change already

> here we are safe I think, it's already impossible to read 0.12 (?)

I think so - that was never requested and so with this PR I put a hard limit on v0.13

> so in order to use the old ekos (0.13) you'd need to convert them, can't we add some checks while applying the conversion and warn the user?

just warning would be doable - but is that sufficient?

> _from the nnpdf side_ I don't mind not having this. In practice it would mean that the x-grid is completely fixed. If I understand correctly this feature has been more a convenience for people that could create more custom ekos and then got the 192x LHAPDF grid as the output? (did I understood it correctly @giacomomagni ?)

> Wait, I thought the _pdf_ ekos that we have were already ok without rotation? Or do you mean the _fktables_ ekos?

that's the thing: eko can not know for what you use the it - and FK-ekos are rotated by default, even on multiple axes (I think). The main post-fit eko might be safe by chance. You can still rotate the eko in any way you like, but it is now always stored in the "canconical basis" (of course you can force your way)

scarlehoff:
Yes, they might be correct by chance... but if they are correct then we are happy I'd say, chance or not

src/eko/io/metadata.py
```diff
@@ -55,13 +57,21 @@ def load(cls, path: os.PathLike):
...
         path = pathlib.Path(path)
-        content = cls.from_dict(
-            yaml.safe_load(InternalPaths(path).metadata.read_text(encoding="utf-8"))
-        )
+        paths = InternalPaths(path)
+        # read raw file first to catch version
+        raw = yaml.safe_load(paths.metadata.read_text(encoding="utf-8"))
+        version = parse(raw["version"])
+        # patch if necessary
+        if version.major == 0 and version.minor == 13:
```
giacomomagni, line 69: I wonder if we should add a case also for the `0.0.0` ?

scarlehoff: In the end we decided that no

felixhekhorn:
sorry @evagroenendijk for the delay but I added some todo items to the PR description above on which you can work as a warm-up; are the items sufficiently clear? Please ping me if they are not ...

@giacomomagni should we pin the test cases to LHA LO FFNS unpolarized? and check the actual numbers are fine? should we check for failure (i.e. "non-squared" or x rotated)?

tests/eko/io/test_legacy.py
```diff
@@ -1,20 +1,84 @@
...
+            th_card = evolution_operator.theory_card
+            
+            print(th_card.raw)
+            print(op_card.raw)
+
+        path = pathlib.Path(TEST/name)
+        path.unlink(missing_ok=True)
+        eko.solve(th_card,op_card,path)
+        evolve_pdfs(
+            [pdf], th_card, op_card, install=True, name="Evolved_pdf", store_path=path
```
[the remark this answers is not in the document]
felixhekhorn: 1. see my comment above
2. `evolve_pdfs` is doing way too much things, which we don't need here in the first place - so drop this line
3. instead we want to use `apply_pdf` which was already imported - for an example on how to use it see https://eko.readthedocs.io/en/latest/overview/tutorials/pdf.html#Method-1:-Using-apply_pdf

tests/eko/io/test_legacy.py
```diff
@@ -1,20 +1,84 @@
...
+        pdf_test = []
+        for k in range(len(x_grid)):
+            pdf_test.append(evolved_pdf.xfxQ2(21, x_grid[k], Q2))
+
+        # Import the values of the LHA benchmark tables. This is not very nice yet, but should work
+        lha_path = pathlib.Path(__file__).parents[4] / 'eko/src/ekomark/benchmark/external/LHA.yaml'
+        with open(lha_path,'r') as file:
+            lha_benchmark = yaml.safe_load(file)
+
+        xpdf_benchmark = lha_benchmark["table2"]["part2"]["g"]
```
felixhekhorn, line 63: use the `ekomark` interface to do all this - it knows and you don't need to hack here

tests/eko/io/test_legacy.py
```diff
@@ -1,20 +1,84 @@
...
+        lha_path = pathlib.Path(__file__).parents[4] / 'eko/src/ekomark/benchmark/external/LHA.yaml'
+        with open(lha_path,'r') as file:
+            lha_benchmark = yaml.safe_load(file)
+
+        xpdf_benchmark = lha_benchmark["table2"]["part2"]["g"]        
+
+        pdf_benchmark = []      # gluon PDF at 10000 GeV^2 from the LHA benchmark tables
+        """
+        --> have to divide by x values to compare or not
+        Was the LHAPDF output xf(x,q2) or f(x,q2) again?
```
felixhekhorn, line 68: the LHAPDF function is `xfxQ2` - however EKO returns just `f` so you need to do something

tests/eko/io/test_legacy.py
```diff
@@ -1,20 +1,84 @@
...
+            pdf_benchmark.append(xpdf_benchmark[j]/x_grid[j])
+
+        #print(pdf_test, pdf_benchmark)
+        
+        # Test that the PDF values are the same 
+        for i in range(len(pdf_test)):
+            if pdf_test[i] != pdf_benchmark[i]:
+                print(name, False)
+            else:
+                print(name, True)
```
felixhekhorn, line 80: why not just use `np.testing`?

tests/eko/io/test_legacy.py
```diff
@@ -1,20 +1,84 @@
...
+       
+    return evolved_pdf
```
felixhekhorn, line 82: no need to return anything here, this is expected to run under pytest which throws everything away

tests/eko/io/test_legacy.py
```diff
@@ -1,20 +1,84 @@
...
+
+test_read_legacy()
```
felixhekhorn, line 84: fine for the moment, but keep in mind that you can do the same thing using the pytest interface

tests/eko/io/test_legacy.py
```diff
@@ -1,20 +1,84 @@
 import pathlib
-
-TEST = pathlib.Path(__file__).parents[2] / "data"
+from math import nan
+TEST = pathlib.Path(__file__).parents[2] / "data"   # directory of the EKO object
```
felixhekhorn, line 3: How about
```suggestion
# path of the EKO objects
TEST_DATA_DIR = pathlib.Path(__file__).parents[2] / "data"
```
also `ruff` will complain that you have to do all the imports first

tests/eko/io/test_legacy.py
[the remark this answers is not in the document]
evagroenendijk: Thank you! I'm doing it with apply_pdf now and seems to work fine. One question: do the pdfs have to be spot-on the same as in the LHA benchmark or is agreement up to some precision also alright?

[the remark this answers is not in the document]
felixhekhorn: well, it has to be as good as is master currently, i.e.
<details>
<summary> $ poetry poe lha -m "lo and ffns" </summary>

```
 ─── 
  S  
 ─── 
               x       Q2         eko     eko_error         LHA  percent_error
0   1.000000e-07  10000.0  196.013420  2.070574e-04  196.021092      -0.003914
1   1.000000e-06  10000.0   98.182659  1.004108e-04   98.207532      -0.025327
2   1.000000e-05  10000.0   46.918471  6.991565e-05   46.923979      -0.011737
3   1.000000e-04  10000.0   21.203157  2.303485e-05   21.208365      -0.024556
4   1.000000e-03  10000.0    9.010651  8.842617e-06    9.012788      -0.023707
5   1.000000e-02  10000.0    3.646971  1.776247e-06    3.647790      -0.022450
6   1.000000e-01  10000.0    1.440969  2.043208e-06    1.440964       0.000377
7   3.000000e-01  10000.0    0.575928  1.937128e-08    0.575929      -0.000203
8   5.000000e-01  10000.0    0.173979  5.302622e-10    0.173983      -0.001839
9   7.000000e-01  10000.0    0.026810  2.180479e-08    0.026811      -0.000617
10  9.000000e-01  10000.0    0.000454  3.694197e-09    0.000457      -0.597290

 ─── 
  g  
 ─── 
               x       Q2  ...          LHA  percent_error
0   1.000000e-07  10000.0  ...  1316.200000      -0.000565
1   1.000000e-06  10000.0  ...   600.080000      -0.028675
2   1.000000e-05  10000.0  ...   254.190000      -0.016463
3   1.000000e-04  10000.0  ...    97.371000      -0.029621
4   1.000000e-03  10000.0  ...    32.078000      -0.030995
5   1.000000e-02  10000.0  ...     8.054600      -0.037837
6   1.000000e-01  10000.0  ...     0.887660      -0.001243
7   3.000000e-01  10000.0  ...     0.082676      -0.004348
8   5.000000e-01  10000.0  ...     0.007924      -0.004414
9   7.000000e-01  10000.0  ...     0.000373      -0.068088
10  9.000000e-01  10000.0  ...     0.000001       8.994289

[11 rows x 6 columns]

 ─── 
  V  
 ─── 
               x       Q2       eko     eko_error       LHA  percent_error
0   1.000000e-07  10000.0  0.000092  1.301617e-10  0.000092       0.000420
1   1.000000e-06  10000.0  0.000532  1.064871e-09  0.000532       0.000623
2   1.000000e-05  10000.0  0.002979  2.901840e-09  0.002979       0.002073
3   1.000000e-04  10000.0  0.015964  1.376862e-08  0.015965      -0.001269
4   1.000000e-03  10000.0  0.079689  6.163553e-08  0.079688       0.000638
5   1.000000e-02  10000.0  0.343877  1.564116e-07  0.343880      -0.000781
6   1.000000e-01  10000.0  0.856808  1.662353e-08  0.856800       0.000946
7   3.000000e-01  10000.0  0.521109 -4.859107e-10  0.521110      -0.000148
8   5.000000e-01  10000.0  0.170121  1.510332e-10  0.170124      -0.001738
9   7.000000e-01  10000.0  0.026717 -2.732362e-09  0.026717       0.000657
10  9.000000e-01  10000.0  0.000454 -1.011684e-09  0.000457      -0.606629

 ──── 
  V3  
 ──── 
               x       Q2       eko     eko_error       LHA  percent_error
0   1.000000e-07  10000.0  0.000023  3.255914e-11  0.000023       0.000771
1   1.000000e-06  10000.0  0.000136  2.663909e-10  0.000136      -0.002105
2   1.000000e-05  10000.0  0.000766  7.272131e-10  0.000766       0.003961
3   1.000000e-04  10000.0  0.004149  3.468569e-09  0.004149      -0.003267
4   1.000000e-03  10000.0  0.021096  1.573728e-08  0.021096       0.002286
5   1.000000e-02  10000.0  0.095220  4.562153e-08  0.095220      -0.000331
6   1.000000e-01  10000.0  0.288539  6.692407e-09  0.288540      -0.000499
7   3.000000e-01  10000.0  0.237386 -2.448215e-10  0.237390      -0.001621
8   5.000000e-01  10000.0  0.099393  1.083280e-10  0.099396      -0.003019
9   7.000000e-01  10000.0  0.019526 -2.042607e-09  0.019529      -0.011735
10  9.000000e-01  10000.0  0.000409 -9.033632e-10  0.000412      -0.814945

 ──── 
  T3  
 ──── 
               x       Q2       eko     eko_error       LHA  percent_error
0   1.000000e-07  10000.0  0.000022  3.077890e-11  0.000022       0.000794
1   1.000000e-06  10000.0  0.000126  2.490182e-10  0.000126      -0.002292
2   1.000000e-05  10000.0  0.000703  6.645188e-10  0.000703       0.004368
3   1.000000e-04  10000.0  0.003768  3.125672e-09  0.003768      -0.003604
4   1.000000e-03  10000.0  0.018973  1.394690e-08  0.018972       0.002227
5   1.000000e-02  10000.0  0.085273  4.117576e-08  0.085274      -0.000443
6   1.000000e-01  10000.0  0.267598  6.410300e-09  0.267600      -0.000744
7   3.000000e-01  10000.0  0.230780 -2.410765e-10  0.230784      -0.001672
8   5.000000e-01  10000.0  0.098537  1.081309e-10  0.098540      -0.003150
9   7.000000e-01  10000.0  0.019495 -2.040230e-09  0.019497      -0.011705
10  9.000000e-01  10000.0  0.000409 -9.034513e-10  0.000412      -0.818107

 ──── 
  T8  
 ──── 
               x       Q2       eko     eko_error       LHA  percent_error
0   1.000000e-07  10000.0  2.182090  3.413575e-06  2.181092       0.045772
1   1.000000e-06  10000.0  1.733373  4.860242e-06  1.733532      -0.009178
2   1.000000e-05  10000.0  1.379013  4.156562e-07  1.378979       0.002490
3   1.000000e-04  10000.0  1.105966  4.637348e-07  1.106165      -0.017972
4   1.000000e-03  10000.0  0.928381  2.681725e-07  0.928488      -0.011559
5   1.000000e-02  10000.0  0.930818  2.943862e-07  0.930900      -0.008810
6   1.000000e-01  10000.0  1.031162  1.766631e-08  1.031160       0.000198
7   3.000000e-01  10000.0  0.540243 -4.925727e-10  0.540243       0.000008
8   5.000000e-01  10000.0  0.171521  1.512387e-10  0.171524      -0.001742
9   7.000000e-01  10000.0  0.026751 -2.734662e-09  0.026751       0.000160
10  9.000000e-01  10000.0  0.000454 -1.011559e-09  0.000457      -0.603362

 ───── 
  T15  
 ───── 
               x       Q2       eko     eko_error       LHA  percent_error
0   1.000000e-07  10000.0  4.364089  6.827019e-06  4.365092      -0.022986
1   1.000000e-06  10000.0  3.466214  9.719419e-06  3.467532      -0.038016
2   1.000000e-05  10000.0  2.755048  8.284105e-07  2.755979      -0.033796
3   1.000000e-04  10000.0  2.195967  9.137010e-07  2.196365      -0.018093
4   1.000000e-03  10000.0  1.777073  4.747096e-07  1.777188      -0.006479
5   1.000000e-02  10000.0  1.517759  4.323608e-07  1.517910      -0.009970
6   1.000000e-01  10000.0  1.205516  1.870909e-08  1.205508       0.000661
7   3.000000e-01  10000.0  0.559377 -4.992348e-10  0.559377      -0.000080
8   5.000000e-01  10000.0  0.172920  1.514443e-10  0.172923      -0.001787
9   7.000000e-01  10000.0  0.026784 -2.736963e-09  0.026784      -0.000332
10  9.000000e-01  10000.0  0.000454 -1.011434e-09  0.000457      -0.600095
```

</details>

so I'd say we can start with some global settings, e.g. rel. error < 0.05% and abs. error < 5e-7 and then let's see if we need to make it slightly looser or we can make it sharper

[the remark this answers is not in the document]
evagroenendijk: So at the moment e.g. for v0.14 all relative errors are <0.09, this is too big then or is it okay?

felixhekhorn:
sorry for touching your file @evagroenendijk in 023643b , but I wanted to see if my changes for the workflow is sufficient (which seems to be the case). In the file nothing changed but cosmetics.

evagroenendijk:
@felixhekhorn no problem at all to touch my script, thanks for testing it! I am back today and will resume.

tests/eko/io/test_legacy.py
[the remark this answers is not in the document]
evagroenendijk: I find a max absolute difference of 1.7E4 at the low-x for g(x) at 10000GeV^2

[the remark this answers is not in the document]
felixhekhorn: okay, this would be much more ... then I'd rather buy the 1/1e3 relative error (and just to be sure: above you meant 0.09% and not 9%, right?)

tests/eko/io/test_legacy.py
```diff
@@ -43,21 +55,28 @@ def test_read_legacy():
 
         pdf_benchmark = []  # gluon PDF at 10000 GeV^2 from the LHA benchmark tables
         """
-        --> have to divide by x values to compare or not
-        Was the LHAPDF output xf(x,q2) or f(x,q2) again?
+        --> have to divide by x values to compare
```
felixhekhorn, line 58: don't (ab)use multiline docstrings for code comments a simple
```py
# remember to divide by x to match eko normalization
```
is sufficient (and a bit more descriptive)

src/eko/io/struct.py
```diff
@@ -408,7 +422,7 @@ def dump(self, archive: Optional[Path] = None):
         Parameters
         ----------
         archive: Path or None
-            path to archive, in general you should keep the default, that will
+            path to archive, in ghttps://data.nnpdf.science/eko/test-data/general you should keep the default, that will
```
felixhekhorn, line 425:
```suggestion
            path to archive, in general you should keep the default, that will
```

src/eko/io/v3.py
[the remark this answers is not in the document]
evagroenendijk: Okay yes makes sense!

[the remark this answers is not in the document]
felixhekhorn: correct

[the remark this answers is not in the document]
evagroenendijk: By the way, I think that v1.py an v2.py can be exactly the same, shall I make them one file or keep them separate?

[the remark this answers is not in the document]
felixhekhorn: Mmm not sure ... maybe I would keep them separated ... v2 could of course reuse stuff from v1 for example ... let's ask @giacomomagni for a second opinion

tests/eko/io/test_legacy.py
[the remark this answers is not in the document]
felixhekhorn: https://github.com/NNPDF/eko/blob/8fa37e9994520f7339f0ae053c6e97eb31e0a45b/tests/eko/io/test_legacy.py#L71
- this is way too much I would say: we can not not be off by almost 10% - the logs in Github are no longer available, but if you managed to get v0.13.5 running just run the benchmarks there and check how good we are.
- setting atol is non-sense here: such an error will consume everything. It only makes sense to use it for small absolute values, which are numerically difficult, e.g. `atol=1e-6` (and if we don't need it in the first place even better)

tests/data/assets.sh
```diff
@@ -0,0 +1,36 @@
...
+# Server path
+URL="https://data.nnpdf.science/eko/test-data/"
+# local paths
+PYDATADIR="tests/data/"
+RUSTDATADIR="crates/dekoder/tests/data/"
+
+# download from the server
+curl -s -C - -o "./${PYDATADIR}v0.13.5.tar" "${URL}v0.13.5.tar"
+curl -s -C - -o "./${PYDATADIR}v0.14.tar" "${URL}ekov014.tar"
+curl -s -C - -o "./${PYDATADIR}v0.15.tar" "${URL}v0.0.tar"
```
felixhekhorn, line 32: please rename `0.15` to `0.0` everywhere (all files and all codes)

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,78 @@
...
+            with evolution_operator.operator((10000.0, 4)) as op:
+                if op.operator.shape != (14, 60, 14, 60):
+                    print("Operator does not have the correct dimensions")
+
+            # Use the operator on the pdf
+            evolved_pdfs, _integration_errors = apply_pdf(
+                evolution_operator, pdf, x_grid
+            )
+
+        print("Evolution grid", evolution_operator.evolgrid)
```
felixhekhorn, line 49: no need to print anything - rely on `pytest`

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,78 @@
...
+        pdf_test = evolved_pdfs[(Q2,4)][21]  # evolved gluon PDF at 10000 GeV^2
```
felixhekhorn, line 50: of course we need all flavors

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,88 @@
...
+        lha_path = (
+            pathlib.Path(__file__).parents[4]
+            / "eko/src/ekomark/benchmark/external/LHA.yaml"
+        )
+        with open(lha_path, "r") as file:
+            lha_benchmark = yaml.safe_load(file)
+
+        pdf_benchmark = {}
+        for key in ["u_v", "d_v", "L_p", "L_m", "s_p", "c_p", "g"]:
+            globals()[f"pdf_benchmark_{key}"] = []
```
felixhekhorn, line 73: never ever use `globals()` :scream: (PS: of course every rule has an exception, but this is surely not the case here)

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,88 @@
...
+            pdf_test[flav] = evolved_pdfs[(Q2,4)][flav]
+
+        pdfs = {}
+        pdfs["u_v"] = pdf_test[2] - pdf_test[-2]
+        pdfs["d_v"] = pdf_test[1] - pdf_test[-1]
+        pdfs["L_p"] = 2*(pdf_test[-2] + pdf_test[-1])
+        pdfs["L_m"] = pdf_test[-1] - pdf_test[-2]
+        pdfs["s_p"] = pdf_test[3] + pdf_test[-3]
+        pdfs["c_p"] = pdf_test[4] + pdf_test[-4]
+        pdfs["g"] = pdf_test[21]
```
felixhekhorn, line 61: again: use the ekomark tools - in this case e.g. https://github.com/NNPDF/eko/blob/af243c46ae505fa2e113707e7b689958bc88c177/src/ekomark/benchmark/external/LHA_utils.py#L35

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,88 @@
...
 
-    mu2 = 1.959
-    op = 29348.2342
-    err = 54225.24
-    op4s = legacy.op5to4([mu2], {legacy.OPERATOR: [op], legacy.ERROR: [err]})
-    assert mu2 in op4s
-    assert op4s[mu2].operator == op
-    assert op4s[mu2].error == err
+x_grid = [1e-7, 1e-6, 1e-5, 1e-4, 1e-3, 1e-2, 0.1, 0.3, 0.5, 0.7, 0.9]
+Q2 = 10000
```
felixhekhorn, line 21:
```suggestion
EP = (10000., 4)
```
and use that consistently. (recall that conventionally EvolutionPoint -> ep)

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,88 @@
...
+
+
+def test_read_legacy_cards():
+    for name in ["v0.13.5.tar", "v0.14.tar", "v0.0.tar"]:
+        with eko.EKO.read(TEST_DATA_DIR / name) as evolution_operator:
+            # import pdb; pdb.set_trace()
```
felixhekhorn, line 27:
```suggestion
```
Remove all debug leftovers

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,88 @@
...
-    op4s = legacy.op5to4([mu2], {legacy.OPERATOR: [op], legacy.ERROR: [err]})
-    assert mu2 in op4s
-    assert op4s[mu2].operator == op
-    assert op4s[mu2].error == err
+x_grid = [1e-7, 1e-6, 1e-5, 1e-4, 1e-3, 1e-2, 0.1, 0.3, 0.5, 0.7, 0.9]
+Q2 = 10000
+
+
+def test_read_legacy_cards():
+    for name in ["v0.13.5.tar", "v0.14.tar", "v0.0.tar"]:
```
felixhekhorn, line 25: Define the names as (module) constant and use that consistently

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,88 @@
...
+            globals()[f"pdf_benchmark_{key}"] = []
+            globals()[f"xpdf_benchmark_{key}"] = lha_benchmark["table2"]["part2"][key]
+            for i in range(len(x_grid)):
+                globals()[f"pdf_benchmark_{key}"].append(globals()[f"xpdf_benchmark_{key}"][i] / x_grid[i])
+            #globals()[f"pdf_benchmark_{key}"] = globals()[f"pdf_benchmark_{key}"][:-1]
+            pdf_benchmark[key] = globals()[f"pdf_benchmark_{key}"]
+            for j in range(len(pdfs[key])-5):
+                np.testing.assert_allclose(pdfs[key][j], pdf_benchmark[key][j], rtol=0.0005) 
+            for k in [6, 7, 8, 9, 10]:
+                np.testing.assert_allclose(pdfs[key][k], pdf_benchmark[key][k], atol=1e-3)
```
felixhekhorn, line 82: Interesting solution, but
- prefer indexing over iteration, e.g. you can do `pdf[keys][:5]` - this way you trade a python loop (slow) for a numpy operation (fast)
- make it more obvious that you cover all elements: at the moment we assume hardcoded that there are exactly 10 elements (and that you correctly typed all numbers between 6 and 10)
- can't we join the two statements? e.g. with `rtol=1e-3, atol=1e-4` or something similar (make it as strong as possible) - this is what I meant in an earlier comment

tests/eko/io/test_legacy.py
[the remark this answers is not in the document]
evagroenendijk: I have now atol=5e-6 and rtol=5e-4, is this okay?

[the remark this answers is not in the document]
felixhekhorn: yeah, I'm willing to buy that

tests/eko/io/test_legacy.py
felixhekhorn: mmm before you had a duplicated line `for name in ["v0.13.5.tar", "v0.14.tar", "v0.0.tar"]` which we could have changed to `for name in NAMES` - think of what will happen when we add a new version. Now you unified the cards check and the PDF check again and the problem has disappeared - but I'm not sure this was a good revert. I would still break the test into two functions: one very trivial for cards and one with a few lines more for PDF. If you think yours is better - convince me :upside_down_face:

evagroenendijk: Ah yes I understand. I thought it would look nicer to have it in one since for both you need to specify the names & define this evolution operator anyway. And I guess once you are testing, you would want to test everything anyway right? But I don't have a strong preference so I can also change it again and define NAMES in the beginning!

felixhekhorn: fine - let's keep as is

tests/eko/io/test_legacy.py
felixhekhorn: note the `.` - this makes a difference!

tests/eko/io/test_legacy.py
felixhekhorn: use ekomark through out! i.e. also get directly the table from there

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,66 @@
...
+            pathlib.Path(__file__).parents[4]
+            / "eko/src/ekomark/benchmark/external/LHA.yaml"
+        )
+        with open(lha_path, "r") as file:
+            lha_benchmark = yaml.safe_load(file)
+
+        # Make the test
+        pdf_test = {}
+        pdf_bench = {}
+        for flav in [-6,-5,-4,-3,-2,-1,21,1,2,3,4,5,6]:
```
felixhekhorn, line 59:
```py
import eko.basis_rotation as br
# ...
for flav in br.flavor_basis_pids:
```

felixhekhorn: actually: can't we drop the whole loop and just use the correct numpy cast?

tests/eko/io/test_legacy.py
evagroenendijk: So using LHA_utils.compute_LHA_data() you mean?

tests/eko/io/test_legacy.py
```diff
@@ -1,19 +1,66 @@
...
+            )  # Check that theory card is read as theory card
+            assert isinstance(
+                op, eko.io.runcards.OperatorCard
+            )  # Check that operator card is read as operator card
+
+            
+            # Check if the operator has the correct dimensions
+            with evolution_operator.operator(EP) as op:
+                if op.operator.shape != (14, 60, 14, 60):
+                    print("Operator does not have the correct dimensions")
```
felixhekhorn, line 41: use `assert`

tests/eko/io/test_legacy.py
felixhekhorn: yes - you can fake the arguments
```py
th = {"FNS": "FFNS", "PTO":0, } # etc
```

evagroenendijk: Ah okay, I didn't do it before because I thought it will not be much shorter if I define the theory and operator card manually there. But is it then the convention to use ekomark because otherwise you 'hack'?

felixhekhorn: ekomark is the interface - let's not interfere with its internals. And here in the test it is sufficient to just fake the card (with just the necessary information needed)

tests/eko/io/test_legacy.py
```diff
@@ -14,51 +14,48 @@
...
                 evolution_operator, pdf, x_grid
             )
 
         # Import the values of the LHA benchmark tables
-        lha_path = (
-            pathlib.Path(__file__).parents[4]
-            / "eko/src/ekomark/benchmark/external/LHA.yaml"
+        lha_benchmark = LHA_utils.compute_LHA_data(
+            {"FNS": "FFNS", "PTO": 0, "XIF": 1},
+            {"polarized": 0, "mugrid": 100}
```
felixhekhorn, line 50:
```suggestion
            {"polarized": False, "mugrid": 100.}
```
even if we fake, let's stick to the correct types

tests/eko/io/test_legacy.py
felixhekhorn: did you check that we can drop the loop?

evagroenendijk: Oh sorry! I tried removing the loop and just comparing the totals but ran into a problem because the pdf_bench has to be divided by the xgrid for every flavour so then I still needed a loop

evagroenendijk (approved):

PR #418, opened 2024-10-21
Add some pol tests

giacomomagni:
I think we can add this test on the polarized anomalous dimension.
(I messed up with the history because I started branching from the wrong branch)

### DIFF
--- a/benchmarks/eko/benchmark_inverse_matching.py
+++ b/benchmarks/eko/benchmark_inverse_matching.py
@@ -90,10 +90,14 @@ def benchmark_inverse_matching():
     with pytest.raises(AssertionError):
         np.testing.assert_allclose(op1_nf3.operator, op2_nf3.operator)
 
-    pdf1, _ = apply.apply_pdf(eko_output1, toy.mkPDF("ToyLH", 0))
-    pdf2, _ = apply.apply_pdf(eko_output2, toy.mkPDF("ToyLH", 0))
+    pdf1 = apply.apply_pdf(eko_output1, toy.mkPDF("ToyLH", 0))
+    pdf2 = apply.apply_pdf(eko_output2, toy.mkPDF("ToyLH", 0))
 
     # test that different PTO matching is applied correctly
-    np.testing.assert_allclose(pdf1[(MC**2, 4)][C_PID], pdf2[(MC**2, 4)][C_PID])
+    np.testing.assert_allclose(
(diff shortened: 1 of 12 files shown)


giacomomagni:
how do I access the beta function in the rust crates?

felixhekhorn:
> how do I access the beta function in the rust crates?

eeehm, good point - it is currently not available

mmm let me think a moment ... :thinking: 

1) the easy solution is just to ignore the Rust again for the moment in this PR (and delay the problem)
2) if we want to do the proper thing, we need to find a good place on the Rust side; I'm thinking about adding yet another crate, random name suggestion: `ekoupling`, where we provide quasi the equivalent to the `eko.couplings` module: specifically, the beta functions and their solution. I'm worried about dependency management and how to disentangle stuff. Do you have an opinion? @tgiani may I ask for your opinion as well?

giacomomagni:
> 
> 1. the easy solution is just to ignore the Rust again for the moment in this PR (and delay the problem)
> 2. if we want to do the proper thing, we need to find a good place on the Rust side; I'm thinking about adding yet another crate, random name suggestion: `ekoupling`, where we provide quasi the equivalent to the `eko.couplings` module: specifically, the beta functions and their solution. I'm worried about dependency management and how to disentangle stuff. Do you have an opinion? @tgiani may I ask for your opinion as well?

So regarding this PR, I'd postpone the problem, it's just an additional test, I can add the one unrelated to be beta function. 
For the long terms strategy I think we should provide the `ekoupling` module, but maybe you can add this to the `oxidation` road map or another issue.

felixhekhorn (approved):
as already said: maybe we can clean the commit history a bit (or we just keep it as is)

felixhekhorn:
ehm, now I wonder if we should hard code beta0, i.e. sacrifice code duplication for an additional test ...

tgiani:
> > how do I access the beta function in the rust crates?
> 
> eeehm, good point - it is currently not available
> 
> mmm let me think a moment ... 🤔
> 
> 1. the easy solution is just to ignore the Rust again for the moment in this PR (and delay the problem)
> 2. if we want to do the proper thing, we need to find a good place on the Rust side; I'm thinking about adding yet another crate, random name suggestion: `ekoupling`, where we provide quasi the equivalent to the `eko.couplings` module: specifically, the beta functions and their solution. I'm worried about dependency management and how to disentangle stuff. Do you have an opinion? @tgiani may I ask for your opinion as well?

I would agree with what Giacomo said, so delay the problem for the time being and put an `ekoupling` implementation on the to do list

PR #421, opened 2024-10-22
Rework ekobox.apply

giacomomagni:
Continue #415.


felixhekhorn:
I think that's it for `apply`, right? i.e. is there something else to be fixed? if not, we can just merge this and be happy ... do we know why our LO (!) numbers changed?

Issue #440, opened 2025-02-04
Add regression tests


felixhekhorn:
- this would be an alternative to #397 and #411 
- the strategy will be: search in the (dedicated) DB for given theory+op+PDF and return `"eko"` column
- data can be generated with `external = "void"` - which will be available with #416

PR #443, opened 2025-02-12
Rustify N3LO Ad

giacomomagni:


crates/ekore/src/harmonics/log_functions.rs
```diff
@@ -0,0 +1,280 @@
...
+    #[test]
+    fn test_lm1pm2() {
+        
+        fn mellin_lm1pm1(x: f64, k: usize, n: f64) -> f64 {
+            pow(pow(x, n - 1.0) * (1.0 - x) * (1.0 - x).ln(), k)
+        }
+
+        const Ns: [f64; 5] = [1.0, 1.5, 2.0, 2.34, 56.789];
+        for N in Ns.iter().enumerate() {
+            let n = Complex::new(*N.1, 0.);
```
felixhekhorn, line 217: you may find the `cmplx` macro useful, which is defined in the utils module

felixhekhorn:
Also here is another suggestion for the harmonics module: all functions called "`S` + positive integer" take as single argument $N$, all others take as single argument the cache - i.e. no messing with passing the correct arguments. (Since we are now in a compiled language I'm no longer worried of nesting things) we could even shift the "raw" S-es into the polygamma module ...

what do you think @giacomomagni @tgiani ?

crates/ekore/src/harmonics/log_functions.rs
```diff
@@ -0,0 +1,280 @@
...
+                    + 4. * pow(1. + n, 3) * pow(2. + n, 3) * S3)
+            + 6. * pow(1. + n, 4) * pow(2. + n, 4) * S4);
+}
+
+#[cfg(test)]
+mod tests {
+
+    use crate::harmonics as h;
+    use crate::{assert_approx_eq_cmplx};
+    use quad::integrate;
```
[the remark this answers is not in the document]
giacomomagni: Originally I wanted to do the integration, but so far this doesn't work :(

giacomomagni:
> Also here is another suggestion for the harmonics module: all functions called "`S` + positive integer" take as single argument N , all others take as single argument the cache - i.e. no messing with passing the correct arguments. (Since we are now in a compiled language I'm no longer worried of nesting things) we could even shift the "raw" S-es into the polygamma module ...
> 
> what do you think @giacomomagni @tgiani ?
 
Sure this require a bit of restructure though...

felixhekhorn:
> Sure this require a bit of restructure though...

I made an attempt in 462f62c - so, do we want to do this? I can take care of the copy&paste if you wish ;-)

I also made some adjustments to make pre-commit happy

crates/ekore/src/harmonics/log_functions.rs
```diff
@@ -88,22 +87,22 @@ pub fn lm15m1(
...
 /// Mellin transform of :math:`\ln(1-x)`.
 pub fn lm11(n: Complex<f64>, S1: Complex<f64>) -> Complex<f64> {
-    return -S1 / n;
+    -S1 / n
 }
 
 /// Mellin transform of :math:`\ln^2(1-x)`.
-pub fn lm1(n: Complex<f64>, S1: Complex<f64>, S2: Complex<f64>) -> Complex<f64> {
-    return (pow(S1, 2) + S2) / n;
+pub fn lm12(n: Complex<f64>, S1: Complex<f64>, S2: Complex<f64>) -> Complex<f64> {
```
felixhekhorn, line 99: I also fixed a typo here

felixhekhorn:
I'm also starting to think, that the log expressions are so complicated (look at `lm14m2`), that we should cache them - what do you think?

crates/ekore/src/harmonics/log_functions.rs
giacomomagni: oh thanks!!

giacomomagni:
> I'm also starting to think, that the log expressions are so complicated (look at `lm14m2`), that we should cache them - what do you think?

Some of them might not be needed, because the Moch et al. expressions are simpler... 
So I wanted to finish the conversion and then polish the harmonic file.

felixhekhorn:
> I made an attempt in [462f62c](https://github.com/NNPDF/eko/commit/462f62c475d2ebe74cc86423722f1829f2a41def) - so, do we want to do this?

I forgot to mention: the prize that we had to pay for this was
https://github.com/NNPDF/eko/blob/462f62c475d2ebe74cc86423722f1829f2a41def/crates/ekore/src/harmonics/w3.rs#L36-L37
i.e. we had to create a new cache, which is potentially expensive ... there are in principle two strategies I can imagine to avoid that (if we want to avoid it):
1) math: i.e. maybe we can express $g_3(N+1)-g_3(N)$, like [I was playing in the QED PR](https://github.com/NNPDF/eko/pull/416#discussion_r1876119975) - but looking at the notes still laying on my desk, I'm not sure we can simplify this
2) more interface: we add a `Cache::set` method, which allows us to fill an entry into the cache. Then I know $g_3(N+1)$ will need $S_1(N+1)$ and that I can derive from $S_1(N)$ (instead of recalculating)

crates/ekore/src/harmonics/log_functions.rs
```diff
@@ -126,24 +125,28 @@ pub fn lm15(
...
     S5: Complex<f64>,
 ) -> Complex<f64> {
-    -((((pow(S1, 5) + ((10. * pow(S1, 3)) * S2)) + ((20. * pow(S1, 2)) * S3))
-        + ((15. * S1) * (pow(S2, 2) + (2. * S4))))
-        + (4. * (((5. * S2) * S3) + (6. * S5))))
+    -(S1.powu(5)
+        + 10. * S1.powu(3) * S2
+        + 20. * S1.powu(2) * S3
+        + 15. * S1 * S2.powu(2)
+        + 2. * S4
```
felixhekhorn, line 132: is there a typo here by chance, i.e. `2. * S4 * S1`? because I guess all weights should add up to 5 and the one possibility you're missing is $S_4 \cdot S_1$

giacomomagni: yes, good catch! Fixed in 353bba6f

crates/ekore/Cargo.toml
```diff
@@ -19,6 +19,7 @@ rustdoc-args = ["--html-in-header", "doc-header.html"]
 
 [dependencies]
 num = "0.4.1"
+GSL = "7.0"
```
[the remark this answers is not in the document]
giacomomagni: right, done 491d6890

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs
```diff
@@ -0,0 +1,58 @@
...
+pub mod gnsm;
+pub mod gnsp;
+pub mod gnsv;
+pub mod gps;
+pub mod gqg;
+
+use crate::harmonics::cache::Cache;
+use num::complex::Complex;
+
+// Compute the singlet anomalous dimension matrix.
```
felixhekhorn, line 14: 1. please take a look to `poe rdocs` + `poe rdocs-view` to see if things are explained well in the documentation
2. doing so, you will notice that to annotate functions in rust you need `///` as `//` is just for comments
3. since we are doing an iteration any ways, let's improve the docs: e.g. I think it would be good to spell explicitly the order of variations here `gg,gq,...`
4. for adding new refences to rust: add them to the (relevant) bib file and run `poe rbib`

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs
```diff
@@ -0,0 +1,58 @@
...
+    use crate::harmonics::cache::Cache;
+    use crate::{assert_approx_eq_cmplx, cmplx};
+    use num::complex::Complex;
+
+    #[test]
+    fn test_momentum_conservation() {
+        let NF = 5;
+        let mut c = Cache::new(cmplx!(2., 0.));
+        let quark_refs: [f64; 3] = [0.053441, 0.225674, -0.118792];
+        let gluon_refs: [f64; 3] = [-0.0300842, 0.283004, -0.343172];
```
felixhekhorn, line 39: 1. can you please add a reference? (python is acceptable)
2. shouldn't they be 0?

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
```diff
@@ -0,0 +1,201 @@
...
+pub fn gamma_gg(c: &mut Cache, nf: u8, variation: u8) -> Complex<f64> {
+    let n = c.n();
+    let S1 = c.get(K::S1);
+    let S2 = c.get(K::S2);
+    let S3 = c.get(K::S3);
+    let S4 = c.get(K::S4);
+    let nf2 = nf.pow(2) as f64;
+    let nf3 = nf.pow(3) as f64;
+
+    //The known large-x coefficients [except delta(1-x)]
```
felixhekhorn, line 21: you are not running pre-commit (yet :upside_down_face: ) and so I'm not sure you are running `cargo fmt`, thus I'm not sure if it would change this to `// The` ... if it is not too much effort, I'd rather prefer the latter (throughout of course)

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
felixhekhorn, line 1: I'm not sure we want to expose publicly this file - maybe we want instead just re-export the `gamma_gg` (which hopefully is possible), what do you think? This way we would have a unique structure across all PTO and can still split the lengthy files. If so we should harmonize the doc string of `gamma_gg`, with the others.

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
```diff
@@ -0,0 +1,201 @@
...
+    use crate::harmonics::cache::Cache;
+    use crate::{assert_approx_eq_cmplx, cmplx};
+    use num::complex::Complex;
+
+    #[test]
+    fn test_reference_moments() {
+        fn gg3_moment(N: usize, nf: f64) -> f64 {
+            let nf2 = nf * nf;
+            let nf3 = nf2 * nf;
+            let mom_list = [
```
felixhekhorn, line 172: please add reference: in this case I guess it is Eq. xy of [blub]

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
```diff
@@ -0,0 +1,201 @@
...
+};
+
+// The routine is taken from [\[Falcioni:2024qpd\]][crate::bib:Falcioni:2024qpd].
+pub fn gamma_gg(c: &mut Cache, nf: u8, variation: u8) -> Complex<f64> {
+    let n = c.n();
+    let S1 = c.get(K::S1);
+    let S2 = c.get(K::S2);
+    let S3 = c.get(K::S3);
+    let S4 = c.get(K::S4);
+    let nf2 = nf.pow(2) as f64;
```
felixhekhorn, line 18: should we also cast `nf` here once and for all? (note that we want `u8` as input, and only cast internally)

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/gnsm.rs
```diff
@@ -0,0 +1,236 @@
...
+
+// The routine is taken from [\[Moch:2017uml\]][crate::bib:Moch:2017uml].
+
+// The nf^{0,1} leading large-nc contributions and the nf^2 part are
+// high-accuracy (0.1% or better) parametrizations of the exact
+// results. The nf^3 expression is exact up to numerical truncations.
+
+// The remaining nf^{0,1} terms are approximations based on the first
+// eight odd moments together with small-x and large-x constraints.
+// The two sets spanning the error estimate are called via  IMOD = 1
```
felixhekhorn, line 18: I like the literal copy of the comment by Moch et al., but we should replace `IMOD` -> `variation` (or change the name of the argument)

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/gqg.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    // The resulting part of the function
+    let P3QG01 = bfkl1 * 2. / (-1. + n).powu(3)
+        + x0L6cff * 720. / n.powu(7)
+        + x0L5cff * -120. / n.powu(6)
+        + x0L4cff * 24. / n.powu(5)
+        + x1L4cff * lm14(n, S1, S2, S3, S4)
+        + x1L5cff * lm15(n, S1, S2, S3, S4, S5)
+        + y1L4cff * lm14m1(n, S1, S2, S3, S4)
+        + y1L5cff * lm15m1(n, S1, S2, S3, S4, S5);
```
felixhekhorn, line 42: since `lm15m1` appears both in qg and gq, maybe we should cache it ... (if you want to do a proper job, we should first check if that makes a difference at all :see_no_evil:)

felixhekhorn (changes requested):

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
giacomomagni: I did run `cargo fmt` so I think this is correct.

felixhekhorn: I've done it myself in 54331631 - see also description there

crates/ekore/Cargo.toml
```diff
@@ -22,3 +22,4 @@ num = "0.4.1"
 
 [dev-dependencies]
 float-cmp = "0.9.0"
+GSL = "7.0"
```
felixhekhorn, line 25: I'm still not fully convinced this is what we want, but if so then
1. you need to update Cargo.lock - which by the way currently breaks pre-commit
2. change the workflow: you have to install `gsl` inside the action runner, as the rust packages [relies on the C package](https://github.com/NNPDF/eko/actions/runs/13388466311/job/37390482840#step:9:116)

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
felixhekhorn: step 1: c30aa6c2

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs
giacomomagni: I can't run `poe rbib`, do you know how to fix this? 

```
(eko_dev)  pip list | grep bibtexparser
bibtexparser                    1.4.3
(eko_dev) poe rbib                      
Poe => python crates/make_bib.py > crates/ekore/src/bib.rs
  File "/Volumes/Git_Workspace/physicstools/NN3PDF/eko/crates/make_bib.py", line 45, in <module>
    bib_database = bibtexparser.parse_string(bib_str)
AttributeError: module 'bibtexparser' has no attribute 'parse_string'
```

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs
giacomomagni: Added a reference to python, I guess that is the precision of the approximations ?
(I think to have checked the numbers also in Mathematica...)

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
giacomomagni: done in b0869d11

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/gnsm.rs
giacomomagni: done in b0869d11

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs
felixhekhorn: yes: a7760553 - sorry I was lazy :see_no_evil:

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
felixhekhorn: completed in 3df416ea

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs
giacomomagni: Thanks

felixhekhorn: > since we are doing an iteration any ways, let's improve the docs: e.g. I think it would be good to spell explicitly the order of variations here gg,gq,...

can you please take care of this?

giacomomagni: https://github.com/NNPDF/eko/blob/a72d78cd19782476d1766ab31765911b979b0eb8/crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs#L24-L26

felixhekhorn: sorry I'm blind :see_no_evil:

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
giacomomagni: done in f8eb85ce

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/gqg.rs
giacomomagni: indeed will it make a difference, all the harmonics are already coming from the cache?

crates/ekore/Cargo.toml
giacomomagni: This is now working, so I think we can keep it. But feel free to change it.

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/gqg.rs
felixhekhorn: yes, the harmonics are cached - maybe let's not do an optimization without checking it ... the numerical argument can be: loops and branching are expensive (as e.g. in `polygamma`), sums and product are fast (as in `lmblub`)

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggg.rs
felixhekhorn: Improved in e91e0bda :upside_down_face:

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs
felixhekhorn: they are non-zero also at $a_s^3$ so fine I guess?

felixhekhorn:
Any comment on the [cache recreation issue](https://github.com/NNPDF/eko/pull/443#issuecomment-2659410184)?

Else: is this ready? if yes, let's mark it, thus make the benchmarks run and the we can quickly merge

giacomomagni:
> Else: is this ready? if yes, let's mark it, thus make the benchmarks run and the we can quickly merge

Not yet! I need to add the variation to the kernel. I'll do it asap.

giacomomagni:
> Any comment on the [cache recreation issue](https://github.com/NNPDF/eko/pull/443#issuecomment-2659410184)?

Maybe all these optimisation can be addressed once the full ekore is in place ?
I guess that for the moment having a new cache in these function is fine as it will be mostly empty (filled only with g3 and related harmonics). But probably the math solution is also doable...

crates/eko/src/lib.rs
```diff
@@ -205,6 +238,7 @@ pub unsafe extern "C" fn rust_quad_ker(u: f64, rargs: *mut c_void) -> f64 {
         args.a_half_x,
         args.a_half_y,
         args.alphaem_running,
+        args.n3lo_ad_variation,
```
felixhekhorn, line 241: There is no need to pass the `n3lo_ad_variation` back to python - they are consumed here

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs
```diff
@@ -0,0 +1,71 @@
...
+pub use gnsp::gamma_nsp;
+pub use gnsv::gamma_nsv;
+pub use gps::gamma_ps;
+pub use gqg::gamma_qg;
+
+/// Compute the singlet anomalous dimension matrix.
+///
+/// `variation = (gg, gq, qg, qq)` is a list indicating which variation should
+/// be used. `variation = 1,2` is the upper/lower bound, while any other value
+/// returns the central (averaged) value.
```
felixhekhorn, line 26: Let's lift that from here to the top level, i.e. `ad::u::s::gamma_singlet_qcd` and add an explicit reference to that instead (something like "for the sorting of variation see blub")

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs
```diff
@@ -12,9 +12,16 @@ pub mod as1;
...
+pub mod as4;
 
 /// Compute the tower of the non-singlet anomalous dimensions.
-pub fn gamma_ns_qcd(order_qcd: usize, mode: u16, c: &mut Cache, nf: u8) -> Vec<Complex<f64>> {
+pub fn gamma_ns_qcd(
+    order_qcd: usize,
+    mode: u16,
+    c: &mut Cache,
+    nf: u8,
+    n3lo_variation: [u8; 3],
```
felixhekhorn, line 23: I'm torn between
1. having a uniform interface, i.e. passing everywhere [u8; 7] and
2. passing only the relevant information, i.e. [u8; 3]

look also to `gamma_ns_qed` ... let's also ask to @tgiani for a second opinion ... maybe at the top level (i.e. here) I slightly prefer 1.?
for 1. we could also define some constants á là `const IDX_VARIATION_GG: usize = 0;` etc.

in any case I suggest to have the documentation top-level: for singlet see my other comment and also here we need to have an explicit spell out (be it 1. or second)

felixhekhorn (changes requested):
- if only we had https://github.com/NNPDF/eko/issues/440 :see_no_evil: 
- do you have by chance a time (ratio) estimate?
- on second thought: should we also make `w1::S1` take only the cache as argument?

besides these small issues the PR is good to go

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4.rs
giacomomagni: done in 5882487e

crates/eko/src/lib.rs
giacomomagni: I should have fixed.

giacomomagni:
Just to have a quick feeling on the spped-up.
Running an eko with an `x-grid` of 30 points and 2 `Q2s`, in VFNS I get: 

* Python (numba already compiled): 
- `order: (0,0), matching (0,0):  71.05s user 2.86s system 103% cpu 1:11.70 total`
- `order: (4,0), matching (2,0):  147.30s user 4.52s system 100% cpu 2:30.35 total`

* Rust : 
- `order: (0,0), matching (0,0):  10.76s user 1.59s system 125% cpu 9.809 total`
- `order: (4,0), matching (2,0):  18.48s user 1.63s system 114% cpu 17.566 total`

So the speed up ratio is more or less the same at both orders.

giacomomagni:
> FFNS I assume? other it would be not fair

VFNS, recall now we're able to decouple the 2 ptos!

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs
felixhekhorn: some documentation added in 4d9dfa45 - let's postpone the decision on 1. vs. 2. to later

felixhekhorn (approved):

PR #444, opened 2025-02-18
[WIP] X space conversion of the N3LO matching 

giacomomagni:
Just to keep track, not sure if we really want to merge.

### DIFF
--- a/extras/ome_n3lo/convert_ome_xspace.py
+++ b/extras/ome_n3lo/convert_ome_xspace.py
@@ -1,0 +1,91 @@
+
+import numpy as np
+from click import progressbar
+from eko.mellin import Path
+from ekore.harmonics import cache as c
+from ekore.operator_matrix_elements.unpolarized.space_like import as3
+from scipy import integrate
+
+XGRID = np.geomspace(1e-6, 1, 500)  # 500
+"""X-grid."""
+
+LOG = 0
+"""Matching threshold displaced ?"""
+
+MAP_ENTRIES = {
+    "gg": (0, 0),
+    "gq": (0, 1),
+    "qg": (1, 0),
+    "qq": (1, 1),
+    "Hg": (2, 0),
+    "Hq": (2, 1),
+    "gH": (0, 2),
+    "HH": (2, 2),
+    "qq_ns": (0, 0),
+}
+
+
+def compute_ome(nf, n, is_singlet):
+    """Get the correct ome from eko."""
+    cache = c.reset()
+    if is_singlet:
+        return as3.A_singlet(n, cache, nf, L=0)
+    else:
+        return as3.A_ns(n, cache, nf, L=0)
+
+
+def compute_xspace_ome(entry, nf, x_grid=XGRID):
+    """Compute the x-space transition matrix element, returns A^3(x)."""
+    mellin_cut = 5e-2
+    is_singlet = "ns" not in entry
+
+    def integrand(u, x):
+        """Mellin inversion integrand."""
+        path = Path(u, np.log(x), is_singlet)
+        integrand = path.prefactor * x ** (-path.n) * path.jac
+        if integrand == 0.0:
+            return 0.0
+
+        ome_n = compute_ome(nf, path.n, is_singlet)
+        idx1, idx2 = MAP_ENTRIES[entry]
+        ome_n = ome_n[idx1, idx2]
+
+        # recombine everything
+        return np.real(ome_n * integrand)
+
+    ome_x = []
+    print(f"Computing operator matrix element {entry} @ pto: 3")
+    # loop on xgrid
+    with progressbar(x_grid) as bar:
+        for x in bar:
+            if x == 1:
+                ome_x.append(0)
+                continue
+            res = integrate.quad(
+                lambda u: integrand(u, x),
+                0.5,
+                1.0 - mellin_cut,
+                epsabs=1e-12,
+                epsrel=1e-6,
+                limit=200,
+                full_output=1,
+            )[0]
+            ome_x.append(res)
+
+    return np.array(ome_x)
+
+
+def save_files(entry, nf, ome_x, xgrid=XGRID):
+    """Write the space reuslt in a txt file."""
+    with open(f"x_space/A_{entry}_nf{nf}.txt", "w") as file:
+        for x, a in zip(xgrid, ome_x):
+            file.write(f"{x}\t{a}\n")
+
+
+if __name__ == "__main__":
+    nf = 3 # nf = 3,4,5.
+    # non diagonal temrms
+    for k in ["gq", "qg", "Hg", "Hq"]:
+        result = compute_xspace_ome(k, nf)
+        save_files(k, nf, result)
+    # ["ns", "gg", "qq",]:


extras/ome_n3lo/convert_ome_xspace.py
```diff
@@ -0,0 +1,91 @@
...
+                epsrel=1e-6,
+                limit=200,
+                full_output=1,
+            )[0]
+            ome_x.append(res)
+
+    return np.array(ome_x)
+
+
+def save_files(entry, nf, ome_x, xgrid=XGRID):
```
felixhekhorn, line 78: https://numpy.org/doc/stable/reference/generated/numpy.savetxt.html#numpy-savetxt ?

felixhekhorn (commented):
let's keep this - it's not much of a burden and it might be useful in the future (or for MSHT to check we're doing the right thing)

giacomomagni:
Hi @felixhekhorn, 
so I managed to extract the large-N limit $\mathcal{O}(1/N)$ of our $A_{qq}^{(3)}$ and $A_{gg}^{(3)}$ expressions,
i.e. the coefficients of $S_1(N)$ and the constant bit. 

The idea is to explicitly subtract them before doing the $N \to x$ conversion and give to the MSHT people also the two expansions (converted in x-space) so the they can add the terms as they like. 
Do you have any comment ? Robert has confirmed me that the `.txt` format I'm using is fine for them.

extras/ome_n3lo/large_n_limit.py
```diff
@@ -0,0 +1,38 @@
+"""This file contains the large-N limit of the diagonal Matrix elements.
+
+The expansions are obtained using the notebook Agg_Aqq_largex_expansion.nb.
+
+We note that:
+    * the limit og :math:`A_{qq}` is the same for non-singlet like and singlet-like expansions.
+    I.e. the local and singular part are the same
+    * the :math:`A_{qq,ps}` temr is vanishing in the large-x limit, i.e. it's only regular.
```
felixhekhorn, line 8:
```suggestion
    * the limit of :math:`A_{qq}` is the same for valence-like and singlet-like expansions,
       i.e. the local and singular parts are the same
    * the :math:`A_{qq,ps}` term is vanishing in the large-x limit, i.e. it's only regular.
```

extras/ome_n3lo/large_n_limit.py
```diff
@@ -0,0 +1,38 @@
...
+"""
```
felixhekhorn, line 9:
```suggestion

In the following we make use of a small trick, which is available due to complex continuation. Although we are interested in the large-N limit, we can obtain that by observing the small-N limit. It is :math:`S_1(0)=0` (by continuation), :math:`S_1(1)=1` and :math:`S_1(N\to\infty) = \ln(N) - \gamma_E`.

"""
```
It took me 10 min to realize that you can indeed do that - but I believe you messed up the local part, because the $S_1$ contributes to the local part:
https://github.com/NNPDF/eko/blob/ee8f53310c27689dd345dfedd9cdc5e04fae9bfb/src/ekore/harmonics/w1.py#L33
and `Series[PolyGamma[0, n + 1] + EulerGamma, {n, \[Infinity], 0}]` yields `(EulerGamma-Log[1/n])+O[1/n]^1`

you can keep the "trick" or just return a tuple `(s1_coeff, explicit_const)` and recombine that in the suitable places

felixhekhorn (changes requested):
please double check that the bug I claim actually is one

extras/ome_n3lo/convert_ome_xspace.py
```diff
@@ -0,0 +1,100 @@
...
+    cache = c.reset()
+    if is_singlet:
+        return as3.A_singlet(n, cache, nf, L=0)
+    else:
+        return as3.A_ns(n, cache, nf, L=0)
+
+
+def compute_xspace_ome(entry, nf, x_grid=XGRID):
+    """Compute the x-space transition matrix element, returns A^3(x)."""
+    mellin_cut = 5e-2
```
felixhekhorn, line 45: I only checked `["qq_ns", "Hg", "Hq"]`, but for them I could push that down. I only tried `5e-3`, but we may make it even smaller. This would help again in the large-x region (because there we need to integrate "more")

extras/ome_n3lo/convert_ome_xspace.py
```diff
@@ -0,0 +1,100 @@
...
+    ome_x = []
+    print(f"Computing operator matrix element {entry} @ pto: 3, nf: {nf}")
+    # loop on xgrid
+    with progressbar(x_grid) as bar:
+        for x in bar:
+            if x == 1:
+                ome_x.append(0)
+                continue
+            res = integrate.quad(
+                lambda u: integrand(u, x),
```
felixhekhorn, line 77: minor detail: I would pass directly $\log(x)$ and make $x^{-n} =\exp(-n\log(x))$ explicit

extras/ome_n3lo/convert_ome_xspace.py
```diff
@@ -0,0 +1,100 @@
...
+def compute_xspace_ome(entry, nf, x_grid=XGRID):
+    """Compute the x-space transition matrix element, returns A^3(x)."""
+    mellin_cut = 5e-2
+    is_singlet = "ns" not in entry
+
+    def integrand(u, x):
+        """Mellin inversion integrand."""
+        path = Path(u, np.log(x), is_singlet)
+        integrand = path.prefactor * x ** (-path.n) * path.jac
+        if integrand == 0.0:
```
felixhekhorn, line 52: is that ever true?

extras/ome_n3lo/convert_ome_xspace.py
```diff
@@ -0,0 +1,100 @@
...
+from eko.interpolation import lambertgrid
+from scipy import integrate
+
+from large_n_limit import Agg_asymptotic, Aqq_asymptotic
+
+XGRID = lambertgrid(500, 1e-6)
+"""X-grid."""
+
+LOG = 0
+"""Matching threshold displaced ?"""
```
felixhekhorn, line 19:
```suggestion
```
it is not used, is it?

extras/ome_n3lo/convert_ome_xspace.py
```diff
@@ -0,0 +1,100 @@
+"""Dump a fast  x-space grid of the N3LO transition matrix elements.
```
felixhekhorn, line 1:
```suggestion
"""Dump a fast x-space grid of the N3LO transition matrix elements.
```

extras/ome_n3lo/convert_ome_xspace.py
giacomomagni: Okay shall I set this to `5e-3` then?

extras/ome_n3lo/convert_ome_xspace.py
giacomomagni: 😅, but it won't change anything ?

extras/ome_n3lo/convert_ome_xspace.py
giacomomagni: At x=1 might cause problem, so let's keep it!

extras/ome_n3lo/convert_ome_xspace.py
giacomomagni: Fixed in 00fad3a9

extras/ome_n3lo/convert_ome_xspace.py
giacomomagni: Fixed in 00fad3a9

extras/ome_n3lo/large_n_limit.py
giacomomagni: Agreed, I'll provide them a mathematica file with the coefficients. 
Thanks for double checking.

extras/ome_n3lo/convert_ome_xspace.py
felixhekhorn: yeah mostly cosmetics (it's just more natural to me as it really depends on log(x), but they are bijective - fine)

felixhekhorn (approved):

giacomomagni:
As the paper has been accepted, I'm happy to merge.

PR #448, opened 2025-03-10
Use `dynamic_versioning` as backend

### LINKED ISSUE #365: Version being saved as 0.0.0

scarlehoff:
put dynamic_versioning as the backend so that one gets 0.14.whatever instead of 0.0.0 when using it locally, not sure whether that will mess up with something?

closes #365


felixhekhorn:
> I also put dynamic_versioning as the backend so that one gets 0.14.whatever instead of 0.0.0 when using it locally, not sure whether that will mess up with something?

Mmm, that is dangerous: because e.g. master is at the moment definitely NOT 0.14 ... and as I was arguing in #417 with @evagroenendijk it is wrong to call master v0.15, because v0.15 might (or might not) contain even more breaking changes than now; so I think I prefer v0.0 - which should never appear in public codes any ways, but only for developers (who can deal with it)

scarlehoff:
> which should never appear in public codes any ways, but only for developers

But in practice we have many fktables and ekos produced with 0.0.0.

But I do see the point, perhaps a better solution would be to (for a developing installation) get the commit without the tag?

felixhekhorn:
> But I do see the point, perhaps a better solution would be to (for a developing installation) get the commit without the tag?

is it possible to just have the commit?
 
>  `0.14.6+<number of commits since tag>+<commit hash>`

or do we always need the full thing?

in any case we will need to adjust the legacy handling (initialised in #417) - specifically, we need to adjust this logic: https://github.com/NNPDF/eko/blob/35e1a55b32ba0fbc5115790f7a12e73103cecae9/src/eko/io/metadata.py#L67
and upload a new file version as explained here https://github.com/NNPDF/eko/blob/master/tests/data/assets.sh

felixhekhorn:
> At the moment if EKO reads an object through a symlink and it is not in readonly mode, it will overwrite the symlink instead of the original file.
> 
> This PR modifies `read` so that it goes to the original file instead.

do you want to split this change from this PR? since I guess that change is uncontroversial (and does hopefully have no tail)

pyproject.toml
```diff
@@ -1,6 +1,6 @@
 [build-system]
 requires = ["poetry-core>=1.0.0", "poetry-dynamic-versioning"]
-build-backend = "poetry.core.masonry.api"
+build-backend = "poetry_dynamic_versioning.backend"
```
scarlehoff, line 3:
```suggestion
build-backend = "poetry.core.masonry.api"
```

felixhekhorn:
> How does the legacy handler deals with 0.0.0 since it can be any version? (and adding the commit won't solve that)

ehm simply this https://github.com/NNPDF/eko/blob/35e1a55b32ba0fbc5115790f7a12e73103cecae9/src/eko/io/metadata.py#L69
and this
https://github.com/NNPDF/eko/blob/35e1a55b32ba0fbc5115790f7a12e73103cecae9/src/eko/io/metadata.py#L71
is false and so no recover is applied - i.e. it assumes it is master. We only support versions after v0.13.5

- if we do 0.14.whatever we have to change the conditions: i.e. check for `data_version` first 
- if we do something else, we need to check if [packaging](https://packaging.pypa.io/en/stable/) likes that and then adjust the conditions if necessary

src/eko/io/struct.py
```diff
@@ -386,6 +386,8 @@ def read(
         format. Otherwise, the `path` is interpreted as the location of an
         already extracted folder.
         """
+        # Take the absolute path in case we need to modify the eko in-place
+        path = path.resolve()
```
felixhekhorn, line 390:
```suggestion
```

evagroenendijk:
@felixhekhorn I added some stuff, but note that I'm still working on it! Not everything is working properly yet

src/eko/io/v3.py
felixhekhorn, line 1: Do you expect we need this file?

felixhekhorn (commented):
just some comments

src/eko/io/v3.py
evagroenendijk: It was what we agreed right? To start it already at the beginning of a new version and then add all the stuff we change along the way? If I misremember please let me know of course!

felixhekhorn: yes and no :upside_down_face: we should add it at the beginning of a new _data version_ , not a new API version. v3 is the new default and the program is written such that it work natively with that version - any other version has to be cast to this version (which is what happens in `vX.py`) - since v3 does not need any transformation, I don't think we need this file (it would be and is at present an identity). The moment we work on v4, i.e. the moment we change the content of the theory file etc. again (and we will sooner or later), we start `v3.py` (which will then cast to v4 - and we have to adjust all the others of course as well).

just as a reminder: the data version defines how the files look like, the API version defines how the interface looks like

evagroenendijk: Ah yes okay, so the moment we break something (I mean when the theory/operator files start looking differently) we define the v3 and start putting every change there. I guess I was getting ahead of myself by already defining the new data version. Thanks for the reminder!

felixhekhorn:
> @felixhekhorn So I thought it would be good if I write some instructions about what to do when we add a new data version, and was wondering where I should put this? Should I add it to the readme.md or is there a better place?

good idea and good question: I'd say the `README.md` is the landing page for our users, let's not expose "internal" stuff there. Let's use a dedicated file ... `data-versions.md`? `maintainers.md`? `releases.md`? more idea from you? :upside_down_face:  what do we want to stuff in there?

We should maybe also push [my blabla](https://github.com/NNPDF/eko/blob/37bee66e4617e9650b5350f756a38f2feb214b19/tests/data/assets.sh#L2) from the assets.sh there

src/eko/io/v3.py
felixhekhorn: remember to resolve conversations if you believe to have addressed the issue

evagroenendijk:
> > @felixhekhorn So I thought it would be good if I write some instructions about what to do when we add a new data version, and was wondering where I should put this? Should I add it to the readme.md or is there a better place?
> 
> good idea and good question: I'd say the `README.md` is the landing page for our users, let's not expose "internal" stuff there. Let's use a dedicated file ... `data-versions.md`? `maintainers.md`? `releases.md`? more idea from you? 🙃 what do we want to stuff in there?
> 
> We should maybe also push [my blabla](https://github.com/NNPDF/eko/blob/37bee66e4617e9650b5350f756a38f2feb214b19/tests/data/assets.sh#L2) from the assets.sh there

Okay, I will make a dedicated file! I'll commit it here when I'm done and you can check if you want anything different

evagroenendijk:
@felixhekhorn I added some instructions, if you want to have a look over it! Other than that, I think it is ready for review, what do you think?

felixhekhorn:
> I think it is ready for review, what do you think?

if you think it is ready, please mark it so from the interface :upside_down_face: and then I start to complain :see_no_evil: (as usual the direction is good, but details can be improved)

src/eko/maintainers.md
felixhekhorn, line 1: The `.md` extension of this file is to indicate [Markdown - or in this case GitHub Flavored Markdown](https://docs.github.com/en/get-started/writing-on-github/getting-started-with-writing-and-formatting-on-github/basic-writing-and-formatting-syntax) (the formal definition is [here](https://github.github.com/gfm/)) - please format the file accordingly

src/eko/maintainers.md
```diff
@@ -0,0 +1,61 @@
...
+#    Note that this may overwrite an existing file, so be sure on what you do!
+#    If necessary, add the file to the list in this script.
+# 4. Update the data version in the workflows.
+#    Update both Rust and Python workflows, by incrementing the data version,
+#    such that the new files are also triggered in the CI.
+#
+# You can test the compatibility of your eko version & the test assets
+# with pytest, specifically with the test_legacy.py script.
+#
+#
```
felixhekhorn, line 59: Let's add a note that we refer to the unreleased version as v0.0

tests/data/assets.sh
```diff
@@ -29,8 +11,8 @@ RUSTDATADIR="crates/dekoder/tests/data/"
 # download from the server
 curl -s -C - -o "./${PYDATADIR}v0.13.tar" "${URL}v0.13.tar"
 curl -s -C - -o "./${PYDATADIR}v0.14.tar" "${URL}v0.14.tar"
-curl -s -C - -o "./${PYDATADIR}v0.0.tar" "${URL}v0.0.tar"
+curl -s -C - -o "./${PYDATADIR}v0.15.tar" "${URL}v0.15.tar"
```
felixhekhorn, line 14: let's rename the files both locally and remotely (i.e. on the server):

- `v0.13.tar` -> `v1-0.13.tar`
- `v0.14.tar` -> `v1-0.14.tar`
- `v0.15.tar` -> `v3.tar`

for the usual reason we have to fake v1, but henceforth the files correspond to a given data version and not a API version

src/eko/io/metadata.py
```diff
@@ -65,11 +65,23 @@ def load(cls, path: os.PathLike):
         # read raw file first to catch version
         raw = yaml.safe_load(paths.metadata.read_text(encoding="utf-8"))
         version = parse(raw["version"])
+        data_version = raw["data_version"]
```
felixhekhorn, line 68: let's cast to `int` here

src/eko/io/metadata.py
```diff
@@ -65,11 +65,23 @@ def load(cls, path: os.PathLike):
...
-        if version.major == 0 and version.minor == 13:
-            raw = v1.update_metadata(paths, raw)
-        elif version.major == 0 and version.minor == 14:
-            raw = v2.update_metadata(paths, raw)
+        if data_version == 1:
+            if version.major == 0 and version.minor == 13:
+                raw = v1.update_metadata(paths, raw)
+            elif version.major == 0 and version.minor == 14:
+                raw = v2.update_metadata(paths, raw)
+            elif version.major == 0 and version.minor == 0 and "post" in raw["version"]:
```
felixhekhorn, line 75: `raw` should not be accessed any longer at this point - use the object instead! (i.e. `version`) I'm sure you can do a similar trick there

crates/dekoder/tests/test_load.rs
```diff
@@ -39,19 +39,19 @@ fn destroy() {
...
 #[test]
 fn save_as_other() {
-    let src = v00tar();
+    let src = v015tar();
     let dst = assert_fs::TempDir::new().unwrap();
     // open
     let eko = EKO::extract(src.to_owned(), dst.to_owned()).unwrap();
     // write to somewhere else
-    let tarb = assert_fs::NamedTempFile::new("v0.0b.tar").unwrap();
+    let tarb = assert_fs::NamedTempFile::new("v0.15b.tar").unwrap();
```
felixhekhorn, line 47:
```suggestion
    let tarb = assert_fs::NamedTempFile::new("write_test.tar").unwrap();
```
this is a temporary name anyways, let's fix this

felixhekhorn (changes requested):

src/eko/io/metadata.py
```diff
@@ -72,20 +72,22 @@ def load(cls, path: os.PathLike):
...
-            elif "post" in str(version):
+            if (
+                version.major == 0
+                and version.minor == 0
+                and version.micro == 0
+                and "post" in str(version)
+            ):
                 raise NotImplementedError(
                     "Unsupported version; use an eko from a published eko version!"
                 )
```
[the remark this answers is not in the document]
felixhekhorn: definitely we need to be able to read the present file format - don't we have a unit test somewhere for that? (it might be somewhere else in .../io/) In which situation would a version `0.15.1-post123` appear?

src/eko/io/metadata.py
```diff
@@ -65,11 +65,33 @@ def load(cls, path: os.PathLike):
...
+        if data_version == 1:
+            if version.major == 0 and version.minor == 13:
+                raw = v1.update_metadata(paths, raw)
+            elif version.major == 0 and version.minor == 14:
+                raw = v2.update_metadata(paths, raw)
+            elif (
+                version.major == 0
+                and version.minor == 0
+                and version.micro == 0
+                and "post" in str(version)
```
felixhekhorn, line 79: again, let's use the object `version` and not it's string representation; i.e. isn't this https://packaging.pypa.io/en/stable/version.html#packaging.version.Version.post what you want? or better this https://packaging.pypa.io/en/stable/version.html#packaging.version.Version.is_postrelease ?

src/eko/io/metadata.py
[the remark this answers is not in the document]
evagroenendijk: @felixhekhorn Yes indeed there is a unit test for it in test_struct.py I think! Okay, then I will leave it like this

src/eko/io/metadata.py
evagroenendijk: Okay, I did not realise that it was a problem to use strings, I thought this would be the easiest way?

But I'll change it then!

felixhekhorn: we may need to force poetry to comply with [PEP404](https://peps.python.org/pep-0440/) - check here https://pypi.org/project/poetry-dynamic-versioning/, which [packaging](https://packaging.pypa.io/en/stable/index.html) assumes

felixhekhorn:
Could you please check your git config (which I believe is the problem), such that your commits are associated to you:
grafik
(your avatar should appear there)

It's not very important, but just a bit confusing - and you won't get any credits [here](https://github.com/NNPDF/eko/graphs/contributors), which is what Juan CM is showing at the NNPDF meetings and where we demonstrate we actually do something :see_no_evil:

src/eko/io/metadata.py
evagroenendijk: Okay got it. Thanks!

evagroenendijk: > we may need to force poetry to comply with [PEP404](https://peps.python.org/pep-0440/) - check here https://pypi.org/project/poetry-dynamic-versioning/, which [packaging](https://packaging.pypa.io/en/stable/index.html) assumes

So I think this is fine now, because I am using the ".post", which seems to work in both, assuming that is what you meant!

felixhekhorn: what is the difference with https://packaging.pypa.io/en/stable/version.html#packaging.version.Version.is_postrelease?

evagroenendijk: It is the same, only version.is_postrelease gives back True or False, while version.post gives the postrelease number

felixhekhorn:
> @felixhekhorn I get an error from pre-commit that does not really make me any wiser... Have you perhaps seen it before?

I'm not sure either, but I think something just got a time-out (like reading from github) and I don't think we should worry, for know - let's see at the next attempt

evagroenendijk:
> Could you please check your git config (which I believe is the problem), such that your commits are associated to you: grafik (your avatar should appear there)
> 
> It's not very important, but just a bit confusing - and you won't get any credits [here](https://github.com/NNPDF/eko/graphs/contributors), which is what Juan CM is showing at the NNPDF meetings and where we demonstrate we actually do something 🙈

Ah sorry, I just see this now. I updated my git user.name in the git config, would that be enough? Sorry, I wasn't aware of this problem

src/eko/maintainers.md
```diff
@@ -0,0 +1,50 @@
+# INSTRUCTIONS REGARDING (DATA) VERSIONS
+
+## Instructions for eko maintainers on how to deal with breaking data versions
+
+### a) Instructions on what to do when we break the data format:
+
+When we break the data format (i.e. if a new data version of eko creates EKO's with a different format of theory/operator cards), the new eko version will read these cards in a different way, and we need to make some changes.
```
felixhekhorn:
```suggestion
When we break the data format, we need to ensure backward compatibility and thus we need to make some changes:
```
let's be more concise

src/eko/maintainers.md
```diff
@@ -0,0 +1,50 @@
...
+
+1. Create a file `vX.py` with X the data version (analogous to `v1.py` and `v2.py`).
```
felixhekhorn, line 9: Mention somewhere which directory we are talking about

src/eko/maintainers.md
```diff
@@ -0,0 +1,50 @@
...
+3. Modify `versions.py` to the new data version.
+4. During the development of this new data version (i.e. until the moment the data version breaks again and we move to version X+1), fill up `vX.py` with all breaking changes.
+5. Also implement these changes in the old `vX.py` scripts (at the moment these are `v1.py` and `v2.py`).
+
+> **Note**: We call the unreleased eko version `v0.0`. EKO's created with unpublished versions (`0.0.0` versions) are not part of this backward compatibility system.
+> They have version `0.0.0-{commit}`, and they cannot necessarily be read by newer eko versions.
+
+---
+
+### b) Instructions on how to use the test assets used by pytest:
```
felixhekhorn, line 20:
```suggestion
### b) Instructions on how to use the test assets:
```
since Rust is also in the game

src/eko/maintainers.md
```diff
@@ -0,0 +1,50 @@
...
+
+Using `pytest`, we also test if eko handles legacy EKO's in the correct way. When we move to a new eko version, it is important to keep these test assets up to date.
```
felixhekhorn, line 22:
```suggestion
Using unit tests we check if eko handles legacy EKO's in the correct way. When we move to a new eko version, it is important to keep these test assets up to date.
```

src/eko/maintainers.md
```diff
@@ -0,0 +1,50 @@
...
+
+In the directory `tests/data`, you will find the script `assets.sh` that downloads the currently existing test assets. The script is to be executed in the root directory of the repository. Currently, these assets are only older and current eko objects, which we test for readability. The number of versions here determines how backward compatible we are. The eko objects must all correspond to the LO FFNS Les Houches benchmark.
```
felixhekhorn, line 24:
```suggestion
In the directory `tests/data`, you will find the script `assets.sh` that downloads the currently existing test assets from the NNPDF server.
The script is to be executed in the root directory of the repository. Currently, these assets are only older and current eko objects, which we test for readability. The number of versions here determines how backward compatible we are. The eko objects must all correspond to the LO FFNS Les Houches benchmark.
```

src/eko/maintainers.md
```diff
@@ -0,0 +1,50 @@
...
+To generate a new file, proceed as follows:
+
+1. **Invoke the benchmark**
+   This can be easily achieved with Poe. Execute (in the repo root dir):
+   ```bash
+   poe lha -m "ffns and lo"
+   ```
+
+2. **Move the generated object locally**
+   The above command will create a file in `benchmarks/data/`. Move the file here and include it into the unit tests (consider both Rust and Python).
```
felixhekhorn, line 35:
```suggestion
   The above command will create a file in `benchmarks/data/`. Move the file to `tests/data` and include it into the unit tests (consider both Rust and Python).
```

src/eko/maintainers.md
```diff
@@ -0,0 +1,50 @@
...
+   ```
+   > Note that this may overwrite an existing file, so be sure about what you're doing!
+   > If necessary, add the file to the list in this script.
+
+4. **Update the data version in the workflows**
+   Update both Rust and Python workflows by incrementing the data version, so the new files are also triggered in the CI.
+
+---
+
+You can test the compatibility of your eko version and the test assets with `pytest`, specifically with the `test_legacy.py` script.
```
felixhekhorn, line 50:
```suggestion
Finally, you can test the compatibility of your eko version and the test assets in the usual way with `pytest`, specifically with the `test_legacy.py` script.
```

felixhekhorn (changes requested):
Can you please also add an explicit item to the `Changelog.md`? in fact, maybe we should add a general reminder for the maintainers into the new file (since this is related to releasing). You could even define a new prefix there `data` for data related stuff

felixhekhorn:
> Ah sorry, I just see this now. I updated my git user.name in the git config, would that be enough? Sorry, I wasn't aware of this problem

not 100% sure what is need (search for it), but maybe the e-mail is more important?

CHANGELOG.md
```diff
@@ -24,3 +26,6 @@ Items without prefix refer to a global change.
 - py: Resolve files in output to actually edit the true file
 
 ## [0.15.0](https://github.com/NNPDF/eko/compare/v0.14.6...v0.15.0) - 2025-03-06
+
+### Added
+- data: Added backwards compatibility system as explained in `src/maintainers.md` ([#448](https://github.com/NNPDF/eko/pull/448))
```
felixhekhorn, line 31: - this belongs of course to the "Unreleased" section
- let's be more concise:
  - we didn't add the backward comp. sys. here - that we did #417 
  - instead, we changed for versioning system for unreleased EKOs: from `0.0.0` to `0.0.0-sha` (this is the new format, right?)
  - keep the reference to the PR
  - I would [classify](https://keepachangelog.com/en/1.1.0/) this as "Changed", or what do you think?

CHANGELOG.md
```diff
@@ -5,10 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+
```
felixhekhorn, line 8: No need to introduce more whitespace here (unless I'm missing something?) Stupid Github won't let me suggest the change right away :upside_down_face:

src/eko/maintainers.md
```diff
@@ -0,0 +1,49 @@
+# INSTRUCTIONS REGARDING (DATA) VERSIONS
+
+## Instructions for eko maintainers on how to deal with breaking data versions
+
+### a) Instructions on what to do when we break the data format:
```
felixhekhorn, line 5:
```suggestion

## API versions

- Remember to update the `Changelog.md`

## Data versions

### Instructions on what to do when we break the data format:
```

src/eko/maintainers.md
```diff
@@ -0,0 +1,49 @@
...
+3. Modify `src/eko/version.py` to the new data version.
+4. During the development of this new data version (i.e. until the moment the data version breaks again and we move to version X+1), fill up `vX.py` with all breaking changes.
+5. Also implement these changes in the old `vX.py` scripts (at the moment these are `v1.py` and `v2.py`).
+
+> **Note**: We call the unreleased eko version `v0.0`. EKO's created with unpublished versions (`0.0.0` versions) are not part of this backward compatibility system.
+> They have version `0.0.0-{commit}`, and they cannot necessarily be read by newer eko versions.
+
+---
+
+### b) Instructions on how to use the test assets:
```
felixhekhorn, line 20:
```suggestion
### Instructions on how to update the test assets:
```

src/eko/maintainers.md
```diff
@@ -0,0 +1,49 @@
...
+1. Create a file `vX.py` with X the data version (analogous to `v1.py` and `v2.py`) in `src/eko/io`.
+2. Make the corresponding changes to `metadata.py` and `struct.py` (also in `src/eko/io`).
+3. Modify `src/eko/version.py` to the new data version.
+4. During the development of this new data version (i.e. until the moment the data version breaks again and we move to version X+1), fill up `vX.py` with all breaking changes.
+5. Also implement these changes in the old `vX.py` scripts (at the moment these are `v1.py` and `v2.py`).
+
+> **Note**: We call the unreleased eko version `v0.0`. EKO's created with unpublished versions (`0.0.0` versions) are not part of this backward compatibility system.
+> They have version `0.0.0-{commit}`, and they cannot necessarily be read by newer eko versions.
+
+---
```
felixhekhorn, line 18:
```suggestion
```

src/eko/maintainers.md
felixhekhorn, line 1: - Move this file in to root, i.e. `/`
- since apparently I chose all upper case for `CHANGELOG.md` let's do the same here, i.e. `/MAINTAINERS.md`

felixhekhorn (changes requested):
more minor changes, but then I hope we are done :see_no_evil:

tests/data/assets.sh
```diff
@@ -1,24 +1,5 @@
-#!/usr/bin/bash
```
felixhekhorn, line 1: Why did you remove this line? is it a problem with Mac?

evagroenendijk: Ah yes for me it was a problem, but I meant to put it back only I forgot sorry! I'll put it back now

Issue #216, opened 2023-02-23
Add CHANGELOG/updated Documentation


Radonirinaunimi:
Since EKO is now being used in multiple projects, would it be possible to have **change logs** whenever breaking changes are added in tags/releases? For the past couple of weeks, in our [nnusf](https://github.com/NNPDF/nnusf/) project, I've often found myself  finding changes that break backward compatibility (see for example https://github.com/NNPDF/nnusf/issues/85). Or at least the corresponding tutorials [alpha_s.ipynb](https://github.com/NNPDF/eko/blob/master/doc/source/overview/tutorials/alpha_s.ipynb) should be updated.

AleCandido:
It has always been a nice-to-have, but never top-priority.

Essentially, doing it manually requires making all developers and reviewers responsible for the updates, making the process heavier and fallible.
And since there is not a single way to translate PRs in a change-log, I always postponed the problem for better investigation.

If you find a similar project with a good enough workflow (requiring little manual effort), I'd be glad to adopt the strategy!

felixhekhorn:
the above refers to the CHANGELOG - the documentation surely has to be maintained as much as we can ...

just to mention: of course this is closely related to #145

AleCandido:
> the above refers to the CHANGELOG - the documentation surely has to be maintained as much as we can ...

Agreed, but we are already putting a bit of effort to maintain the docs, and contributors are more and more aware of the necessity of maintaining updated docs (I hope...). And this for sure has to be done manually.

The automated CHANGELOG proposal is just to avoid extra burden. If we have too many things to maintain manually, at some point none of them will be sufficiently important any longer...

Radonirinaunimi:
I fully understand that (unfortunately) in all cases some efforts will be needed. But as @AleCandido alluded there might be some ways to alleviate this burden. The best one I know for example is [github-changelog-generator](https://github.com/github-changelog-generator/github-changelog-generator), which also has been implemented in various github action for further ease.

AleCandido:
> The best one I know for example is [github-changelog-generator](https://github.com/github-changelog-generator/github-changelog-generator), which also has been implemented in various github action for further ease.

Not very maintained in the last year, but sufficiently popular. It seems stable enough to fit the purpose.
Can you also point out one of these actions?

Radonirinaunimi:
[...]

PR #459, opened 2025-04-28
Remove `ev_op_iterations` from truncated

### LINKED ISSUE #447: Don't use `ev_op_iterations` for "trucated"
this was already confusing in [evolven3fit](https://github.com/NNPDF/nnpdf/blob/17b53d34f1eb6c3ed067e1475742517802b5ddd4/n3fit/src/evolven3fit/eko_utils.py#L16) and it is also confusing for others (cc @mcerutti996) - let's not use it (i.e. hard-code it to 1)

felixhekhorn:
Closes #447 

As a consequence `ev_op_iterations` also dropped from `kernels.ns.dispatcher`, i.e. in the non-singlet sector there is no iteration anywhere (since exact _is_ exact). This triggers a few more changes on the testing side.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,26 @@
+# Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
+and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
+
+We may prefix the items to signal the scope:
+
+- py: Python library
+- rust: Rust library
+
+Items without prefix refer to a global change.
+
+## [Unreleased](https://github.com/NNPDF/eko/compare/v0.15.1...HEAD)
+
+### Fixed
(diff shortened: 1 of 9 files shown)


giacomomagni:
This looks fine for me, the only other option I see is to hardcode it to 1 in the function signature. But also dropping it's fine!

felixhekhorn:
> This looks fine for me, the only other option I see is to hardcode the default value to 1 in the function signature. But also dropping it's fine!

I would prefer to remove it completely to also not confuse developers :upside_down_face: - if you're happy please approve :innocent: 

Since I just told our PhD students here how important documentation and as a part of that a Changelog is I finally started our Changelog here :see_no_evil: thus: Closes https://github.com/NNPDF/eko/issues/216

Note that in 0.15 there are so many changes that it was impossible for me to spell it out quickly (remember that all the long-standing PRs from back in the days (e.g. #295) got merged there)

giacomomagni (approved):

Issue #464, opened 2025-05-20
Version format key is not correct


felixhekhorn:
Looking [here](https://github.com/NNPDF/eko/actions/runs/15119575048/job/42533050293?pr=463#step:7:14) I see that you mistyped the version format configuration key
https://github.com/NNPDF/eko/blob/aaa0325cb0187b8408e93093231e594574ad8c8f/pyproject.toml#L95
`format_jinja` -> `format-jinja` as said [here](https://github.com/mtkennerly/poetry-dynamic-versioning)

PS: why didn't you hit this during development? i.e. did you actually try if it does what you want 🙈 ?

evagroenendijk:
> Looking [here](https://github.com/NNPDF/eko/actions/runs/15119575048/job/42533050293?pr=463#step:7:14) I see that you mistyped the version format configuration key
> 
> [eko/pyproject.toml](https://github.com/NNPDF/eko/blob/aaa0325cb0187b8408e93093231e594574ad8c8f/pyproject.toml#L95)
> 
> Line 95 in [aaa0325](/NNPDF/eko/commit/aaa0325cb0187b8408e93093231e594574ad8c8f)
> 
>  format_jinja = "{% if distance == 0 %}{{ base }}{% else %} 0.0.0.{{ commit }}{% endif %}" 
> 
> `format_jinja` -> `format-jinja` as said [here](https://github.com/mtkennerly/poetry-dynamic-versioning)
> PS: why didn't you hit this during development? i.e. did you actually try if it does what you want 🙈 ?

Ah, so sorry! I tested it in the beginning and it worked but I must have made a change & the typing error after this. Sorry, should have tested it at the end! I'll fix it now

PR #465, opened 2025-05-20
Solve issue #464

evagroenendijk:


felixhekhorn (changes requested):
First, some stupid remarks:
- For our own records it's better to have a more descriptive title of a PR. "Solve issue 464" is not very clear afterwards, since you need to know what 464 was in the first place. GitHub allows to link PR and issues in a native way, so we can make that connection elsewhere (I did it here already) (see also [here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests))
- similar considerations apply to commit messages: let's try to keep them more descriptive and to the point - I know it is not easy. If you haven't done so yet, please read [this blogpost](https://cbea.ms/git-commit/) :upside_down_face: 

Second, more serious comments:
- can you please add an explicit unit test against this? Instead of me checking it once, let's automatise. So, it either has to be a definite version XOR 0.0.0-postNUMBER-COMMIT ; i.e. in `tests/eko/io/test_struct.py`
- can you please update the CHANGELOG? modify the existing entry with the correct format and just append this PR

evagroenendijk:
> First, some stupid remarks:
> 
> * For our own records it's better to have a more descriptive title of a PR. "Solve issue 464" is not very clear afterwards, since you need to know what 464 was in the first place. GitHub allows to link PR and issues in a native way, so we can make that connection elsewhere (I did it here already) (see also [here](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests))
> * similar considerations apply to commit messages: let's try to keep them more descriptive and to the point - I know it is not easy. If you haven't done so yet, please read [this blogpost](https://cbea.ms/git-commit/) 🙃
> 
> Second, more serious comments:
> 
> * can you please add an explicit unit test against this? Instead of me checking it once, let's automatise. So, it either has to be a definite version XOR 0.0.0-postNUMBER-COMMIT ; i.e. in `tests/eko/io/test_struct.py`
> * can you please update the CHANGELOG? modify the existing entry with the correct format and just append this PR

Thanks a lot for your comments, will do! Then I wanted to ask one other thing, if you have time: in the metadata.py I have now added this extra condition wrt the version, but was wondering if there is a more elegant way to define the eko version that you are using? I remember that defining stuff with strings was not so nice, but I also did not really find a direct command for it

tests/eko/io/test_struct.py
```diff
@@ -195,3 +196,20 @@ def test_load_opened(self, tmp_path: pathlib.Path, eko_factory: EKOFactory):
...
+        """Test asserted version. Should either be supported version, or have a postrelease addition"""
+        eko = eko_factory.get()
+        eko.close()
+        eko_factory.cache = None
+        assert eko.access.path is not None
+        read_closed = EKO.read(eko.access.path, dest=tmp_path)
+        supported_versions = [13, 14, 15]
+
+        version = parse(read_closed.metadata.version)
+        assert (version.major == 0 and version.minor in supported_versions) or (
```
felixhekhorn, line 210: let's not be so restrictive (else we need to remember to update here), so I suggest to change the first part of the `or` to `(version.major + version.minor + version.micro > 0 and not version.is_postrelease)` i.e. it is an actual release

CHANGELOG.md
```diff
@@ -17,6 +17,7 @@ Items without prefix refer to a global change.
 
 ### Changed
 - data: Changed version naming for unreleased eko versions from `0.0.0` to `0.0.0-sha` ([#448](https://github.com/NNPDF/eko/pull/448))
+- data: Change version naming for unreleased eko versions to `0.0.0-post.#+{commit hash}` ([#465](https://github.com/NNPDF/eko/pull/465))
```
felixhekhorn, line 20: Merge the two, e.g. 
```
- data: Change version naming for unreleased eko versions to `0.0.0-post.{distance}+{commit hash}` ([#448](https://github.com/NNPDF/eko/pull/448)) ([#465](https://github.com/NNPDF/eko/pull/465))
```
they are really one change: this one is just a bugfix

tests/eko/io/test_struct.py
```diff
@@ -204,10 +204,12 @@ def test_version(self, tmp_path: pathlib.Path, eko_factory: EKOFactory):
         eko_factory.cache = None
         assert eko.access.path is not None
         read_closed = EKO.read(eko.access.path, dest=tmp_path)
-        supported_versions = [13, 14, 15]
 
         version = parse(read_closed.metadata.version)
-        assert (version.major == 0 and version.minor in supported_versions) or (
+        assert (
+            version.major + version.minor + version.micro >= 0
+            and version.is_postrelease is False
```
felixhekhorn, line 211:
```suggestion
            and not version.is_postrelease
```
my personal opinion :upside_down_face: 3 characters against 8

src/eko/io/metadata.py
```diff
@@ -77,6 +77,8 @@ def load(cls, path: os.PathLike):
                 and version.minor == 0
                 and version.micro == 0
                 and version.is_postrelease
+                and vmod.__version__
```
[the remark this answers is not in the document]
evagroenendijk: Hmm, yes the problem is that the names are defined slightly differently I think (a . instead of a -). I will see if I can fix that!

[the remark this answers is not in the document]
felixhekhorn: mh? one is derived from the other - actually now that I think about it: why is that needed in the first place? why do we read the API version (`vmod.__version__`)? this tells us nothing about the file we are dealing with, no? maybe we want
```py
elif (
                version.major == 0
                and version.minor == 0
                and version.micro == 0
                and not version.is_postrelease):
```
maybe? or something similar - i.e. if it is just `0.0.0` (and no suffixes), it is an old eko and we give up. Or we drop the check and just hope for the best - and if we can't read we will fail at another point (with a random error message like "key not found" or whatever)

tests/eko/io/test_struct.py
```diff
@@ -195,3 +196,22 @@ def test_load_opened(self, tmp_path: pathlib.Path, eko_factory: EKOFactory):
...
+        """Test asserted version. Should either be supported version, or have a postrelease addition"""
+        eko = eko_factory.get()
+        eko.close()
+        eko_factory.cache = None
+        assert eko.access.path is not None
+        read_closed = EKO.read(eko.access.path, dest=tmp_path)
+
+        version = parse(read_closed.metadata.version)
+        assert (
+            version.major + version.minor + version.micro >= 0
```
felixhekhorn, line 210:
```suggestion
            version.major + version.minor + version.micro > 0
```
no? we don't want 0.0.0

felixhekhorn: we can even make it stricter: already major+minor has to be > 0

src/eko/io/metadata.py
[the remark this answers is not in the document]
evagroenendijk: So what I was thinking is that the eko version you are running has to be able to to read its own eko's (as is also tested in the unit tests). So I want to do "and not (eko version you have installed now) = (eko version of object)", but now that I am thinking of course I did not accurately define that at all now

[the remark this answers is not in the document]
evagroenendijk: I agree! And at least we catch the old eko's with 0.0.0 here already. I'll make the last changes

MAINTAINERS.md
```diff
@@ -17,7 +17,7 @@ We need to ensure backward compatibility and thus we need to make some changes.
 5. Also implement these changes in the old `vX.py` scripts (at the moment these are `v1.py` and `v2.py`).
 
 > **Note**: We call the unreleased eko version `v0.0`. EKO's created with unpublished versions (`0.0.0` versions) are not part of this backward compatibility system.
-> They have version `0.0.0-{commit}`, and they cannot necessarily be read by newer eko versions.
+> They have version `0.0.0-{commit}` or `0.0.0`, and they cannot necessarily be read by newer eko versions.
```
felixhekhorn, line 20: why? we don't want that - that defines the purpose of the original idea (and this is indeed not what the unit test checks)

src/eko/io/metadata.py
```diff
@@ -73,23 +73,17 @@ def load(cls, path: os.PathLike):
             elif version.major == 0 and version.minor == 14:
                 raw = v2.update_metadata(paths, raw)
             elif (
```
felixhekhorn, line 75: I suggest to drop this whole `elif` branch

src/eko/io/metadata.py
```diff
@@ -73,23 +73,17 @@ def load(cls, path: os.PathLike):
...
-                version.major == 0
-                and version.minor == 0
-                and version.micro == 0
-                and version.is_postrelease
+                 version.base_version == str(version) == '0.0.0'
             ):
                 raise NotImplementedError(
                     "Unsupported version; use an eko from a published eko version!"
                 )
         else:
```
felixhekhorn, line 81: I suggest to drop this whole `else` branch

MAINTAINERS.md
evagroenendijk: So it can be that there is an older eko made before our dynamic-versioning changes with version "0.0.0", right? I can also remove this line of course, but I thought this would be more complete?

src/eko/io/metadata.py
evagroenendijk: I can do it of course! But I am wondering why? Because now at least we catch the old 0.0.0 eko's and give an understandable error message

MAINTAINERS.md
felixhekhorn: but henceforth there will be no more `0.0.0`, right? this document is talking about *new* version only

felixhekhorn:
I think the remaining problem is that poetry-dynamic-versioning is not working correctly.
From [here](https://github.com/NNPDF/eko/actions/runs/15186213930/job/42707150022#step:8:17) you see it wants to do the right thing, but this is then not correctly written to the source code. Please check how to tell it where to write the information to (=`src/eko/version.py`). Now that I have written that sentence and have been thinking about it: what about the other modules, meaning ekobox + ekomark? they should have the same version and we need to tell them

evagroenendijk:
> I think the remaining problem is that poetry-dynamic-versioning is not working correctly. From [here](https://github.com/NNPDF/eko/actions/runs/15186213930/job/42707150022#step:8:17) you see it wants to do the right thing, but this is then not correctly written to the source code. Please check how to tell it where to write the information to (=`src/eko/version.py`). Now that I have written that sentence and have been thinking about it: what about the other modules, meaning ekobox + ekomark? they should have the same version and we need to tell them
> 
> PS: from [here](https://github.com/NNPDF/eko/actions/runs/15186213930/job/42707150022#step:13:163) you see still the hardcoded version is present (instead of the dynamic one)

Sorry for my late reply, I was a bit busy with other things! What I don't completely understand is that when I run the unit tests on my computer (with installed eko version 0.0.0.post57+374c7f9b) everything does go correctly. Also if I make an eko with the lha benchmark, in the metadata it shows the correct version of(0.0.0-post.57+374c7f9b). So I don't really get why here it is not working correctly

felixhekhorn:
> regarding ekobox and ekomark, do you want to also define some version.py file in the src there?

the answer is: yes, please (they don't need there own `__data_version__`)

Issue #441, opened 2025-02-04
Revert `ev_op_max_order` to integer


felixhekhorn:
- during the developments for QED `ev_op_max_order` has become a tuple `(QCD, QED)`, because we thought we needed it
- recall that `ev_op_max_order` is the maximum perturbative order in which we expand the $U$ matrices for "perturbative-exact" - see also https://eko.readthedocs.io/en/latest/theory/DGLAP.html#id7
- however, later we found out that there is no "perturbative-exact" solution for QED, but only the iterated one - see QED paper

Question: do we want to change `ev_op_max_order` back to just an integer? @tgiani @giacomomagni 

This would be of course a breaking change (and thus directly related to #417) and to make it even more explicit we could also rename it to `ev_op_max_order_qcd`

giacomomagni:
Since we need to do a compatibility layout in any case we can add this additional conversion...

scarlehoff:
Just to say, the mismatch of `ev_op_max_order` being an integer here:

https://github.com/NNPDF/eko/blob/c08fc01ff72b167cb024122e8469d9e5d27d2d3b/src/eko/evolution_operator/__init__.py#L372

but being passed here:

https://github.com/NNPDF/eko/blob/c08fc01ff72b167cb024122e8469d9e5d27d2d3b/src/eko/evolution_operator/__init__.py#L415

to a function that understand it as a tuple:

https://github.com/NNPDF/eko/blob/c08fc01ff72b167cb024122e8469d9e5d27d2d3b/src/eko/kernels/singlet.py#L584

it's making the tests fail in my computer for python 3.13

From what I can see numba is constructing a path by which it arrives with an int to a place where it is used as a tuple:

```
E           There are 22 candidate implementations:
E                - Of which 22 did not match due to:
E                Overload of function 'getitem': File: <numerous>: Line N/A.
E                  With argument(s): '(int64, int64)':
E                 No match.
EE
E           File "../src/eko/kernels/singlet.py", line 371:
E           def r_vec(gamma_singlet, beta, ev_op_max_order, order, is_exact):
E               <source elided>
E               r = np.zeros(
E                   (ev_op_max_order[0] + 1, 2, 2), dtype=np.complex128

```

felixhekhorn:
I wouldn't rely on the documentation, that might well be outdated - instead, one would needs to look at the code.

[...]

Issue #438, opened 2025-01-29
Support Python 3.13


felixhekhorn:
preliminary feedback from @scarlehoff 
> fwiw, also eko seems to do fine with the release candidate of numba and numpy 2 (I only had to search and replace np.float_ to np.float64, probably for the full eko to work there will be some more changes)

- [ ] `eko/src/eko/io/raw.py:65: DeprecationWarning: Python 3.14 will, by default, filter extracted tar archives and reject files or modify their metadata. Use the filter argument to control this behavior` ; see also https://docs.python.org/3/library/tarfile.html#tarfile-extraction-filter

scarlehoff:
Regarding that error, not only is a deprecation warning but can be a serious security issue. At some point in NNPDF one could upload an evil tar theory / fit to the server and break the computer of whoever downloaded it 🙅

felixhekhorn:
of that risk some random bot made us already aware: https://github.com/NNPDF/eko/commit/9f0c6cf3b7b4e26352708c6ddf21085c826aa7de - but we can now abandon that hack if I understood correctly

alecandido:
> of that risk some random bot made us already aware: [9f0c6cf](https://github.com/NNPDF/eko/commit/9f0c6cf3b7b4e26352708c6ddf21085c826aa7de) - but we can now abandon that hack if I understood correctly

Now = when you will drop support for py<3.12.

If you keep following [Python's schedule](https://devguide.python.org/versions/), that means `2027-10`.

scarlehoff:
fwiw this is a relatively cheap workaround to have it working for before <3.12 https://github.com/NNPDF/nnpdf/blob/8327d16719c638ef0fd2eb6af788d9a7836146d3/validphys2/src/validphys/loader.py#L911

(I don't think we need to worry about north korean hackers messing up with eko anyway, I mentioned it more as a curiosity)

alecandido:
That's exactly why I mentioned py<3.12:

> Changed in version 3.12: Added the `filter` parameter.
> https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.extractall

(your URL is pointing to an exception raised by a filter)
```py
            try:
                res_tar.extractall(path=dest_path, filter="data")
            except tarfile.LinkOutsideDestinationError as e:
                if sys.version_info > (3, 11):
                    raise e
```

felixhekhorn:
[...]

PR #470, opened 2025-06-12
Python 3.13 & numpy 2.0 support

scarlehoff:
The only changes necessary for numpy 2 were to change from `np.<type>_` to `np.<type><size>`

everything else (meaning, the tests) seem to work ootb. I can evolve fits and benchmarks seem to do ok.

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -393,10 +393,26 @@ description = "Composable command line interface toolkit"
 optional = true
 python-versions = ">=3.7"
 groups = ["main"]
-markers = "extra == \"mark\" or extra == \"box\""
+markers = "python_version < \"3.11\" and (extra == \"mark\" or extra == \"box\")"
 files = [
     {file = "click-8.1.8-py3-none-any.whl", hash = "sha256:63c132bbbed01578a06712a2d1f497bb62d9c1c0d329b7903a866228027263b2"},
     {file = "click-8.1.8.tar.gz", hash = "sha256:ed53c9d8990d83c2a27deae68e4ee337473f6330c040a31d4225c9574d16096a"},
+]
+
+[package.dependencies]
+colorama = {version = "*", markers = "platform_system == \"Windows\""}
+
+[[package]]
+name = "click"
+version = "8.2.1"
+description = "Composable command line interface toolkit"
+optional = true
+python-versions = ">=3.10"
+groups = ["main"]
+markers = "python_version >= \"3.11\" and (extra == \"mark\" or extra == \"box\")"
+files = [
+    {file = "click-8.2.1-py3-none-any.whl", hash = "sha256:61a3265b914e850b85317d0b3109c7f8cd35a670f963866005d6ef1d5175a12b"},
+    {file = "click-8.2.1.tar.gz", hash = "sha256:27c491cc05d968d271d5a1db13e3b5a184636d9d930f148c50b038f0d0646202"},
 ]
 
 [package.dependencies]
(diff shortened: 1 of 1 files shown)


scarlehoff:
Well, not very useful since banana doesn't work in 3.13...

scarlehoff:
wait what

felixhekhorn:
We should also address #438 here

felixhekhorn:
Similar story as in https://github.com/NNPDF/banana/pull/78: it wants [`scipy==1.13.1`](https://github.com/NNPDF/eko/blob/8e1a1ade94708e5f556e3e20725d33d867d0cc68/poetry.lock#L2817) (compatible with [banana](https://github.com/NNPDF/banana/blob/38a27a95a77732e9118f88eaac0fd7070e5c7a10/poetry.lock#L1750)), but only [>1.14.1 supports 3.13](https://scipy.org/news/#scipy-1141-released)

scarlehoff:
I don't know how come locally the code found a path to get an int into the `ev_op_max_order` https://github.com/NNPDF/eko/issues/441#issuecomment-2966210855 I wonder whether there was some cached numba function? I'm quite curious but it doesn't really matter.

What worries / annoys me a bit is that the solution to both this problem and banana's was to run `poetry lock` without python 3.9 in the toml (so it gets a newer scipy), then running it again with it (then it keeps the newer scipy for 3.13).
Otherwise it tries to use the old scipy for 3.13, which it doesn't find so it tries to build it -> openblas failure.

felixhekhorn:
> What worries / annoys me a bit is that the solution to both this problem and banana's was to run `poetry lock` without python 3.9 in the toml (so it gets a newer scipy), then running it again with it (then it keeps the newer scipy for 3.13).
> Otherwise it tries to use the old scipy for 3.13, which it doesn't find so it tries to build it -> openblas failure.

but this sounds like a poetry bug, no?

src/eko/version.py
felixhekhorn, line 1: @evagroenendijk is this correct? i.e. is it correct to have such a thing commited?

pyproject.toml
```diff
@@ -40,15 +40,16 @@ packages = [
...
 scipy = "^1.10.1"
 PyYAML = "^6.0"
 lz4 = "^4.0.2"
-numba = "^0.59.0"
+# 0.60.0 is the last version compatible with 3.9 and 0.61 the first with 3.13
+numba = "~0.60.0 || ~0.61.0"
 packaging = "^24.1"
 # ekomark
-banana-hep = { version = "^0.6.12", optional = true }
+banana-hep = { git = "https://github.com/NNPDF/banana", branch = "relax_restrictions", optional = true}
```
felixhekhorn, line 52:
```suggestion
banana-hep = { version = "^0.6.16", optional = true}
```
+ `poetry lock` of course

felixhekhorn (changes requested):
should we postpone the fix for tar to later?

src/eko/version.py
scarlehoff: I guess I should revert it back to 0.0.0 and add it to .gitignore

scarlehoff:
> should we postpone the fix for tar to later?

I think the tar fix is more a curiosity than anything elsee at this point.

Question, is the master branch ready for a 0.15.2 release or should I back port this to the commit from 0.15.1 so that other things merged to master are not included?

(the importance is a bit academic, tensorflow/pytorch only work for python 3.13 in their developement versions so I still need to install those two by hand, might as well do so with eko, but I don't mind backporting this, it should be painless)

felixhekhorn:
> Question, is the master branch ready for a 0.15.2 release or should I back port this to the commit from 0.15.1 so that other things merged to master are not included?

I think we are ready. The release procedure is now similar to PineAPPL, but still manual and not automated: 1) Adjust the Changelog (make a new version there) 2) make that commit the new version

scarlehoff:
Could this be merged before it driftes away?

felixhekhorn:
> I think it is better if you do the changelog.

it wasn't that complicated https://github.com/NNPDF/eko/pull/470/commits/5fbe230d8cb8b5a43c1d389896036ee4e3243937 :innocent: 

> Could this be merged before master drifts away?

I'll do as soon as the unit tests have passed

src/eko/version.py
felixhekhorn: delayed to #475

Issue #397, opened 2024-08-06
Add LHA aN3LO to ekomark


felixhekhorn:
Let's add the tables from `extras/lh_bench_23/tables` = https://arxiv.org/abs/2406.16188 to our official benchmarks, i.e. `src/ekomark/benchmark/external/LHA_utils.py` and eventually to the CI.

This would be helpful for #383

giacomomagni:
@felixhekhorn, I'm not sure if we really want to address this. 
The values of those tables are obtained with some splittings that are right now superseded (both Pgg and Pgq have improved) and Vogt&co. are now working on other updates....  
I understand you might want to get how different is the current evolution wrt to the old one, but this will not tell you much more.

PR #380, opened 2024-05-23
Add new P_gq parametrisation

giacomomagni:
Address #376

PR #354, opened 2024-03-07
N3LO benchmark study

giacomomagni:
This PR is to keep track of the scripts needed for the N3LO benchmark study.

PR #484, opened 2025-12-18
Update N3LO (2512.10783)

giacomomagni:
This PR aims to update the N3LO splitting function to include the results from: https://arxiv.org/abs/2512.10783


crates/ekore/src/anomalous_dimensions/unpolarized/spacelike/as4/ggq.rs
```diff
@@ -172,15 +197,15 @@ mod tests {
...
+            for NF in [3, 4, 5, 6] {
                 for N in [2.0, 4.0, 6.0, 8.0, 10., 12., 14., 16., 18., 20.] {
                     let mut c = Cache::new(cmplx!(N, 0.));
                     let test_value = gamma_gq(&mut c, NF, variation);
                     assert_approx_eq_cmplx!(
                         f64,
                         test_value,
                         cmplx!(gq3_moment(N as usize, NF as f64), 0.),
-                        rel = 4e-4
+                        rel = if NF != 6 { 4e-4 } else { 2e-3 }
```
felixhekhorn: both `gamma_gq` and `gq3_moment` are coming from Falcioni et al., right? should we tell them that they got worse, i.e. they don't follow their own boundary condition? I know we are talking 1e4 here, so it is stupid - it may still indicate a possible typo ...

giacomomagni: eh yes, I saw your email, but it could well be that the typo is from my side!! I'll try to double check the changes nex week.

felixhekhorn: Mail from Sven Moch on 08.01.26, 17:28 EET says "don't worry"

felixhekhorn:
> I haven't checked the impact of the changes on P g q at pheno level yet.

I think running the N3LO benchmark tables might be sufficient

felixhekhorn:
> `apply_pdf_flavor` seems not working anymore... Do you have idea how to fix this ?

I haven't tried the fix in 8d5f7068 (since I'm too lazy to run the script, i.e. I'm worried of the runtime :see_no_evil: ), but can you please try that? (It's most likely an unaccounted side effect of 4ff9381)

> Another way to solve this is address #397 by adding the paper table to the yaml file.

agreed

giacomomagni:
> do I understand correctly this is up to a ~4% correction? this would be quite significant ... i.e. PDF change by 4%? actually this can not be, N3LO should be of that order, no?

So this is the current version of eko vs the paper table.
To check the impact of this specific update I need to run against master, but I'd prefer to do it after #489.  
If I recall correctly #414, #380 were merged after the benchmark paper #354, and this could explain the magnitude of the change.  Also this 3.5% difference is at $x=10^{-7}$, while for $x > 10^{-4}$ everything seems in good order.

Issue #518, opened 2026-05-15
Improve interface Python - Rust


felixhekhorn:
- Identify weak points and improvements between numba and Rust
- Identify weak points and improvements between Python and numba and Rust
- Identify weak points and improvements between Rust/eko and Rust/ekore
- Implement improvements
- Update architecture documentation #516 

Things to consider:
- build system
- passing numba function pointer around #526 
- memory allocation and unravelling
- caching strategy
- more optimisation closer to hardware

felixhekhorn:
@AkshatRai07
I suggest to have a copy of the existing [quad_ker_qcd](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L329) where I replace these lines (and only these lines)
https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L390-L401
with a single call to the eko crate `gamma_singlet = ekors.rust_quad_ker(is_polarized, order, blub)`.

In the first iteration `ekors.rust_quad_ker` will be: 1) call `ekore.blub` and 2) return gamma from there. So no more calling python inside Rust, but just returning the ekore computation; also [SV](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L403) or [exponentiation](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L407) remain identical to the current Python implementation.

In a second step we can move as a next candidate SV from Python to Rust (while leave exp untouched). This way we can shift pieces slowly one at a time.

felixhekhorn:
Step 1:
-  split `ev_op/init.py` into two files
- `ev_op/quad_ker.py` will hold all `quad_ker*` related code (=functions); current content can be dropped; this will leave the (whole) library unchanged but for a few (single?) new import(s) in `ev_op/init.py`
- the new Rust logic flow will then patch `quad_ker.py` (possibly instead of current `ev_op/init.py`)
- the definition of the integration path will remain in Python for the moment and will be deactivated for now on the Rust side
- decide something for OME

this will remove the passing of the Python function pointer

AkshatRai07:
@felixhekhorn I debugged as to why this new code in #526 is slower than the original one.
[...]

Issue #519, opened 2026-05-15
Add interfaces to rust/ekore


felixhekhorn:
- Add C/C++ interface
- Test C/C++ interface in CI
- Document C/C++ interface
- Repeat for Python
- Repeat for Fortran

We define a part of our public API #145 here. The design rationale is to expose the minimal amount in the beginning and have that synchonized across all languages

felixhekhorn:
Question: do we want to expose a wrapper for
https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L390-L401
or are we fine with just exposing directly [ad.u.s.gamma_ns_qcd](https://github.com/NNPDF/eko/blob/ba40d2be721b647b82259bb998c69bc7feedd1dd/crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs#L22) and friends and leave the user with the if branches?

felixhekhorn:
we decided against #435

felixhekhorn:
we decided against #500

felixhekhorn:
> Repeat for Fortran

postponed to #566

Issue #516, opened 2026-05-15
Document current architecture


felixhekhorn:
- Define format
- make sure it can be updated later on
- it might make sense to define one document for the Python flow and one for the Rust flow?
- Generated document can be stored in the `extras` folder

The document should at least mention the following steps:
- build system: poetry vs. maturin
- [User entry point](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/runner/managed.py#L22)
- split in [Operator](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L622) and [OperatorMatrixElement](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/operator_matrix_element.py#L205) ?
- [Loop on target interpolation point](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L993) which can be parallelized
- [Loop on source interpolation node](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L896) and [loop over flavors](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L901)
- [scipy.intergate.quad](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L902)
- [quad_ker](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/src/eko/evolution_operator/__init__.py#L195) numba entry point in Python workflow vs. [rust_quad_ker](https://github.com/NNPDF/eko/blob/0287086a437c0d45dca8931666f561602ccd3ba2/crates/eko/src/lib.rs#L82) entry point in Rust workflow together with the [return point to numba](https://github.com/NNPDF/eko/blob/master/src/eko/evolution_operator/quad_ker.py)
[...]

Issue #136, opened 2022-07-11
HELL Interface


felixhekhorn:
see https://www.ge.infn.it/~bonvini/hell/

felixhekhorn:
Moved to https://github.com/NNPDF/eko/wiki/List-of-Possible-Future-Features

AleCandido:
About this there was something ongoing between @giacomomagni and Marco Bonvini.

Are you sure you want to close?

giacomomagni:
no way in the short term. We cal always reopen

PR #526, opened 2026-06-03
Split ev_op/init

AkshatRai07:
a step for #518 

- split `src/eko/evolution_operator/__init__.py` into two parts: the new part (`quad_ker.py`) containing the integration kernels only and the remaining part containing only the class
- this leaves the Python workflow untouched, but pushes the [newly proposed Rust workflow](https://github.com/NNPDF/eko/issues/518#issuecomment-4534168736)
- this will (hopefully, to be confirmed) only make it necessary for the Rust-activation-hack to patch this new part
- the new strategy will address:
   - make Rust independent from Python/Numba (only the reverse dependency remains)
   - in particular: no more "passing numba function pointer around"
   - Rust will only get the variables it (currently) needs (once Rust controls the full integration kernel in the future it will of course receive again everything)

---
(EDIT: this would have been better placed in a comment, since it is not describing the PR but rather one specific problem)

@felixhekhorn @scarlehoff I have one small issue with this and hence I'd like to now if you are facing this or not. So after changing all the files I tried to run poe test and got this error:
```text
Poe => pytest tests
ImportError while loading conftest '/home/akshatrai/Documents/GSoC/eko/tests/conftest.py'.
tests/conftest.py:11: in <module>
    from eko import interpolation
src/eko/__init__.py:6: in <module>
    from .runner import solve
src/eko/runner/__init__.py:4: in <module>
    from .managed import solve
src/eko/runner/managed.py:19: in <module>
    from . import operators, parts, recipes
src/eko/runner/parts.py:15: in <module>
    from .. import evolution_operator as evop
src/eko/evolution_operator/__init__.py:25: in <module>
    from .quad_ker import quad_ker
src/eko/evolution_operator/quad_ker.py:631: in <module>
    @nb.cfunc(
.venv/lib/python3.11/site-packages/numba/core/decorators.py:275: in wrapper
    res.compile()
.venv/lib/python3.11/site-packages/numba/core/compiler_lock.py:35: in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/ccallback.py:68: in compile
    cres = self._compile_uncached()
           ^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/ccallback.py:82: in _compile_uncached
    return self._compiler.compile(sig.args, sig.return_type)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/dispatcher.py:84: in compile
    raise retval
.venv/lib/python3.11/site-packages/numba/core/dispatcher.py:94: in _compile_cached
    retval = self._compile_core(args, return_type)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/dispatcher.py:107: in _compile_core
    cres = compiler.compile_extra(self.targetdescr.typing_context,
.venv/lib/python3.11/site-packages/numba/core/compiler.py:739: in compile_extra
    return pipeline.compile_extra(func)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/compiler.py:439: in compile_extra
    return self._compile_bytecode()
           ^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/compiler.py:505: in _compile_bytecode
    return self._compile_core()
           ^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/compiler.py:484: in _compile_core
    raise e
.venv/lib/python3.11/site-packages/numba/core/compiler.py:473: in _compile_core
    pm.run(self.state)
.venv/lib/python3.11/site-packages/numba/core/compiler_machinery.py:367: in run
    raise patched_exception
.venv/lib/python3.11/site-packages/numba/core/compiler_machinery.py:356: in run
    self._runPass(idx, pass_inst, state)
.venv/lib/python3.11/site-packages/numba/core/compiler_lock.py:35: in _acquire_compile_lock
    return func(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/compiler_machinery.py:311: in _runPass
    mutated |= check(pss.run_pass, internal_state)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/compiler_machinery.py:272: in check
    mangled = func(compiler_state)
              ^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/typed_passes.py:112: in run_pass
    typemap, return_type, calltypes, errs = type_inference_stage(
.venv/lib/python3.11/site-packages/numba/core/typed_passes.py:91: in type_inference_stage
    infer.build_constraint()
.venv/lib/python3.11/site-packages/numba/core/typeinfer.py:1027: in build_constraint
    self.constrain_statement(inst)
.venv/lib/python3.11/site-packages/numba/core/typeinfer.py:1389: in constrain_statement
    self.typeof_assign(inst)
.venv/lib/python3.11/site-packages/numba/core/typeinfer.py:1464: in typeof_assign
    self.typeof_global(inst, inst.target, value)
.venv/lib/python3.11/site-packages/numba/core/typeinfer.py:1564: in typeof_global
    typ = self.resolve_value_type(inst, gvar.value)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
.venv/lib/python3.11/site-packages/numba/core/typeinfer.py:1485: in resolve_value_type
    raise TypingError(msg, loc=inst.loc)
E   numba.core.errors.TypingError: Failed in nopython mode pipeline (step: nopython frontend)
E   Untyped global name 'select_singlet_element': Cannot determine Numba type of <class 'function'>
E
E   File "src/eko/evolution_operator/quad_ker.py", line 705:
E   def cb_quad_ker_qcd(
E       <source elided>
E               ) @ np.ascontiguousarray(ker)
E           ker = select_singlet_element(ker, mode0, mode1)
E           ^
E
E   During: Pass nopython_type_inference
```
In short, the function `select_singlet_element` is of type `function`, and not whatever numba wants it to be (i.e. njit function). I have no idea why this is taking place since the function is placed above in the file, and the structure of the file before was the same only but it did not encounter this error. I ran this and the error got fixed:
```bash
find . -type d -name "__pycache__" -exec rm -rf {} +
.venv/bin/python -c "import sys; sys.path.insert(0, 'src'); import eko.evolution_operator.quad_ker"
```
So remove all the cache, and then call only the quad_ker file such that it caches without any problems. Then running `poe test` does not result in any error.

I had faced this error before. After running rustify.sh, poe test was returning the same error. I imported all the dependencies of quad_ker.py in terminal (similar way as shown above) and the error was resolved.

I have no clue why on Earth numba is showing this behaviour, but we should not split the file if the error persists.


AkshatRai07:
So apparently because of the three `cb_quad_ker_qcd/qed/ome` functions (which are cfunc for the Rust callback) all of this is happening (since they are not disabled by `D:NUMBA_DISABLE_JIT=1`). No clue why this was not happening before on master but is happening now. This will be solved after we fully convert to Rust.

Unfortunately there is no dedicated `NUMBA_DISABLE_CFUNC`. One thing which we can do is **replace** cfunc at the time of testing by adding this at the top of the file:
```python
import os

if os.environ.get("NUMBA_DISABLE_JIT", "0") == "1":

    def cfunc_wrapper(sig, *args, **kwargs):
        def decorator(func):
            return func

        return decorator

    nb.cfunc = cfunc_wrapper
```
Please do let me know if you are satisfied with this or not so that I can start fixing patch files for `LHA Benchmarks (Rust)` to run successfully.

felixhekhorn:
> No clue why this was not happening before on master but is happening now. This will be solved after we fully convert to Rust.

Actually, I would say this is a feature and not a bug.
1. the old layout:
   - recall that `@cfunc` compiles a C function - which is completely different from a Python function. In particular, while Rust can execute C functions (easily), it can't with Python functions (because they are far more complex due the language). So when we pass a function to Rust, it must be a C function, always - even within testing. Deactivating this feature will break the code.
   - the main reason why we deactivate Numba in testing is for debugging reasons; i.e. with Numba it is impossible to properly debug, so we just deactivate it. However, this should not apply to `@cfunc` due to the above reason.
2. the new layout:
   - I suggest to immediately push for the [new layout](https://github.com/NNPDF/eko/issues/518#issuecomment-4534168736) - and there we would not need any `@cfunc` any longer, since we do not pass Python functions around any longer
   - in order to do so, you can drop all the current content of `quad_ker.py` - yes, this will break Rust for a moment, but then we fix it in a completely new way: with the new layout

Do you agree? (and to be explicit: Copilot is telling non-sense in this case I would say)

Then there is the unrelated matter of Numba+caching: this yields sometimes strange behaviour since the caching is not always consistent. See also https://github.com/numba/numba/issues/8926 . Thus, the first action in debugging Numba is always to delete the (Python) cache (files) and let Numba recompile; typically it is sufficient to delete "the relevant cache" (which you need to guess) and not all, which for us can take a very long time to recompile.

AkshatRai07:
In summary, today, I removed `cfunc`s, and I changed the eko crate binding from `cffi` to `pyo3` (which led to changes in root `Cargo.lock`). Took me some time to figure out it's pretty hard to call a `cffi` function from an `nb.njit` function. Also, I knew that njit cannot compile PyO3 object from the GSoC task I did. So the way out of this (for now) is to call such functions from inside objmode. In this setting, numba drops back into Python Interpreter, execute whatever is in it, and then return the result to nopython code. As we expand more and more of the library, we will entirely drop numba, and hence this workaround.

I also changed poe compile to use maturin, added that as a dev-dependency. This led to changes in `poetry.lock` and `lha_bot_rust.yaml`.

crates/eko/src/lib.rs
```diff
@@ -74,6 +79,64 @@ fn unravel_qed_ns(res: Vec<Vec<Complex<f64>>>, order_qcd: usize, order_qed: usiz
     target
 }
 
+/// Type for gamma singlet return
+type GammaSinglet = PyResult<(Py<PyArray1<f64>>, Py<PyArray1<f64>>)>;
```
felixhekhorn:
```suggestion
type GammaQCDSinglet = PyResult<(Py<PyArray1<f64>>, Py<PyArray1<f64>>)>;
```
let's be specific and anticipate the generalization that we will need to do for the (QCDx)QED case

crates/eko/src/lib.rs
```diff
@@ -74,6 +79,64 @@ fn unravel_qed_ns(res: Vec<Vec<Complex<f64>>>, order_qcd: usize, order_qed: usiz
...
+#[pyfunction]
+#[allow(clippy::too_many_arguments)]
+pub fn qcd_gamma_singlet(
+    py: Python<'_>,
+    order_qcd: usize,
+    re_n: f64,
+    im_n: f64,
+    nf: u8,
+    is_polarized: bool,
+    is_time_like: bool,
```
felixhekhorn: I would move those two arguments up as they are the most "general" in some sense ... they are consumed first

crates/eko/src/lib.rs
felixhekhorn: this can be dropped, right?

crates/eko/src/lib.rs
```diff
@@ -74,6 +79,64 @@ fn unravel_qed_ns(res: Vec<Vec<Complex<f64>>>, order_qcd: usize, order_qed: usiz
...
+    ))
+}
+
+/// Python extension module exposing EKO's Rust kernels.
+///
+/// Currently exposes:
+/// - [`qcd_gamma_singlet`]: QCD singlet anomalous dimension matrix in Mellin space
+#[pymodule]
+fn ekors(_py: Python<'_>, m: &Bound<'_, PyModule>) -> PyResult<()> {
+    m.add_function(wrap_pyfunction!(qcd_gamma_singlet, m)?)?;
```
felixhekhorn, line 282: If this works, please also add the non-singlet counterpart

crates/eko/Cargo.toml
```diff
@@ -22,3 +22,5 @@ crate-type = ["cdylib"]
 [dependencies]
 num = "0.4.1"
 ekore = { path = "../ekore", version = "0.0.1" }
+pyo3  = { version = "0.21", features = ["extension-module"] }
```
felixhekhorn, line 25: Can you please briefly explain to me what the `extension-module` does?

src/eko/evolution_operator/operator_matrix_element.py
felixhekhorn: Let's move this function (and it's dependency = the upper) to `quad_ker.py` as well, so we have all functions in one place

src/eko/evolution_operator/operator_matrix_element.py.patch
felixhekhorn, line 1: of course we need to counteract this deletion (which is the right thing), with the corresponding new patch

crates/eko/src/lib.rs
```diff
@@ -74,6 +79,64 @@ fn unravel_qed_ns(res: Vec<Vec<Complex<f64>>>, order_qcd: usize, order_qed: usiz
...
+    py: Python<'_>,
+    order_qcd: usize,
+    re_n: f64,
+    im_n: f64,
+    nf: u8,
+    is_polarized: bool,
+    is_time_like: bool,
+    variations: [u8; 4],
+) -> GammaSinglet {
+    if is_polarized && is_time_like {
```
felixhekhorn, line 110: Let's add a TODO here; something like
```rust
// TODO: once the Rust kernels grows we take be the computation of the integration path back to here
// in the old implementation it was ~
// // prepare Mellin stuff
// let path = mellin::TalbotPath::new(u, args.logx, is_singlet);
// let jac = path.jac() * path.prefactor();
```

src/eko/evolution_operator/quad_ker.py.patch
```diff
@@ -0,0 +1,45 @@
...
+-            if is_time_like:
+-                gamma_singlet = ad_ut.gamma_singlet(order, ker_base.n, nf)
+-            else:
+-                gamma_singlet = ad_us.gamma_singlet(
+-                    order, ker_base.n, nf, n3lo_ad_variation, use_fhmruvv
+-                )
++        order_qcd = order[0]
++        n_re = ker_base.n.real
++        n_im = ker_base.n.imag
++        with nb.objmode(gamma_singlet="complex128[:,:,:]"):
```
felixhekhorn, line 41: I guess the problem is not so much calling a foreign function in numba, but rather a PyO3 object ...

For the context of #136 (which will introduce another C library into the mess of anomalous dimensions), I was experimenting with how to do that and for plain C functions it seems to be straight:

```python
from numba import njit
import ctypes
DSO = ctypes.CDLL('./shared_library.so')

# Add typing information
c_func = DSO.sum
c_func.restype = ctypes.c_int
c_func.argtypes = [ctypes.c_int, ctypes.c_int]

@njit
def example(x, y):
   return c_func(x, y)

print(example(3, 4)) # 7
```
```C
// gcc lib.c -fPIC -shared -o shared_library.so
#include <stdint.h>

int64_t sum(int64_t a, int64_t b){
   return a + b;
}
```
here we don't have a direct C library in hand, so it does not really help us, right?

well, actually it shows another alternative to PyO3: we could add a C interface also to crates/eko (remember that the target of #519 is crates/ekore) and use that under scipy ... in the end we don't care so much crates/eko being a Python library since it is purely internal (since it is just a glue) ....

any thoughts/opinions?

what is the easiest way? which we can grow smoothly? if we ever would want to change strategy, we most likely want to do so in a next step, provided `objmode` works ...

.github/workflows/lha_bot_rust.yml
```diff
@@ -41,6 +41,8 @@ jobs:
       - name: Install task runner
         run: pip install poethepoet
       - name: Run benchmark
+        env:
+          VIRTUAL_ENV: ${{ env.pythonLocation }}
```
felixhekhorn, line 45: why is that needed?

felixhekhorn (changes requested):

crates/eko/Cargo.toml
AkshatRai07: So when I first did `poe compile` with `maturin`, it gave me this warning:
```
⚠️  Warning: You're building a library without activating pyo3's `extension-module` feature. See https://pyo3.rs/v0.21.2/building-and-distribution.html#the-extension-module-feature
```
Without `extension-module`, a pyo3 package links its own `libpython` into the `.so`. But when Python imports our package, there's already a libpython running, the Python interpreter itself. Having two copies of libpython in the same process causes conflicts. extension-module disables this. It tells the package to just use the libpython that's already running instead of bundling its own.

The downside is that `cargo test` will break for any pyo3 functions, since the test binary runs standalone and has no libpython to rely on. This shouldn't be an issue for now as we don't plan to write tests for these.

Also, as per the docs linked in the warning:
```
Python extensions on Unix must not link to libpython for manylinux compliance.
```
So extension-module is a necessity here.

Do tell me if you are satisfied by reacting with an emoji so that I can resolve the conversation :)

.github/workflows/lha_bot_rust.yml
AkshatRai07: `poe compile` did not work [here](https://github.com/NNPDF/eko/actions/runs/26970428564/job/79583944556) because `maturin` was not able to identify the virtual environment. This was the root cause told by the error:
```
For maturin to find your virtualenv you need to either set VIRTUAL_ENV or have a virtualenv called .venv
```
GitHub Actions' `setup-python` installs Python at `pythonLocation` (i.e. /opt/hostedtoolcache/Python/3.12.13/x64) but does not set up a virtual environment. So I pointed `VIRTUAL_ENV` to `pythonLocation`, which is already set by setup-python, so maturin can find the correct Python to build against.

Do tell me if you are satisfied by reacting with an emoji so that I can resolve the conversation :)

src/eko/evolution_operator/quad_ker.py.patch
AkshatRai07: The issue with using ctypes was passing tuples, lists, and arrays. There was some sort of type mismatch between python and Rust. In the end I just gave up. I thought to myself that all of this is just temporary, let us make a working solution, because in the future we will remove the `#[pyfunction]` from `qcd_quad_ker`, and add that macro to the `quad_ker` function. Ultimately we got to remove this `#[pyfunction]` entirely and directly pass a single C-ABI function to `scipy.integrate.quad` with `LowLevelCallable`.

For #519, I thought you wanted to make that crate to expose the `ekore` crate to other C/C++ projects and not the python one. The quad function requires almost half of the `src/eko` python code to be a part of it's hot loop. We can try to add them in the `eko_capi`, but I don't think we should should change the whole structure of the crate for just one function. Otherwise, we will have call that one function from `eko` crate, expose it as C-ABI in `eko_capi`, this will bloat the `eko_capi` crate as now `eko` crate will be required for the compilation. After that use the `.so` file for just one function while it exposes other functions too, again feels bloated to me.

It's better to stick to what we are currently doing.

src/eko/evolution_operator/operator_matrix_element.py.patch
AkshatRai07: Please do let me know whether I should replace this:
https://github.com/NNPDF/eko/blob/d247da307b63fe715d7b2bf2e07d7abd1028599e/src/eko/evolution_operator/quad_ker.py#L721-L742
with ekors in this PR only.

AkshatRai07:
2 doubts.
Should we keep the pyproject.toml.patch? I think it has no effect on our development, publishing the eko-rs package, and running `LHA Benchmarks (Rust)`. I think in a meet we discussed that we should keep poetry as the main build backend, and maturin as the builder for eko-rs.
Second, during `poe test` I received some deprecation warnings. Should I fix them now or open a separate PR after this is closed?

src/eko/evolution_operator/operator_matrix_element.py.patch
felixhekhorn: yes, please - this will replace
https://github.com/NNPDF/eko/blob/2c49156022235da1bc3582b4b61bd323e42d8d7f/src/eko/evolution_operator/operator_matrix_element.py.patch#L189

src/eko/evolution_operator/quad_ker.py.patch
felixhekhorn: Re #519: actually we also want to expose `ekore` to Python; e.g. I'm using it [here](https://github.com/felixhekhorn/geko/blob/c88357256a57cde5820f15d48e1fbd9faeaba9d8/src/geko/anomalous_dimensions.py#L18). However, since we kind of know how to do that I didn't put it on the list at first - now, I've changed my mind and added it explicitly.

Re `eko_capi`: let's discuss in person. If lists are a problem that's an issue, which maybe we have to solve when coming to #519 ... although I can imagine it should be possible, if we just pass numpy objects around ... On the other side I'm happy to expose a single C function if that saves us a lot of time.

Anyway, the conclusion is very likely: for this PR we keep the currently proposed solution

felixhekhorn:
> Should we keep the pyproject.toml.patch?

if it doesn't do anything for us, maybe we can drop it ...

> Second, during `poe test` I received some deprecation warnings. Should I fix them now or open a separate PR after this is closed?

is there something Rust related? We know we have some Python deprecations flying around for which we haven't had the time so far to address them. I'd say we should rather focus on the Rust side ...

crates/eko/src/lib.rs
```diff
@@ -74,334 +81,213 @@ fn unravel_qed_ns(res: Vec<Vec<Complex<f64>>>, order_qcd: usize, order_qed: usiz
...
 
-/// Integration kernel inside quad.
-///
-/// # Safety
-/// This is the connection from Python, so we don't know what is on the other side.
-#[no_mangle]
-pub unsafe extern "C" fn rust_quad_ker(u: f64, rargs: *mut c_void) -> f64 {
-    let args = *(rargs as *mut QuadArgs);
+/// Type for qcd gamma singlet return
+type GammaQCDSinglet = PyResult<(Py<PyArray1<f64>>, Py<PyArray1<f64>>)>;
```
felixhekhorn:
```suggestion
/// Type for list of complex numbers (real + imaginary part)
type ComplexResult = PyResult<(Py<PyArray1<f64>>, Py<PyArray1<f64>>)>;
```
no need to define three times the same thing (i.e. drop the other two)

crates/eko/src/lib.rs
```diff
@@ -74,334 +81,213 @@ fn unravel_qed_ns(res: Vec<Vec<Complex<f64>>>, order_qcd: usize, order_qed: usiz
...
-                    args.mode0,
-                    &mut c,
-                    args.nf,
-                    n3lo_ad_variation[4..7].try_into().unwrap(),
-                ),
-                args.order_qcd,
-                args.order_qed,
-            );
-        }
+// TODO: once the Rust kernels grows we take be the computation of the integration path back to here
```
felixhekhorn:
```suggestion
// TODO: once the Rust kernels grow we take the computation of the integration path back to here
```

felixhekhorn (changes requested):
- I think this change is big enough to merit an entry in the Changelog
- update #516

src/eko/evolution_operator/quad_ker.py.patch
AkshatRai07: > Re #519: actually we also want to expose `ekore` to Python

Yes, in the meet you also said about another crate, which exposes ekore to python. The `geko` example you attached uses the `ekore` crate in normal functions (not njit), which can be done using PyO3.

> Re `eko_capi`: let's discuss in person. If lists are a problem that's an issue

The soltuion is to pass these lists, tuples and dictionaries as a C Structure as shown [here](https://gitlab.com/akshatrai07-gsoc/collectionpythontorust).

felixhekhorn: A solution is postponed to later

felixhekhorn:
- please remember to remove the `pyproject.toml` patch as we agreed
- did we mention anything else? which is needed here?
- I was thinking (yesterday) to merge this after #528 , but if this is ready and I can see that #528 still might need some time, I wonder if we should swap the order and merge this one first ... the two conflict by the new maturin dependency added here
- if we wait here and you want to continue and open a new PR with new changes just make this the base

AkshatRai07:
> I wonder if we should swap the order and merge this one first

Let us merge this first, but please read [this](https://github.com/NNPDF/eko/issues/518#issuecomment-4672733551) and lemme know how we should proceed.

felixhekhorn:
Oookay, mmm ... now, I think this change was a bad idea and we should give up on the idea, i.e. close the PR _without_ merging. The whole Rust business is for speed and if the current master is better then this (by a factor of 3(!)), let's not do this change.

The main cause (to me) seems to be: while Rust can deal (easily) with C functions, Numba can not deal with Rust functions. This makes sense to me, because Numba is just an attempt for a shortcut for a compiled language - it can not provide the same features (and it does not have to - if you want everything do the right thing and use a compiled language in the first place). In the end passing a C function pointer around is not such a bad idea; it is a true C function, so no Python (or Numba) in the middle.

What do you think? (If you agree please close the PR.)

felixhekhorn:
as for salvaging pieces from here: the only relevant change mixed among the other changes seems to me to add maturin as dev-dependency and use it for compilation. We can split that into a separate PR either sitting on top of #528 (or even directly there)

AkshatRai07:
The other changes included:
- Moving quad_ker (__init__ and ome) to quad_ker.py
- Adding maturin as dev dependency
- Removing pyproject.toml.patch

React with a thumbs up, I'll close this PR, implement and commit the changes on top of #528.

felixhekhorn:
> The whole Rust business is for speed

just to stress again: I don't expect short time to be able to give up on numba and have a full Rust kernel, but we want to get as soon as possible the heavy lifting into Rust

PR #527, opened 2026-06-08
add support for python 3.14

jekoorn:
bumped numba for python 3.14


src/eko/version.py
```diff
@@ -1,2 +1,2 @@
-__version__ = "0.0.0"
```
felixhekhorn, line 1: @evagroenendijk people are not running pre-commit 🙈  i.e. https://github.com/NNPDF/eko/blob/2c49156022235da1bc3582b4b61bd323e42d8d7f/.pre-commit-config.yaml#L56

pyproject.toml
```diff
@@ -40,12 +40,13 @@ packages = [
...
 [tool.poetry.dependencies]
-python = "^3.10,<3.14"
+python = "^3.10,<3.15"
 numpy = "^2"
 scipy = "^1.10.1"
 PyYAML = "^6.0"
 lz4 = "^4.0.2"
-numba = "^0.61.0"
+# 0.60.0 is the last version compatible with 3.9 and 0.63 the first with 3.14
+numba = ">=0.60,<0.66"
```
felixhekhorn: - this + the Python line + `poetry.lock` are the only real changes all others need to vanish
- should we lift the lower version? it doesn't hurt, but we also don't need to care about 3.9 any longer

felixhekhorn (changes requested):
- test 3.14 in workflows and remove 3.9
- add a Changelog entry

pyproject.toml
jekoorn: I agree that we should lift 3.9, just wanted to get some verification before I actually do so. No reason not to bump your environment to a newer python version.

pyproject.toml
```diff
@@ -80,7 +80,7 @@ pylint = "^3.1.0"
 pdbpp = "^0.11.6"
 ipython = "^8.0"
 # benchmark
-asv = "^0.4.2"
+asv = "^0.6.5"
```
felixhekhorn, line 83: you are not sitting on top of master! see also https://github.com/NNPDF/eko/commit/8b215357540687bba22e62477b2289ad209bbb6e -> maybe rebase (or given the bunch of unrelated changes restart?)

felixhekhorn (changes requested):

pyproject.toml
jekoorn: ahh oops, I thought I had rebased. I'll probably restart

PR #522, opened 2026-05-28
Correctly publish existing crates

AkshatRai07:
Solves #517

There were a few issues in maturin.yml. First, it was not calling bump-versions.py. Second, it used Maturin to publish (which is now deprecated). Both of these issues are now solved. [This](https://github.com/AkshatRai07/eko/actions/runs/26567726153) is the action that ran and [these](https://pypi.org/project/akshat-eko-rs-test/#files) are the packages it published.

For the README of crates, I have changed their Cargo.toml, and added README of only the dekoder crate, since eko and ekore crate are under development.

PR #528, opened 2026-06-08
add support for python 3.14

### LINKED ISSUE #486: Support python 3.14
👀 

(fwiw, it can be installed)

jekoorn:
reattempt of #527 

closes #486 

From #526 we import adding maturin as dev-dependency and drop the `pyproject.toml` patch


jekoorn:
It looks like the only version of scipy that's compatible with python 3.14 is incompatible with 3.10. Do we then regard 3.10 as obsolete and update it for 3.11 .. 3.14?

jekoorn:
i've scoured the workflow for any mention of python 3.10, and there are none. I am not sure what else can then be done in terms of updating the workflow. I took it that the idea was to standardize it over all workflows, but for that we still need this one to get working. Where do I change what I need to change?

felixhekhorn:
- I'd say let's use something like [this](https://docs.github.com/en/actions/how-tos/write-workflows/choose-what-workflows-do/use-variables#creating-configuration-variables-for-a-repository) - we can call the variable PYTHON_VERSION_WORKFLOWS or something similar
- please check _all_ workflows in this repository e.g. also this one https://github.com/NNPDF/eko/blob/2c49156022235da1bc3582b4b61bd323e42d8d7f/.github/workflows/python-poetry-pypi-with-data.yml#L8 and use the variable everywhere
- we need to worry about the [isolated benchmarks](https://github.com/NNPDF/eko/blob/master/.github/workflows/isobench.yml), which run through a container at the moment: either we update the container or, better, we check what do we actually need from the container, do we need anything at all, and potentially find a way to abandon it in favor of a "normal" workflow
- also remember to add a Changelog entry

jekoorn:
For changing the repo settings I will need more permissions though. 
As checking what do we actually need from the container, how can we find out exactly what it's doing?

AkshatRai07:
In isobench.yml, we can add this:
```yaml
- name: Install Python 3.11
  uses: actions/setup-python@v5
  with:
    python-version: "3.11"

- name: Use Python 3.11
  run: poetry env use 3.11
```

jekoorn:
why 3.11 and not 3.14?

AkshatRai07:
We can use 3.14 too that's not an issue, I just gave an example

jekoorn:
Gotcha thanks

felixhekhorn:
> For changing the repo settings I will need more permissions though. 

ok; I was hoping you had enough for variables ... anyway I now added
grafik

> why 3.11 and not 3.14?

That's precisely why I want to add a variable :innocent: 

> As checking what do we actually need from the container, how can we find out exactly what it's doing?

so this
https://github.com/NNPDF/eko/blob/2c49156022235da1bc3582b4b61bd323e42d8d7f/.github/workflows/isobench.yml#L11
is eventually coming from here:
https://github.com/NNPDF/workflows/blob/v2/containers/bench-evol/Containerfile

and there you can see that 3.10 is used.

I think it is the easiest if we just drop the container and then we see where we fail (if we fail) - and then we know what we need. So in addition to @AkshatRai07 s changes replace the image with the usual one (i.e. just drop it I think)

The most likely thing where we will fail is LHAPDF (as usual), but I'm not 100% sure we will and if so we need to come up with something

felixhekhorn:
also @jekoorn sorry we will slightly abuse this PR to save some simple (dependency) stuff from #526 (@AkshatRai07 the `quad_ker` business does not fit here - it is too much)

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ Items without prefix refer to a global change.
 
 ## [Unreleased](https://github.com/NNPDF/eko/compare/v0.15.3...HEAD)
 
+### Changed
+- py: Implement support for Python 3.14, dropping Python 3.10 ([#528](https://github.com/NNPDF/eko/pull/528))
+
```
felixhekhorn, line 22:
```suggestion
- rust: Make maturin a dev-dependency and use it directly for compilation ([#528](https://github.com/NNPDF/eko/pull/528))


```

jekoorn:
is this pr ready to be merged?

felixhekhorn (changes requested):
Use repo variable in all workflows

felixhekhorn (approved):

felixhekhorn:
@jekoorn you missed https://github.com/NNPDF/eko/blob/38e7f8623a02202a363c0c2feef4e8dbad104858/.github/workflows/lha_bot.yml#L21 please open a new PR with the necessary fix

felixhekhorn:
@jekoorn one more: https://github.com/NNPDF/eko/blob/38e7f8623a02202a363c0c2feef4e8dbad104858/.github/workflows/lha_bot_rust.yml#L23

felixhekhorn:
@jekoorn another problem with this PR: in https://github.com/NNPDF/eko/commit/cf7b2d8b43d62355766fe7f7a854409853341a65 you added a bunch of big wheels and while they got removed again in https://github.com/NNPDF/eko/commit/2f93013120eb3cc04c251284e6c59c38ac334007 they are now part of the version history on main and thus when someone pulls he get's all the history including these (useless) files ...

Conclusion for you: please be more careful with committing files to a repo!

Conclusion for myself: be more careful with individual commits - we could have e.g. just squashed the commits and in this way they would have at least only existed on this branch ...

jekoorn:
Ahh I didn't realize this would have this effect... Since I tested the wheel building thing quite often, it just has to go wrong once. I will be more careful in the future, sorry.

AkshatRai07:
There haven't been many commits since. @felixhekhorn you can try (if you have time) to force rewrite the git history from "Merge #522" Commit. There will be some hash changes, but since there aren't many comments yet referencing them, we can edit the comments.

There is a tool called filter-repo which does this thing, but unfortunately it changed all the commit hashes since the beginning of the repo when I tried it, which is much more disastrous.

To prevent such things in future, we can add `*.whl` in `.gitignore`, and this in pre-commit YAML:
```yaml
    hooks:
      - id: check-added-large-files
        args: ['--maxkb=1024']
```

PR #537, opened 2026-06-23
Add ekore_capi

AkshatRai07:
A step for #519 ; we explicitly shape our public API along the way #145

In this PR we:
- Add the ekore_capi crate.
- Expose anomalous_dimensions and operator_matrix_elements as C functions.

### TODO

Create and merge a PR for:
- [x] #538. #540
- [x] `--release` tag. #541 
- [x] No malloc in ekore. #542 
- [x] Single buffer passing throughout the integration. #542 
- [x] `ekore` visibility. #543 
- [x] #539 #543 
- [x] `spacelike.rs` tests. #546 
- [x] Bump Rust and add `msrv.yml` #545 
- [x] Define version of a dependency in workspace and make crates point to that. #544
- [x] port N3LO test #547 
- [x] #550

Only after all that, in this PR:
- [x] Expose necessary constants.
- [x] Add helper functions.
- [x] Add guards in ekore_capi.
- [x] Add tests.
- [x] Add CI
- [x] Add CD.
- [x] Add doc-strings.
- [x] Explore removal of duplicate code.
- [x] Add Changelog entry


AkshatRai07:
For change in visibility of ekore functions, I changed all of them to `pub(super)` and only the functions in the `spacelike.rs` files are now `pub`. Those who caused errors on being `pub(super)` (because they are called by functions outside their mod but inside the crate), their visibility was changed to `pub(crate)`.

What I will do later:
- Create C, C++, and maybe Fortran files for running tests.
- Add `ccompile`, `ctest`, and other relevant tasks in pyproject.toml for poethepoet.
- Add CI/CD files.

There are a few doubts for which I want your opinion:

For ekore:
- What should be the visibility for functions in harmonics?

For ekore_capi:
- Should we expose `constants.rs`?
- Should we expose `harmonics`? If yes then which functions?
- In ekore crate, tests are only for the `ad_us`, should I write tests for all spacelike.rs (as they are there on the python side), and hence write the same tests for capi?
- How do we want to publish this? Do we want to push this crate to crates.io? In pineappl it is pointed out that cargo-c cannot fetch over the internet, and cargo install won't work. What I've been doing is:
```bash
cargo cinstall --release -p ekore_capi --destdir=./crates/ekore_capi/dist --prefix=/
```
We can then publish the dist as a GitHub release (as it is done in pineappl).
- Should we write extensive doc comments, detailing the function's purpose and parameters?

My initial thought is to make only the `harmonics::cache::Cache` public, expose that through c_api. If required, we can change visibility of others later. We should also expose constants.rs, write more tests, publish the way pineappl does it, and add doc comments.

Other things I thought we could've done for #518, but can be done here:
- Add a `--release` tag in poethepoet's compile task.
- Remove all the mallocs inside the ekore and eko crate. For eg. replacing all the `vec!` with constant size arrays.
- Pass a buffer through rust_quad_ker such that it doesn't malloc at every quadrature point, so one buffer for the entire integration loop.

Do let me know your thoughts.

crates/ekore_capi/src/lib.rs
```diff
@@ -0,0 +1,45 @@
+//! C-language interface for `ekore`.
+
+mod ad_ps;
+mod ad_us;
+mod ome_us;
```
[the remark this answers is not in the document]
AkshatRai07: This is false, it's working without pub. `#[no_mangle] pub extern "C"` controls the visibility.

AkshatRai07:
CoPilot reviews Summary:
- There will be `panic` if we pass order_qcd as 0 (As you cannot initialize a 0 dimension vec). I don't think you'll ever pass that, there was no such guard in `ekore`, but still no problem in adding a guard here.
- CoPilot is stating that there might be accidental panics (by say accidentally passing a wrong value), and these panics can cause Undefined Behaviour, so we should add safeguards. This is partially True. This behavior occurred for versions before `1.81`. For `1.81+`, it results in the crashing of code instead of undefined behaiour. We can either bump the `1.70` to `1.81` in our workspace Cargo.toml or add graceful handling.
- CoPilot says to use `checked_mul`. This is just noise, there's no way `order_qcd * 9` or something will get close to max for `usize`.

My initial thought is to add safeguards and not bump the Rust version. Please let me know your thoughts too.

AkshatRai07:
Small note, the `data.nnpdf.science` site is down, assets.sh is failing locally too, and that is why these tests are failing. Do you know the status of the site?

felixhekhorn:
> Small note, the `data.nnpdf.science` site is down, assets.sh is failing locally too, and that is why these tests are failing. Do you know the status of the site?

Indeed, I noticed. I hope the relevant people become aware quickly and are able to do something ...

AkshatRai07:
Should I fix #538 here or in a separate branch?

crates/ekore_capi/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+- our DOI: [![DOI](https://zenodo.org/badge/DOI/10.5281/zenodo.3874237.svg)](https://doi.org/10.5281/zenodo.3874237)
+- our paper: [![arXiv](https://img.shields.io/badge/arXiv-2202.02338-b31b1b?labelColor=222222)](https://arxiv.org/abs/2202.02338)
+
+## Crates in the eko framework
+
+- [dekoder](https://crates.io/crates/dekoder) - Reading and writing EKO output files
+- [eko](https://crates.io/crates/eko) - Core EKO utilities
+- [ekore](https://crates.io/crates/ekore) - Anomalous dimensions and operator matrix elements
+
```
felixhekhorn, line 20:
```suggestion
- [ekore_capi](https://crates.io/crates/ekore) C API for ekore

```
here and everywhere else

crates/ekore_capi/README.md
```diff
@@ -0,0 +1,22 @@
+# ekore_capi
+
+C API for the [ekore](https://crates.io/crates/ekore) crate, part of the [EKO](https://github.com/NNPDF/eko) framework.
```
felixhekhorn:
```suggestion
C API for the [ekore](https://crates.io/crates/ekore) crate which is a part of the [EKO](https://github.com/NNPDF/eko) framework.
```

crates/ekore_capi/Cargo.toml
felixhekhorn, line 1: this crate needs to be registered in `crates/release.json`

crates/ekore_capi/Cargo.toml
```diff
@@ -0,0 +1,33 @@
...
+
+[lib]
+crate-type = ["cdylib", "staticlib"]
+
+[features]
+capi = []
+
+[dependencies]
+ekore = { path = "../ekore", version = "0.0.1" }
+num = "0.4.1"
```
felixhekhorn: should we write a script (or similar) to enforce the same dependencies in the framework? e.g. all crates use `num="0.4.1"`

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -0,0 +1,53 @@
+//! The polarized, space-like anomalous dimensions at various couplings power.
+
+use crate::ComplexF64;
+use ekore::anomalous_dimensions::polarized::spacelike;
+use ekore::harmonics::cache::Cache;
+use std::slice;
+
+/// Compute the tower of non-singlet anomalous dimensions.
```
felixhekhorn, line 20: We need the full documentation here. At first you can try to copy (or guess from other places) and I can revise it afterwards

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -0,0 +1,53 @@
...
+#[no_mangle]
+pub unsafe extern "C" fn ad_ps_gamma_ns_qcd(
+    order_qcd: usize,
+    mode: u16,
+    c: *mut Cache,
+    nf: u8,
+    n3lo_variation: *const u8,
```
felixhekhorn: If I remember correctly in Rust, unlike in Python, it is good practice to explicitly complain about wrong inputs, right? we are not doing this systematically yet - should we start doing so? How do errors bubble through the C API?

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -0,0 +1,53 @@
...
+    result: *mut ComplexF64,
```
felixhekhorn, line 38: The required length of `result` must be explicitly mentioned in the docstring. Should we provide a convenience wrapper, e.g. `ad_ps_gamma_ns_qcd_len(order_qcd) -> 2*order_qcd` (or something similar)? or would it be too much - people should read the docstring and basta?

crates/ekore_capi/src/ad_ps.rs
felixhekhorn, line 20: Since this is C there is no scoping, right? i.e. the exported name is just `ad_ps_gamma_ns_qcd`, right?
- are we sure about the (so far purely internal) abbreviations? or should it be `anomalous_dimensions_polarized_spacelike_gamma_ns_qcd`? it would be a looong name (no, we will not expand `ns` or `qcd`) ... or we expect people to read the documentation and basta?
- any way, the documentation has to be explicit: `Compute the tower of polarized spacelike non-singlet anomalous dimensions.` If that is too long for the first line, the specification must be noted first in the description below.

crates/ekore_capi/src/ad_ps.rs
felixhekhorn, line 1: I wonder if we should apply some kind of meta programming ... because the function body of `ad_ps_gamma_ns_qcd` will look the same in all variations. The only thing that might change (and I'm not sure how that would manifest in the interface) is the maximum available order ...

crates/ekore_capi/src/ad_ps.rs
felixhekhorn: From what Copilot does, it seems it just wants to leave the result empty - which seems also to be what PineAPPL is doing ...

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -0,0 +1,53 @@
+//! The polarized, space-like anomalous dimensions at various couplings power.
```
felixhekhorn:
```suggestion
//! The polarized, space-like anomalous dimensions.
```
since we don't expose individual couplings any longer

felixhekhorn (changes requested):
- many comments generalize to other cases
- I think it would be a good idea to split cf61e01529a7df33c6bf0ec3dbb6fdaf290e65f1 into a separate PR (since it affects lot's of files and does effectively nothing) and then put this one with just the C API on top

felixhekhorn:
> Other things I thought we could've done for #518, but can be done here:
> 
>     * Add a `--release` tag in poethepoet's compile task.
> 
>     * Remove all the mallocs inside the ekore and eko crate. For eg. replacing all the `vec!` with constant size arrays.
> 
>     * Pass a buffer through rust_quad_ker such that it doesn't malloc at every quadrature point, so one buffer for the entire integration loop.
> 
> 
> Do let me know your thoughts.

> Should I fix #538 here or in a separate branch?

all these go into separate PRs; the `--release` tag and fixing #538 should have priority over this one I would say

felixhekhorn:
> For ekore:
>     * What should be the visibility for functions in harmonics?
> For ekore_capi:
>     * Should we expose `harmonics`? If yes then which functions?
> My initial thought is to make only the `harmonics::cache::Cache` public, expose that through c_api. If required, we can change visibility of others later. 

I would say, let's follow our established logic, that we first expose as little as possible. In this case we have to expose `Cache` since it is an explicit parameter, but everything else should be hidden if possible. Since C and Rust should have the same surface, this implies making Rust more private. Since we think here for the first time seriously about visibility, I'm not worried about hiding stuff.

> For ekore_capi:
>     * Should we expose `constants.rs`?

I think this requires some work, because again I want to expose as little as possible. We need to expose the PID as they are an explicit input, but I want to hide all the other things. I suggest to split the files into several parts corresponding to different math/physics and then we only expose what we need.

>     * In ekore crate, tests are only for the `ad_us`, should I write tests for all spacelike.rs (as they are there on the python side), and hence write the same tests for capi?

> write more tests

yes, please. Please import all tests from the Python side which are not implemented yet. My idea is to use these same tests also in the CI to test the capo.

>     * How do we want to publish this? Do we want to push this crate to crates.io? In pineappl it is pointed out that cargo-c cannot fetch over the internet, and cargo install won't work. What I've been doing is:
> 
> 
> ```shell
> cargo cinstall --release -p ekore_capi --destdir=./crates/ekore_capi/dist --prefix=/
> ```
> 
> We can then publish the dist as a GitHub release (as it is done in pineappl).
>

> publish the way pineappl does it

yes, let's follow what other people do (and make our life easier)


>     * Should we write extensive doc comments, detailing the function's purpose and parameters?

> and add doc comments

yes, we must. If we want to provide an API we must provide the documentation along side it else it is pointless

felixhekhorn:
>     * CoPilot is stating that there might be accidental panics (by say accidentally passing a wrong value), and these panics can cause Undefined Behaviour, so we should add safeguards. This is partially True. This behavior occurred for versions before `1.81`. For `1.81+`, it results in the crashing of code instead of undefined behaiour. We can either bump the `1.70` to `1.81` in our workspace Cargo.toml or add graceful handling.

> and not bump the Rust version

We can safely raise the MSRV I would say. Again, let's follow PineAPPL and they are currently at "raised MSRV to 1.91.0
raised Rust edition to 2024" - [see Changelog](https://github.com/NNPDF/pineappl/blob/dba5c8c6f530b9825fd81badcb221e0c774d2f4e/CHANGELOG.md?plain=1#L49) . We also should add an [MSRV workflow](https://github.com/NNPDF/pineappl/blob/master/.github/workflows/msrv.yml) to our repo

> CoPilot reviews Summary:
> 
>     * There will be `panic` if we pass order_qcd as 0 (As you cannot initialize a 0 dimension vec). I don't think you'll ever pass that, there was no such guard in `ekore`, but still no problem in adding a guard here.
> 
>     * CoPilot says to use `checked_mul`. This is just noise, there's no way `order_qcd * 9` or something will get close to max for `usize`.
> 
> 
> My initial thought is to add safeguards

I agree - I think it is the rust way to add guards

felixhekhorn:
> What I will do later:
>     * Create C, C++, and maybe Fortran files for running tests.
>     * Add `ccompile`, `ctest`, and other relevant tasks in pyproject.toml for poethepoet.
>     * Add CI/CD files.

remember that Fortran has very low priority, but all the other stuff is crucial.

Now, looking at the flood of comments above I think this PR is a good start, but we should split some stuff from this PR and address problems that got revealed here first, before pushing further on this front (when we will surely discover more problems). In short, let's try to make a bunch of small PRs first instead of one big one

crates/dekoder/README.md
```diff
@@ -16,6 +16,7 @@ When using our code please cite
 - [dekoder](https://crates.io/crates/dekoder) - Reading and writing EKO output files
 - [eko](https://crates.io/crates/eko) - Core EKO utilities
 - [ekore](https://crates.io/crates/ekore) - Anomalous dimensions and operator matrix elements
+- [ekore_capi](https://crates.io/crates/ekore) C API for ekore
```
felixhekhorn: hihi, don't trust me blindly - as you can see I copied to much 😇  the URL should point to where you can get the thing; and if we don't publish at crates.io (what I understood) it should point to somewhere else - e.g. this repo if we add it there

AkshatRai07: Oh yeah, it should point to ekore_capi 😅

crates/ekore_capi/Cargo.toml
AkshatRai07: We can define the version in workspace, i.e. root Cargo.toml, then use that everywhere else.

AkshatRai07:
> let's try to make a bunch of small PRs first instead of one big one

Let me add a TODO as a comment here:

EDIT: moved to top

crates/ekore_capi/src/ad_ps.rs
AkshatRai07: By the logic we have shortened `ns`, let it remain `ad_ps`. It won't take long to realize what the full form is.

crates/ekore_capi/src/ad_ps.rs
AkshatRai07: It's good to have helper functions, I'll make them.

crates/ekore_capi/src/ad_ps.rs
AkshatRai07: We can explore this after most of the stuff is done.

felixhekhorn:
Is this relevant for us https://github.com/NNPDF/pineappl/commit/823a4f7478a5f4cc0afa6e6c288fd728ca10e628 ?

crates/ekore_capi/src/ad_ps.rs
AkshatRai07: Using `macro_rules!`, we can remove duplicate code. Do tell me if I should start working on that.

.github/workflows/unittests-capi.yml
```diff
@@ -0,0 +1,35 @@
...
+      - name: Get test data
+        id: cache-test-data
+        uses: actions/cache@v4
+        with:
+          path: test-data
+          key: test-data-v2
+      - name: Download test data
+        if: steps.cache-test-data.outputs.cache-hit != 'true'
+        run: |
+          ./tests/data/assets.sh
```
felixhekhorn: We don't need any data, no?  so drop

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -1,20 +1,80 @@
...
 use ekore::anomalous_dimensions::polarized::spacelike;
 use ekore::harmonics::cache::Cache;
 use std::slice;
 
+/// Required length of `n3lo_variation` for [`ad_ps_gamma_ns_qcd`].
+///
+/// # Returns
+/// * Returns the fixed required buffer length of `3`.
+#[unsafe(no_mangle)]
+pub extern "C" fn ad_ps_gamma_ns_qcd_n3lo_len() -> usize {
```
felixhekhorn: I don't think this function makes sense: `n3lo_variation` in `ad_ps_gamma_ns_qcd` must be exactly 3 and no other and users must provide exactly those and they must know what it refers to - drop

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -0,0 +1,173 @@
...
+        return 0;
+    }
+    order_qcd
+}
+
+/// Compute the tower of non-singlet anomalous dimensions.
+///
+/// # Safety
+/// * `c` must be a valid, non-null pointer to an initialized `Cache`.
+/// * `n3lo_variation` must be a valid, non-null pointer to a buffer of `u8` elements.
```
felixhekhorn:
```suggestion
/// * `n3lo_variation` must be a valid, non-null pointer to a buffer of `u8` three elements.
```

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -0,0 +1,173 @@
...
+/// * `result_len`: The actual length (in elements) of the provided `result` buffer. This should
+///   be at least the value returned by [`ad_ps_gamma_ns_qcd_result_len`].
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn ad_ps_gamma_ns_qcd(
+    order_qcd: usize,
+    mode: u16,
+    c: *mut Cache,
+    nf: u8,
+    n3lo_variation: *const u8,
+    n3lo_len: usize,
```
felixhekhorn: drop as discussed above - the size is (rightfully) forced here https://github.com/NNPDF/eko/blob/76ae5bfd8ee706c58424e9b105f5990e26b1f8ef/crates/ekore_capi/src/ad_us.rs#L82

crates/ekore_capi/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+    assert!(MAX_ORDER_QCD == ekore::constants::MAX_ORDER_QCD);
+    assert!(MAX_ORDER_QED == ekore::constants::MAX_ORDER_QED);
+    assert!(PID_NSP == ekore::constants::PID_NSP);
+    assert!(PID_NSM == ekore::constants::PID_NSM);
+    assert!(PID_NSV == ekore::constants::PID_NSV);
+    assert!(PID_NSP_U == ekore::constants::PID_NSP_U);
+    assert!(PID_NSP_D == ekore::constants::PID_NSP_D);
+    assert!(PID_NSM_U == ekore::constants::PID_NSM_U);
+    assert!(PID_NSM_D == ekore::constants::PID_NSM_D);
+};
```
felixhekhorn, line 105: why are they not just re-imported in the first place? no need to define them twice, no?

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -0,0 +1,173 @@
...
+    result: *mut ComplexF64,
+    result_len: usize,
```
felixhekhorn: I feel like, I might miss something important: why do we ask for `result_len`, but then basically not use it? we check its length, fine, but this does not guarantee that we can write there and then we just do it ... or is that parameter not provided by the user (but the interface somehow)?

crates/ekore_capi/tests/c/test_ad_ps.c
felixhekhorn, line 1: should we add a proper unit testing framework?

crates/ekore_capi/tests/c/test_ad_ps.c
```diff
@@ -0,0 +1,134 @@
+#include "utils.h"
+#include <ekore_capi.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+static int check_len(const char *label, size_t got, size_t expected)
```
felixhekhorn: if we don't go for a proper framework, surely this belongs into `utils`

crates/ekore_capi/tests/cpp/test_ad_ps.cpp
felixhekhorn, line 1: this is (quasi) verbatim the same file as for `.c`, no? why would we do such a thing? I think C is enough and users who write C++ can use that just the same, I'd say

crates/ekore_capi/src/ad_ps.rs
felixhekhorn: I would say yes

.github/workflows/release-capi.yml
```diff
@@ -0,0 +1,102 @@
...
+
+jobs:
+  set-version:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Enforce tag-only execution
+        if: github.ref_type != 'tag'
+        run: |
+          echo "Error: Manual dispatches must be run against a tag, not a branch."
+          exit 1
```
felixhekhorn: This is inconvenient for debugging - can't we change it such that if it is a tag the files are uploaded there, if it is not, the files are just stored as artifacts to the job?

felixhekhorn (changes requested):
some of the comments obviously generalise to other cases, if you agree

crates/ekore_capi/src/lib.rs
AkshatRai07: These are duplicated because of how cbindgen works, it generates the header by parsing this crate's own source, not by resolving re-exports through to their defining crate.
That would require `parse_deps = true` and `include = ["ekore"]` in cbindgen.toml. I tested that without redefining the constants and using `pub use ekore::constants::*`, but cbindgen skipped every constant with `"const's outside of the binding crate are not used"`, it only pulls in dependency items that are used as a type in the `ekore_capi` signatures (that's why the `Cache` type got through, since it's taken as a parameter in the functions).
A glob re-export doesn't count as usage in that sense, there's no cbindgen config that exposes re-exported constants from a dependency.
That's why, constants are defined again with the compile-time `assert!` block which will stop the compilation if there is some change in the constants.

crates/ekore_capi/src/ad_ps.rs
AkshatRai07: `result_len` is caller-provided (they get it from the corresponding helper), then pass back the actual capacity of the buffer they allocated. If they allocate less than needed, writing without checking `result_len` would write past the end of their buffer, corrupting heap or stack memory silently. The check turns that into a safe return instead.

It cannot prove that result really points to that many valid elements, we can only trust what the caller reports. That's written in `# Safety`, the caller is responsible for result actually being a valid buffer of at least stated size. That's a problem with C-ABI, it passes only pointers, and anything could be behind the pointer.

We can, if you want, change the design. Instead of user allocating the buffer, we do it, and then pass the result to the user. Please do let me know if we should switch.

crates/ekore_capi/tests/c/test_ad_ps.c
AkshatRai07: No, the test suite is not that huge, and the helper functions are enough to check the validity IMO

crates/ekore_capi/tests/cpp/test_ad_ps.cpp
AkshatRai07: Yes, I was thinking the same while making the tests. If it works in C, it would definitely work in C++ and Fortran.

crates/ekore_capi/src/lib.rs
felixhekhorn: I understand we can not re-export, but can we then just redefine, like
```rust
pub const PID_NSP_U: u16 = ekore::constants::PID_NSP_U;
```
?

(if we can't improve, we can live with it of course)

crates/ekore_capi/src/ad_ps.rs
felixhekhorn: > the caller is responsible for result actually being a valid buffer

exactly and we should keep it that way (to make live easier for us). I'm simply arguing that we should drop that argument and trust the user knows what he is doing. Let's be specific:
```cpp
int len = gamma_blub_len();
double* data = new double[len];
gamma(data)
```
to me is "safer" or more unambiguous or makes the responsibility more clear then
 ```cpp
int len = gamma_blub_len();
double* data = new double[len];
gamma(data, len)
```
This is because the crucial line is the second, which is present in either case - and it must be. If the user can lie (and will lie) about the `result_len` argument, I would say it is better not to take it in the first place.

But maybe this is just my personal coding style :upside_down_face: you can convince me otherwise. Maybe let's also ask to @scarlehoff for an opinion ...

File: crates/ekore_capi/tests/cpp/test_ad_ps.cpp
Comment (felixhekhorn) [reply crates/ekore_capi/tests/cpp/test_ad_ps.cpp:1]: Re Fortran: actually since there will be a layer in between like [in PineAPPL](https://github.com/NNPDF/pineappl/blob/master/examples/fortran/pineappl.f90) I'd say it is worth testing - just to exclude some typos in the translation layer ... let me repeat Fortran is low on our priority list

re C++: actually, if we want to do something here we could go [in the direction I introduced into PineAPPL](https://github.com/NNPDF/pineappl/blob/master/examples/object-oriented-cpp/PineAPPL.hpp), something like
```cpp
namespace ekore {
namespace anomalous_dimensions {
namespace unpolarized {
// remap the names and take STL arguments like vector<double> and such
}
}
}
```
but I'm not sure we want to go in this directions and surely it is at the very bottom of our priority list; so I would drop C++ for the moment

crates/ekore_capi/src/lib.rs
AkshatRai07: > pub const PID_NSP_U: u16 = ekore::constants::PID_NSP_U;

Got this warning while trying to implement your solution:
```text
WARN: Skip lib::MAX_ORDER_QCD - (Unsupported path expression. Path { leading_colon: None, segments: [PathSegment { ident: Ident(ekore), arguments: PathArguments::None }, PathSep, PathSegment { ident: Ident(constants), arguments: PathArguments::None }, PathSep, PathSegment { ident: Ident(MAX_ORDER_QCD), arguments: PathArguments::None }] })
```
It is not able to resolve what's the value in the ekore crate, cbindgen looks for values inside ekore_capi only. Hence, these constants don't occur in the header file if we directly use the imported values.

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs
```diff
@@ -289,8 +289,11 @@ mod tests {
     #[test]
     fn test_gamma_ns_qcd() {
         const NF: u8 = 3;
+        const NF_: u8 = 5;
```
felixhekhorn:
```suggestion
        const NF3: u8 = 3;
        const NF5: u8 = 5;
```
would be more telling names

crates/ekore_capi/src/ad_ps.rs
AkshatRai07: So we can make macro_rules to create functions, but `cbindgen` doesn't catch those functions and hence they are missing from the header files.  We can force it to make functions first using `cargo expand` and then make the dist, or instead use macros to replace the body and not the entire function. Went with the latter one because cargo expand requires nightly version of Rust.

felixhekhorn: agreed

crates/ekore_capi/tests/c/utils.h
```diff
@@ -0,0 +1,17 @@
+#ifndef EKORE_TEST_UTILS_H
+#define EKORE_TEST_UTILS_H
+
+#define DEFAULT_EPS 1e-15
```
felixhekhorn: Is this even used?

crates/ekore_capi/tests/c/utils.c
```diff
@@ -0,0 +1,34 @@
+#include "utils.h"
+#include <math.h>
+#include <stdio.h>
+
+int check(const char *test, const char *label,
+          double got_re, double got_im,
+          double exp_re, double exp_im, double eps)
+{
+    double diff = fabs(got_re - exp_re) + fabs(got_im - exp_im);
```
felixhekhorn, line 10:
```suggestion
    // This is effectively spanning a diamond-like shape around got_re + I*got_im
    double diff = fabs(got_re - exp_re) + fabs(got_im - exp_im);
```

crates/ekore_capi/tests/run_tests.sh
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env bash
+set -e
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```
felixhekhorn, line 5:
```suggestion
# setup environment
SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
```

crates/ekore_capi/tests/run_tests.sh
```diff
@@ -0,0 +1,43 @@
...
+RPATH="$(realpath "$LIB_DIR")"
+
+run_section() {
+    local lang="$1" compiler="$2" ext="$3"
+    local dir="$SCRIPT_DIR/$lang"
+    local utils_obj="" extra=()
+
+    echo "=== Running $lang tests ==="
+
+    if [ -f "$dir/utils.$ext" ]; then
```
felixhekhorn, line 18:
```suggestion
    # compile utils for everybody
    if [ -f "$dir/utils.$ext" ]; then
```

crates/ekore_capi/tests/run_tests.sh
```diff
@@ -0,0 +1,43 @@
...
+        utils_obj="$dir/utils.o"
+        echo "  Compiling utils.$ext..."
+        "$compiler" -c "$dir/utils.$ext" -I"$INCLUDE_DIR" -o "$utils_obj"
+        extra+=("$utils_obj")
+    fi
+
+    extra+=("-I$INCLUDE_DIR")
+
+    for src in "$dir"/*."$ext"; do
```
felixhekhorn, line 28:
```suggestion
    # iterate test files
    for src in "$dir"/*."$ext"; do
```

crates/ekore_capi/src/macros.rs
```diff
@@ -0,0 +1,191 @@
...
+        $formula
+    }};
+}
+
+/// Body of a non-singlet QCD anomalous dimension tower function.
+macro_rules! gamma_ns_qcd_body {
+    (
+        $order_qcd:expr, $mode:expr, $c:expr, $nf:expr, $n3lo_variation:expr,
+        $result:expr, $result_len:expr, $bound:expr, $path:path
+    ) => {{
```
felixhekhorn:
```suggestion
    ) => {{
    /// Sanity checks
```

crates/ekore_capi/src/macros.rs
```diff
@@ -0,0 +1,191 @@
...
+        if $result_len < $order_qcd {
+            return;
+        }
+        unsafe {
+            let c = &mut *$c;
+            let var: [u8; 3] = std::slice::from_raw_parts($n3lo_variation, 3)
+                .try_into()
+                .unwrap();
+            let out = std::slice::from_raw_parts_mut($result, $order_qcd);
+            for (dst, src) in out.iter_mut().zip($path($order_qcd, $mode, c, $nf, var)) {
```
felixhekhorn:
```suggestion
            // Call function and transfer result
            for (dst, src) in out.iter_mut().zip($path($order_qcd, $mode, c, $nf, var)) {
```

felixhekhorn (changes requested):
Please add more documentation

felixhekhorn:
After #550 is merged, we can then add a sentence to the C API file, something like `For the list of available perturbative orders see [here](ekore::bla::blub)`

crates/ekore_capi/src/lib.rs
```diff
@@ -0,0 +1,98 @@
+//! C-language interface for `ekore`.
```
felixhekhorn: Also this part needs do be expanded. I'm not sure how much, but something needs to be specified. I guess we should at least mention `pkg-config` and `cbindgen`. I think we should also repeat the return array policy here again (that we fill the result up to the requested order). Also links to the ekore create and the main eko documentation would be good.

felixhekhorn:
and also, please fix
```
warning: unresolved link to `ad_ps_gamma_ns_qcd_n3lo_len`
  --> crates/ekore_capi/src/ad_ps.rs:34:37
   |
34 | ///   least the value returned by [`ad_ps_gamma_ns_qcd_n3lo_len`].
   |                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ no item named `ad_ps_gamma_ns_qcd_n3lo_len` in scope
```

crates/ekore_capi/src/lib.rs
felixhekhorn: also we need to explain our short-cut notation there: `ad_us` -> unpolarized, space-like anomalous dimensions

crates/ekore_capi/tests/c/utils.h
AkshatRai07: `#ifndef` is used as an include guard in C files. If a file is imported twice, the compiler will import and write the headers twice, which will cause an error because functions will be defined twice. With this guard the compiler will include the file only if `EKORE_TEST_UTILS_H` is not defined, hence only once.

And `DEFAULT_EPS` is used in the program in `test_ad_us.c` but at only 4 places. We should remove that.

crates/ekore_capi/src/lib.rs
AkshatRai07: > we should at least mention pkg-config and cbindgen

> links to the ekore create and the main eko documentation would be good.

Except for `pkg-config`, these are mentioned in the `README.md`. For other changes I'll write the docs.

crates/ekore_capi/tests/c/utils.h
felixhekhorn: > `#ifndef` is used as an include guard in C files.

this I know :upside_down_face:  I may have misclicked when adding this comment :see_no_evil: 

> And `DEFAULT_EPS` is used in the program in `test_ad_us.c` but at only 4 places. We should remove that.

mmm I tend to agree, since it is numerics (and a random value) anyway ...

crates/ekore_capi/src/lib.rs
felixhekhorn: > these are mentioned in the `README.md`.

I think we need to repeat some stuff, since the Readme is rendered at crates.io, but this file at docs.rs - and as a user I would always consult the latter to find out how this crate works and what it does ...

CHANGELOG.md
```diff
@@ -15,6 +15,9 @@ Items without prefix refer to a global change.
 
 ## [Unreleased](https://github.com/NNPDF/eko/compare/v0.15.4...HEAD)
 
+### Added
+- rust: `ekore_capi` crate, which exposes `ekore` crate as a C-ABI for other languages.
```
felixhekhorn:
```suggestion
- rust: `ekore_capi` crate, which exposes `ekore` crate as a C-ABI for other languages. ([#537](https://github.com/NNPDF/eko/pull/537))
```

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -0,0 +1,98 @@
...
+/// * `mode`: The specific non-singlet sector.
+/// * `c`: Pointer to the harmonic cache.
+/// * `nf`: Number of active flavors.
+/// * `n3lo_variation`: Pointer to the buffer containing N3LO variations.
+/// * `result`: Pointer to the output buffer.
+#[unsafe(no_mangle)]
+pub unsafe extern "C" fn ad_ps_gamma_ns_qcd(
+    order_qcd: usize,
+    mode: u16,
+    c: *mut Cache,
```
felixhekhorn:
```suggestion
    cache: *mut Cache,
```
let's use telling names also here (remember docs)

crates/ekore_capi/src/ad_ps.rs
```diff
@@ -0,0 +1,98 @@
...
+    nf: u8,
+    n3lo_variation: *const u8,
```
felixhekhorn:
```suggestion
```
as consequence from #550

felixhekhorn (changes requested):
after solving the consequences from #550 I think this is good to go!

crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
```diff
@@ -17,7 +17,6 @@ pub fn gamma_ns_qcd(
     mode: u16,
     c: &mut Cache,
```
felixhekhorn, line 18: On `master` this has the correct name (from #550) - did the merge work correctly?

AkshatRai07: Yes it merged correctly.
The name is correct on master, and after merging this PR it will remain that way. Idk why but Github was showing merge conflict in `tests` of `ad_ps` and `ad_us`, hence copied that part from master to this branch.

felixhekhorn (approved):

PR #542, opened 2026-06-24
Remove mallocs and pass single buffer in run_op_integration

AkshatRai07:
A pre-requisite for #537 and part of #518
- Removed all mallocs from `spacelike.rs` and `cache.rs` files with necessary changes in `eko/src/lib.rs`.
- Passed only one buffer for the whole integration operation.
- Removed some minor loops on the Python side.

### DIFF
--- a/crates/eko/src/lib.rs
+++ b/crates/eko/src/lib.rs
@@ -99,9 +99,9 @@ pub unsafe extern "C" fn rust_quad_ker(u: f64, rargs: *mut c_void) -> f64 {
         .try_into()
         .unwrap();
 
-    let n_max = (args.order_qcd + 1) * (args.order_qed + 1) * 16;
-    let out_re = std::slice::from_raw_parts_mut(args.re_gamma, n_max);
-    let out_im = std::slice::from_raw_parts_mut(args.im_gamma, n_max);
+    let max_buffer_size = (args.order_qcd + 1) * (args.order_qed + 1) * 16;
+    let out_re = std::slice::from_raw_parts_mut(args.re_gamma, max_buffer_size);
+    let out_im = std::slice::from_raw_parts_mut(args.im_gamma, max_buffer_size);
 
     if args.is_ome {
         if is_singlet {
--- a/crates/eko/src/lib.rs
+++ b/crates/eko/src/lib.rs
@@ -16,7 +16,7 @@ struct RawCmplx {
 /// Map tensor with shape (o,d,d) to c-ordered list.
 ///
 /// This is needed for the QCD singlet.
-fn unravel<const DIM: usize>(res: Vec<[[Complex<f64>; DIM]; DIM]>, order_qcd: usize) -> RawCmplx {
+fn unravel<const DIM: usize>(res: &[[[Complex<f64>; DIM]; DIM]], order_qcd: usize) -> RawCmplx {
     let mut target = RawCmplx {
         re: Vec::<f64>::new(),
         im: Vec::<f64>::new(),
@@ -36,7 +36,7 @@ fn unravel<const DIM: usize>(res: Vec<[[Complex<f64>; DIM]; DIM]>, order_qcd: us
 ///
 /// This is needed for the QED singlet and valence.
 fn unravel_qed<const DIM: usize>(
-    res: Vec<Vec<[[Complex<f64>; DIM]; DIM]>>,
+    res: &[[[[Complex<f64>; DIM]; DIM]; 3]],
     order_qcd: usize,
     order_qed: usize,
 ) -> RawCmplx {
@@ -60,7 +60,7 @@ fn unravel_qed<const DIM: usize>(
 /// Map tensor with shape (o,o',d) to c-ordered list.
 ///
 /// This is needed for the QED non-singlet.
-fn unravel_qed_ns(res: Vec<Vec<Complex<f64>>>, order_qcd: usize, order_qed: usize) -> RawCmplx {
+fn unravel_qed_ns(res: &[[Complex<f64>; 3]], order_qcd: usize, order_qed: usize) -> RawCmplx {
     let mut target = RawCmplx {
         re: Vec::<f64>::new(),
         im: Vec::<f64>::new(),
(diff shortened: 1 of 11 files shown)


crates/ekore/src/harmonics/cache.rs
```diff
@@ -50,20 +49,49 @@ pub enum K {
...
+    m: [Option<Complex<f64>>; N_CACHE],
 }
 
 impl Cache {
     /// Initialize new, empty Cache at given Mellin N.
     pub fn new(n: Complex<f64>) -> Self {
         Self {
             n,
-            m: HashMap::new(),
+            m: [None; N_CACHE],
```
felixhekhorn: what is the advantage of using a list over `HashMap`? I'm fine with either just want to understand your rationale

crates/ekore/src/harmonics/cache.rs
```diff
@@ -50,20 +49,49 @@ pub enum K {
...
+            K::Sm2o => 15,
+            K::Sm3e => 16,
+            K::Sm3o => 17,
+            K::Sm21e => 18,
+            K::Sm21o => 19,
+        }
+    }
+}
+
+const N_CACHE: usize = 20;
```
felixhekhorn: if you want to go this way: please add a unit test, which explicitly checks that
- the mapping is continuous (no holes)
- the size is the correct one
- `N_CACHE` in this case is not a good name, as there is the `n` flowing around - just rename it to, e.g. `CACHE_SIZE` or similar

crates/eko/src/lib.rs
```diff
@@ -7,71 +7,69 @@ use std::ffi::c_void;
...
+/// Map tensor with shape (o,d,d) into pre-allocated output buffers.
 ///
 /// This is needed for the QCD singlet.
-fn unravel<const DIM: usize>(res: Vec<[[Complex<f64>; DIM]; DIM]>, order_qcd: usize) -> RawCmplx {
-    let mut target = RawCmplx {
-        re: Vec::<f64>::new(),
-        im: Vec::<f64>::new(),
-    };
+fn unravel<const DIM: usize>(
+    res: &[[[Complex<f64>; DIM]; DIM]],
```
felixhekhorn: I think we know the outer dimension: it will be `MAX_ORDER_QCD`. We can also improve the docstring has say explicitly `o` -> `order_QCD`.

Similar comments apply to the other `unravel`.

We can rename this function to `unravel_qcd` if it is only used for the QCD singlet (as the docstring claims)

felixhekhorn (changes requested):
We need to disentangle the integers on the tensors. For that we need to define two constants `const MAX_ORDER_QCD : usize = 4` and `MAX_ORDER_QED=2` and then use them in the appropriate places. In particular, while the inner most dimensions of the tensors are typically given by (flavor) algebra, the outer are given by our current knowledge. This constants will also help with the potential helper functions #537 . Be careful that the size is usually `MAX_ORDER`, but in the QED case it is shifted by one since we fill a matrix. The physics statement is: in eko `order` refers to the power of the coupling (e.g. `as1` is $a_s^1$).

We must clearly document the new behaviour: since this concerns all functions it might be clever to do this in a central location (e.g. `crates/ekore/lib.rs`) and then add a single sentence to each function docstring that the buffer has to follow blub. The documentation e.g. needs to explain that we only fill the slots the function has been asked for, which yields undefined behaviour for the remaining entries

crates/ekore/src/harmonics/cache.rs
AkshatRai07: In Rust, the `HashMap::new()` doesn't immediately does a malloc, but all the other subsequent insertions are mallocs. It's better to use an array built into the binary.

crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
```diff
@@ -15,8 +15,8 @@ pub fn gamma_ns_qcd(
     c: &mut Cache,
     nf: u8,
     _n3lo_variation: [u8; 3],
-) -> Vec<Complex<f64>> {
-    let mut gamma_ns = vec![Complex::<f64>::zero(); order_qcd];
+) -> [Complex<f64>; MAX_ORDER_QCD] {
```
felixhekhorn, line 21: Adding the docstrings to the `unravel*` functions is good and necessary, but the crucial part is in ekore of course, since that will be our interface - here the same logic applies

crates/eko/src/lib.rs
```diff
@@ -93,42 +95,46 @@ pub unsafe extern "C" fn rust_quad_ker(u: f64, rargs: *mut c_void) -> f64 {
...
-        im: Vec::<f64>::new(),
-    };
-    let n3lo_ad_variation = std::slice::from_raw_parts(args.n3lo_ad_variation, 7)
+    let n3lo_ad_variation: [u8; 7] = std::slice::from_raw_parts(args.n3lo_ad_variation, 7)
         .try_into()
         .unwrap();
 
+    let n_max = (args.order_qcd + 1) * (args.order_qed + 1) * 16;
+    let out_re = std::slice::from_raw_parts_mut(args.re_gamma, n_max);
+    let out_im = std::slice::from_raw_parts_mut(args.im_gamma, n_max);
```
felixhekhorn:
```suggestion
    let max_buffer_size = (args.order_qcd + 1) * (args.order_qed + 1) * 16;
    let out_re = std::slice::from_raw_parts_mut(args.re_gamma, max_buffer_size);
    let out_im = std::slice::from_raw_parts_mut(args.im_gamma, max_buffer_size);
```
Let's avoid again the name `n` (and keep it for the Mellin N)

felixhekhorn (approved):
Can you please check why the benchmarks are running now on commit, when they should not?
https://github.com/NNPDF/eko/blob/5aebe0abdea3e75ed7761b5ad4ce31d7fd227cd4/.github/workflows/lha_bot_rust.yml#L5-L7

Any way, once all tests pass, this is good for merge

felixhekhorn:
Actually: maybe you want to add a section to the performance document and maybe also adjust the architecture file?

AkshatRai07:
> Can you please check why the benchmarks are running now on commit, when they should not?

GitHub does not match "*" with a "/", which is a part of this branch name.
Got to replace them "**".

> Actually: maybe you want to add a section to the performance document and maybe also adjust the architecture file?

No need to adjust the architecture file, it doesn't go deep as to what we are passing around.
I'll update the performance doc with the data I had provided in #518 ([here](https://github.com/NNPDF/eko/issues/518#issuecomment-4727938997))

PR #543, opened 2026-06-26
Changing visibility of ekore and Splitting constants.rs

### LINKED ISSUE #539: Split `crates/ekore/src/constants.rs`
We need the PID business from  `crates/ekore/src/constants.rs` in our public API (cc #145 ) since they are explicit arguments.

I suggest to split the file into small logical pieces and only expose what is needed:
1. `qcd_constant.rs`: holds all QCD related constants, in particular: `NC, TR, CF, CA` - if ever a `ekoupling` crate materialises they will move there; hidden for our purpose
2. `math_constants.rs` holds all math constants, in particulat `ZETA*` - will remain most likely forever there; both harmonic sums as well as ad may depend explicitly on them; hidden for our purpose
3. `pid.rs` holds all PID related constants, in particular `PID_*, E*, ChargeCombinations`, but only `PID*` will be publicly exposed - if ever a `basis_rotation` crate materialises they will move there

for a list of potential crates see also #383

AkshatRai07:
This PR fixes #539, and is a pre-requisite to #537.
- Changed visibility of functions and modules in `ekore` crate. `pub` was converted to `pub(super)`, and if that resulted in error it was converted to `pub(crate)`.
- Split `constants.rs` to 3 files:
  - `math_constants.rs` with all the Zeta function values.
  - `qcd_constants.rs` with all QCD constants.
  - `pid.rs` with the rest of the constants, with `PID_*` and `MAX_ORDER_*` being the only public constants (are/will be exposed publicly in the future).


AkshatRai07:
I was a bit confused in placing of `MAX_ORDER_*` constants, they aren't mathematical nor confined to QCD, so added them in pid.rs because other public constants are present in that.

Second, there was a warning in `cargo build --workspace`:
```text
   Compiling ekore v0.0.1 (/home/akshatrai/Documents/GSoC/eko/crates/ekore)
warning: variants `Sm2o`, `Sm3o`, and `Sm21o` are never constructed
  --> crates/ekore/src/harmonics/cache.rs:41:5
   |
 9 | pub(crate) enum K {
   |                 - variants in this enum
...
41 |     Sm2o,
   |     ^^^^
...
45 |     Sm3o,
   |     ^^^^
...
49 |     Sm21o,
   |     ^^^^^
   |
   = note: `K` has a derived impl for the trait `Debug`, but this is intentionally ignored during dead code analysis
   = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: `ekore` (lib) generated 1 warning
   Compiling eko v0.0.1 (/home/akshatrai/Documents/GSoC/eko/crates/eko)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.56s
```
This is not a pre-commit issue, just a warning that these three things, `Sm2o`, `Sm3o`, and `Sm21o` are used nowhere in the ekore crate except at their declarations and `K`.

crates/ekore/src/pid.rs
```diff
@@ -0,0 +1,85 @@
...
+pub const PID_NSM_U: u16 = 10202;
+
+/// valence-like non-singlet down-sector |PID|.
+pub const PID_NSM_D: u16 = 10203;
+
+/// Maximum QCD coupling power implemented.
+pub const MAX_ORDER_QCD: usize = 4;
+
+/// Maximum QED coupling power implemented.
+pub const MAX_ORDER_QED: usize = 2;
```
felixhekhorn: Let's move those two to `constants.rs` - they don't really fit here (also if maybe one day we might move this file somewhere else)

felixhekhorn (changes requested):
- Please add a Changelog entry
- I'm not sure yet what to do with the unused functions ... let's discuss tomorrow; most likely we will just drop them

AkshatRai07:
So `Sm1o` was also unused but compiler was not giving warning about that until I removed the other 3 functions.

felixhekhorn:
I change my mind again: let's keep the functions! sorry :innocent: can you please just revert [0bece9e](https://github.com/NNPDF/eko/pull/543/commits/0bece9efc3c5c462e67af3fcbbce28e74699b17a) ? We need them when we port https://github.com/NNPDF/eko/blob/96032431fea3025eeefe6dead981ac6e5c934250/src/ekore/anomalous_dimensions/polarized/space_like/as3.py#L331 to Rust.

Here comes a math statement: at a given order in the coupling (e.g. $a_s^3$ -> `as3.py`), we explore a given set of special functions, in particular a given set of harmonic sums. In this case: $a_s^3$ explores harmonic sum up to weight 3. Specifically, here we need $S_3, S_{-3},S_{-2,1}$, etc. (i.e. the indices sum in absolute value to 3) and "up to" implies we also need lower order ones, e.g. $S_{-2}$. The polarized and unpolarized anomalous dimensions explore two different spaces (some of which is related by "odd" <-> "even"). Here we see an explicit example thereof.

AkshatRai07:
All 4 functions are back now

felixhekhorn (approved):

PR #544, opened 2026-06-26
Add centralised workspace dependencies

AkshatRai07:
A prerequisite for #537.
- Adds dependencies to root `Cargo.toml`, and all crates point to that.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -21,6 +21,7 @@ repository = "https://github.com/NNPDF/eko"
 rust-version = "1.70.0"
 version = "0.0.1"
 
+# all dependencies should be collected here in the workspace
 [workspace.dependencies]
 # Internal crates
 ekore = { path = "crates/ekore", version = "0.0.1" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,3 +20,22 @@ license = "GPL-3.0-or-later"
 repository = "https://github.com/NNPDF/eko"
 rust-version = "1.70.0"
 version = "0.0.1"
+
+[workspace.dependencies]
+# Internal crates
+ekore = { path = "crates/ekore", version = "0.0.1" }
+
+# External dependencies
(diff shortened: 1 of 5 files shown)


Cargo.toml
```diff
@@ -20,3 +20,22 @@ license = "GPL-3.0-or-later"
 repository = "https://github.com/NNPDF/eko"
 rust-version = "1.70.0"
 version = "0.0.1"
+
```
felixhekhorn, line 23:
```suggestion

# all dependencies should be collected here in the workspace
```

felixhekhorn (commented):
[this line](https://github.com/NNPDF/eko/blob/b372afb350539faf21bd6a08e64dbaa1127ca80a/crates/bump-versions.py#L31) should ensure that this change is still compatible with the version script, right? if yes, I think we can add the additional comment and merge

AkshatRai07:
Thanks for pointing it out, we needed to change that file. Now when bump-versions is ran, it changes the workspace version of the project, as well as that of the internal dependencies. Previously the internal dependencies version was changed in crates, but since there is no version in crate level, file is changed to workspace level.

crates/bump-versions.py
```diff
@@ -29,8 +25,6 @@ def update(path, version, edit):
 
 def main(version):
     update("..", version, workspace)
-    for name in CRATES:
```
felixhekhorn, line 32: If we don't need that any more, let's drop also its definition (here and the actual file)

AkshatRai07: Can you elaborate which definition do you want to drop? The `update` function?

felixhekhorn: I originally thought `CRATES` and `release.json` would no longer be needed - but actually they are.

However, looking at the file again I think we can collapse everything into a single function (say `main`) to make it more plain (and not confuse me :innocent: )

felixhekhorn (approved):

PR #545, opened 2026-06-26
Add msrv.yml

AkshatRai07:
This is prerequisite for #537.
- Adds an `msrv.yml` file to workflows.
- Bumps Rust to `1.85.0`.
- Bumps Edition to 2024.


AkshatRai07:
The workflow [failed](https://github.com/NNPDF/eko/actions/runs/28256700995/job/83721476442?pr=545).
I ran `cargo msrv find`, you can run it too but you first have to do `cargo install cargo-msrv`.
This is the end result:
```
Result:
   Considered (min … max):   Rust 0.11.0 … Rust 1.96.0 
   Search method:            bisect                    
   MSRV:                     1.81.0                    
   Target:                   x86_64-unknown-linux-gnu
```
We must bump the version to `1.81.0`. We should not go any further because we need the "minimum" supported version.

felixhekhorn (commented):
looks good to me
- [as said on the other side](https://github.com/NNPDF/eko/pull/537#issuecomment-4788414323) we should (if possible) synchronize with PineAPPL; setting a lower minimum is always fine
- can you please remind me what "Rust edition" is? is it relevant for us?

AkshatRai07:
I am not well versed with version of languages and stuff, but all rust crates require atleast some version for compilation. In our case, there is this `twox-hash` crate which is required by `lz4_flex` crate which is required by the `dekoder` crate:
```text
error: package `twox-hash v2.1.2` cannot be built because it requires rustc 1.81 or newer
```
So one must upgrade rust to 1.81 or later in order to work with the ecosystem.

AkshatRai07:
For Rust Editions, Rust has a philosophy to keep the code always backwards compatible. So if there are some breaking changes, like introduction of a new keyword which may break code, they are released in Editions.

We can check what changes we should do in the code if we want to avoid breaking changes by doing `cargo fix --edition`. Running this did some changes to the `eko/src/lib.rs` and `ekore/src/utils.rs`, it is because they contain `unsafe` and `macro`, which were both affected in Edition 2024.

We should bump the Edition, because it's not costing us much. And it is better in code auditing and readability. For eg. `#[no_mangle]` becomes `#[unsafe(no_mangle)]` in Edition 2024 which is actually an improvement, it makes the unsafe nature of the `no_mangle` attribute explicit rather than hidden, making the code safer and easier to audit.

AkshatRai07:
Reason for the file changes:
- Rust Edition 2024 requires atleast Rust 1.85.0, hence bumped that.
- Rust Edition 2024 I guess also brought changes to `fmt` and `clippy` so had to change the code to not get pre-commit errors. The import order was then changed accordingly (due to new fmt), and there is no `if !(condition)` now in the code but it's `if (not_condition)` (due to new clippy).
- It might look a lot is changed in `eko/src/lib.rs`, but the only change is the `rust_qud_ker` is wrapped in an `unsafe` block, and `fmt` made an indention to the code.

felixhekhorn (approved):
Please add a Changelog entry mentioning version+edition

PR #546, opened 2026-06-30
Add tests to spacelike.rs files

AkshatRai07:
A prerequisite to #537
- Added tests to `spacelike.rs` files.

### DIFF
--- a/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
+++ b/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
@@ -19,6 +19,9 @@ pub fn gamma_ns_qcd(
     nf: u8,
     _n3lo_variation: [u8; 3],
 ) -> [Complex<f64>; MAX_ORDER_QCD] {
+    if order_qcd >= 4 {
+        panic!("Polarized beyond NNLO is not yet implemented");
+    }
     let mut gamma_ns = [Complex::<f64>::zero(); MAX_ORDER_QCD];
     gamma_ns[0] = as1::gamma_ns(c, nf);
     // NLO and beyond
@@ -54,6 +57,9 @@ pub fn gamma_singlet_qcd(
     nf: u8,
     _n3lo_variation: [u8; 4],
 ) -> [[[Complex<f64>; 2]; 2]; MAX_ORDER_QCD] {
+    if order_qcd >= 4 {
+        panic!("Polarized beyond NNLO is not yet implemented");
+    }
     let mut gamma_S = [[[Complex::<f64>::zero(); 2]; 2]; MAX_ORDER_QCD];
     gamma_S[0] = as1::gamma_singlet(c, nf);
     // NLO and beyond
@@ -65,4 +71,89 @@ pub fn gamma_singlet_qcd(
     //     gamma_S[2] = as3::gamma_singlet(c, nf);
     // }
     gamma_S
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::{
+        assert_approx_eq_cmplx, assert_approx_eq_cmplx_1d, assert_approx_eq_cmplx_2d, cmplx,
+    };
+    use num::complex::Complex;
+
+    #[test]
+    fn test_shapes() {
+        const NF: u8 = 4;
+        const N: Complex<f64> = cmplx!(2., 0.);
+        for order_qcd in 1..=2usize {
+            let mut c = Cache::new(N);
+            let gamma_ns = gamma_ns_qcd(order_qcd, PID_NSP, &mut c, NF, [0u8; 3]);
+            assert_eq!(gamma_ns.len(), MAX_ORDER_QCD);
+            // slots beyond order_qcd must be zero
+            for item in gamma_ns.iter().skip(order_qcd) {
+                assert_approx_eq_cmplx!(f64, *item, cmplx!(0., 0.));
+            }
+            let gamma_s = gamma_singlet_qcd(order_qcd, &mut c, NF, [0u8; 4]);
+            assert_eq!(gamma_s.len(), MAX_ORDER_QCD);
+            assert_eq!(gamma_s[0].len(), 2);
+            assert_eq!(gamma_s[0][0].len(), 2);
+            // slots beyond order_qcd must be zero
+            for item in gamma_s.iter().skip(order_qcd) {
+                assert_approx_eq_cmplx_2d!(f64, item, [[cmplx!(0., 0.); 2]; 2], 2);
+            }
+        }
+    }
+
+    #[test]
+    fn test_gamma_ns_qcd() {
+        const NF: u8 = 3;
+        const N: Complex<f64> = cmplx!(1., 0.);
+        let mut c = Cache::new(N);
+        let n3lo_variation = [0u8; 3];
+
+        // LO
+        assert_approx_eq_cmplx!(
+            f64,
+            gamma_ns_qcd(3, PID_NSP, &mut c, NF, n3lo_variation)[0],
+            cmplx!(0., 0.),
+            epsilon = 1e-14
+        );
+
+        // NLO
+        assert_approx_eq_cmplx_1d!(
+            f64,
+            gamma_ns_qcd(2, PID_NSP, &mut c, NF, n3lo_variation),
+            [cmplx!(0., 0.); 2],
+            2,
+            epsilon = 2e-6
+        );
+    }
+
+    #[test]
+    #[should_panic(expected = "Unkown non-singlet sector element")]
+    fn test_unknown_pid_panics() {
+        const NF: u8 = 4;
+        const N: Complex<f64> = cmplx!(1.234, 0.);
+        let mut c = Cache::new(N);
+        // PID_NSM_U (10202) is not a valid mode for polarized non-singlet
+        gamma_ns_qcd(2, 10202, &mut c, NF, [0u8; 3]);
+    }
+
+    #[test]
+    #[should_panic(expected = "Polarized beyond NNLO is not yet implemented")]
+    fn test_gamma_ns_order4_panics() {
+        const NF: u8 = 4;
+        const N: Complex<f64> = cmplx!(1.234, 0.);
+        let mut c = Cache::new(N);
+        gamma_ns_qcd(4, PID_NSM, &mut c, NF, [0u8; 3]);
+    }
+
+    #[test]
+    #[should_panic(expected = "Polarized beyond NNLO is not yet implemented")]
+    fn test_gamma_singlet_order4_panics() {
+        const NF: u8 = 4;
+        const N: Complex<f64> = cmplx!(2.345, 0.);
+        let mut c = Cache::new(N);
+        gamma_singlet_qcd(4, &mut c, NF, [0u8; 4]);
+    }
 }
--- a/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
+++ b/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
@@ -172,11 +172,12 @@ mod tests {
     #[test]
     #[should_panic(expected = "Unkown non-singlet sector element")]
     fn test_unknown_pid_panics() {
+        use crate::constants::PID_NSM_U;
         const NF: u8 = 4;
         const N: Complex<f64> = cmplx!(1.234, 0.);
         let mut c = Cache::new(N);
         // PID_NSM_U (10202) is not a valid mode for polarized non-singlet
-        gamma_ns_qcd(2, 10202, &mut c, NF, [0u8; 3]);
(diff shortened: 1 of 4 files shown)


AkshatRai07:
There are 3 issues/comments
- There are no `choose_ns_*` and `exp_matrix` functions on Rust side, hence tests using them are not written.
- The functions in ad_us/as4 are not correctly ported from Python, they are instead the `fhmruvv` version of the files. Hence, [N3LO tests](https://github.com/NNPDF/eko/blob/94f593aec67c010ffeedd400673a9086f7776a3e/tests/ekore/anomalous_dimensions/unpolarized/space_like/test_init.py#L118-L143) were failing and were not ported.
- It should be `[PID_NSP_U, PID_NSP_D]` [here](https://github.com/NNPDF/eko/blob/94f593aec67c010ffeedd400673a9086f7776a3e/crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs#L322), right?

felixhekhorn:
>     * There are no `choose_ns_*` and `exp_matrix` functions on Rust side, hence tests using them are not written.

- The `choose_ns_*` are inlined, e.g. [here](https://github.com/NNPDF/eko/blob/fa714f0c02bad0cfcf5e02c6725496f1d26af9ca/crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs#L129) thus we don't need to worry - also it were just `NotImplementedError` checks
- about `exp_*` we need to make a decision if we include it into `ekore` and expose it or we attribute it to `eko` and worry about it later. I suggest the second option. Reasoning: a) it is not really part of the hard math, it is rather a possible application (the relevant one for `eko`); b) they require linear algebra (exponentiation of a matrix), which is non-trivial and we don't want to worry about that now, I'd say

>     * The functions in ad_us/as4 are not correctly ported from Python, they are instead the `fhmruvv` version of the files. Hence, [N3LO tests](https://github.com/NNPDF/eko/blob/94f593aec67c010ffeedd400673a9086f7776a3e/tests/ekore/anomalous_dimensions/unpolarized/space_like/test_init.py#L118-L143) were failing and were not ported.

hm? I'm not sure I understand. Yes, the Rust implementation is directly the `fhmruvv` one and the former "main" one is now deprecated. That being said: any implementation must respect the math constraint that they vanish at $N=1$ (in this specific case) (within reasonable limits). Are you saying this is not the case? this would signal a major bug somewhere.

>     * It should be `[PID_NSP_U, PID_NSP_D]` [here](https://github.com/NNPDF/eko/blob/94f593aec67c010ffeedd400673a9086f7776a3e/crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs#L322), right?

Yes, please add a fix!

crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
```diff
@@ -19,6 +19,9 @@ pub fn gamma_ns_qcd(
     nf: u8,
     _n3lo_variation: [u8; 3],
 ) -> [Complex<f64>; MAX_ORDER_QCD] {
+    if order_qcd >= 4 {
+        panic!("Polarized beyond NNLO is not yet implemented");
+    }
```
felixhekhorn: Let's be consistent: either we add such checks everywhere or nowhere - either is fine for me. What is most idiomatic in Rust?

However, if we add them, we need to be faithful - and in this case polarized is not available beyond NLO, since NNLO is not ported from Python yet

felixhekhorn (changes requested):
While looking at the changes here I found a bug coming from #542 : OME's should have an outer dimension of `MAX_ORDER_QCD - 1`, i.e. one lower than currently - see [here](https://github.com/NNPDF/eko/blob/94f593aec67c010ffeedd400673a9086f7776a3e/crates/ekore/src/operator_matrix_elements/unpolarized/spacelike.rs#L18). This is because the "first" order ($a_s^0$) is trivially an identity and thus most people do not count that as "first" and instead keep that attribute for the first non-trivial. EKO follows that tradition.

Then there is a more conceptual problem: the main purpose of this PR is to provide a set of tests, which we can then use also in other language (first of all C of course). Now, these tests, should check the actual math, i.e. the actual implementation. However, e.g. OME/u/s tests currently only test the shape, but not the content. Instead, the content is only tested in the individual files, e.g. OME/u/s/as1, e.g. [here](https://github.com/NNPDF/eko/blob/94f593aec67c010ffeedd400673a9086f7776a3e/crates/ekore/src/operator_matrix_elements/unpolarized/spacelike/as1.rs#L87). Thus I would say we need to add to each function which we expose at least one test, where we check the actual math. I think, one is sufficient for the moment. The guideline should be: we check non-singlet quantities at $N=1$ and singlet ones at $N=2$ - this should be done already at the individual file levels so you can get inspiration from there (you will see many 0s appearing). If we want to do something more fancy, we could take a "random" N and obtain some reference numbers from there. ad/u/s already has such tests, I think, but some others don't

AkshatRai07:
> this would signal a major bug somewhere.

Running this:
```bash
source ./.venv/bin/activate && python3 -c "
import numpy as np
import ekore.anomalous_dimensions.unpolarized.space_like as ad_us
from ekore import harmonics as h
cache = h.cache.reset()
val = ad_us.as4.fhmruvv.gnsm.gamma_nsm(1+0j, 3, cache, 0)
print('fhmruvv nsm at N=1, nf=3:', val)
"
```
Results in this:
```text
fhmruvv nsm at N=1, nf=3: (-0.07190053625345172-0j)
```
The implementation in python itself is incorrect somewhere, if not then there's a problem in the paper it references (unlikely).

crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
AkshatRai07: We should add them everywhere.

AkshatRai07:
> OME's should have an outer dimension of MAX_ORDER_QCD - 1

Yes you're right. Now looking at the python side this makes sense. This is also true for `ad_ps`, max is upto 3 only.

felixhekhorn:
> > OME's should have an outer dimension of MAX_ORDER_QCD - 1
> 
> Yes you're right. Now looking at the python side this makes sense. This is also true for `ad_ps`, max is upto 3 only.

mmm let's be careful with the interpretation:
- as I tried to explain above OMEs are per se one element shorter since we discard the first trivial element - that's why I suggested MAX_ORDER_QCD - 1
- because MAX_ORDER_QCD is now defined as the maximum order available in literature - which in some cases corresponds to the maximum order currently implemented in the code, but not always
- so to which integer should the return array size correspond? the literature or the code?
- if we go for code then the global definition of MAX_ORDER_* doesn't make much sense since there is a "random" relation. Two cases are the one you mention: for both ome/u/s and ad/p/s we simply didn't have the time yet to port the stuff from Python side (or other reasons). Especially, with this reasoning the ome/u/s size should be just 2 (and not MAX_ORDER_QCD-1)

crates/ekore/src/operator_matrix_elements/unpolarized/spacelike.rs
```diff
@@ -46,3 +52,105 @@ pub fn A_non_singlet(
...
+            assert_eq!(a_ns[0][0].len(), 2);
+            // slots beyond order_qcd must be zero
+            for item in a_ns.iter().skip(matching_order_qcd) {
+                assert_approx_eq_cmplx_2d!(f64, item, [[cmplx!(0., 0.); 2]; 2], 2);
+            }
+        }
+    }
+
+    #[test]
+    fn test_a_non_singlet_math() {
```
felixhekhorn:
```suggestion
    fn test_number_conservation() {
```
this is one special case of the conservation laws you learn in your first physics class: the total number of quarks in the proton must be conserved

crates/ekore/src/operator_matrix_elements/unpolarized/spacelike.rs
```diff
@@ -46,3 +52,105 @@ pub fn A_non_singlet(
...
+        let mut c = Cache::new(N);
+        let a_ns = A_non_singlet(2, &mut c, NF, L);
+        // LO
+        assert_approx_eq_cmplx_2d!(f64, a_ns[0], [[cmplx!(0., 0.); 2]; 2], 2, epsilon = 1e-14);
+        // NNLO
+        assert_approx_eq_cmplx_2d!(f64, a_ns[1], [[cmplx!(0., 0.); 2]; 2], 2, epsilon = 1e-14);
+    }
+
+    #[test]
+    fn test_a_singlet_math() {
```
felixhekhorn:
```suggestion
    fn test_momentum_conservation() {
```
another conservation law: the momentum of the proton is the sum of the momentum of all partons; that's why you see a sum below; momentum as you learn in your first physics class

crates/ekore/src/operator_matrix_elements/unpolarized/spacelike.rs
felixhekhorn: actually now I see that the other tests don't have telling names (although they do rely on those physics features) so I guess also 
```suggestion
    fn test_a_non_singlet() {
```
would be just fine ...

crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
```diff
@@ -66,3 +72,128 @@ pub fn gamma_singlet_qcd(
...
+
+    #[test]
+    #[should_panic(expected = "Unkown non-singlet sector element")]
+    fn test_unknown_pid_panics() {
+        const NF: u8 = 4;
+        const N: Complex<f64> = cmplx!(1.234, 0.);
+        let mut c = Cache::new(N);
+        // PID_NSM_U (10202) is not a valid mode for polarized non-singlet
+        gamma_ns_qcd(2, 10202, &mut c, NF, [0u8; 3]);
```
felixhekhorn:
```suggestion
        gamma_ns_qcd(2, PID_NSM_U, &mut c, NF, [0u8; 3]);
```
maybe?

felixhekhorn:
I think apart from deciding what we want for the array sizes (which is actually a bit off topic from the original purpose of this PR, but strictly needed for #537 ), this is fine

AkshatRai07:
Any comments on the as4 issue?

AkshatRai07:
> so to which integer should the return array size correspond? the literature or the code?

Rust's coding philosophy states there should be no padding, and that is why I decreased the size of the ad/p/s by one since N3LO is not there for that case. But we have not ported NNLO yet, so the correct thing should be to decrease the size by two (i.e. MAX_ORDER_QCD - 2). We will increase this after we implement NNLO.

crates/ekore/src/operator_matrix_elements/unpolarized/spacelike.rs
```diff
@@ -46,3 +52,105 @@ pub fn A_non_singlet(
...
+        let mut c = Cache::new(N);
+        let a_ns = A_non_singlet(2, &mut c, NF, L);
+        // LO
+        assert_approx_eq_cmplx_2d!(f64, a_ns[0], [[cmplx!(0., 0.); 2]; 2], 2, epsilon = 1e-14);
+        // NNLO
+        assert_approx_eq_cmplx_2d!(f64, a_ns[1], [[cmplx!(0., 0.); 2]; 2], 2, epsilon = 1e-14);
+    }
+
+    #[test]
+    fn test_momentum_conservation() {
```
felixhekhorn:
```suggestion
    fn test_a_singlet() {
```

felixhekhorn:
> Any comments on the as4 issue?

I'm worried, but I didn't have the time yet to look into it

crates/ekore/src/operator_matrix_elements/unpolarized/spacelike.rs
felixhekhorn: Let's be consistent between the A's

crates/ekore/src/operator_matrix_elements/unpolarized/spacelike.rs
```diff
@@ -29,15 +32,18 @@ pub fn A_singlet(
 
 /// Compute the tower of the non-singlet |OME|.
 ///
-/// Returns an array of shape `(MAX_ORDER_QCD, d, d)`. Only the first `matching_order_qcd`
+/// Returns an array of shape `(MAX_ORDER_QCD - 2, d, d)`. Only the first `matching_order_qcd`
```
felixhekhorn:
```suggestion
/// Returns an array of shape `(MAX_ORDER_QCD - 2, 2, 2)`. Only the first `matching_order_qcd`
```
Let's be specific and tell what we know (here and everywhere else). (I guess this was copied from `unravel*` at some point, where `d` is indeed not known.)

Just to check I'm not crazy: `(MAX_ORDER_QCD - 2, 2, 2)` is the correct shape and NOT `(2,2,MAX_ORDER_QCD - 2)`, right? i.e. the outermost is spelled first

felixhekhorn (changes requested):

felixhekhorn (approved):

PR #550, opened 2026-07-13
Update rust/ekore documentation

felixhekhorn:
As a side effect of #543 we do no longer expose the references for our implementations. However, since we rely 100% on external sources we must do so.

### DIFF
--- a/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
+++ b/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
@@ -12,6 +12,11 @@ mod as2;
 ///
 /// Returns an array of shape `(MAX_ORDER_QCD - 2,)`. Only the first `order_qcd` entries
 /// are filled; remaining slots are zero.
+///
+/// # Available perturbative orders:
+///
+/// - $a_s^1$ [\[Moch:2004pa\]][crate::bib::Moch2004pa]
+/// - $a_s^2$ [\[Moch:2004pa\]][crate::bib::Moch2004pa]
 pub fn gamma_ns_qcd(
     order_qcd: usize,
     mode: u16,
@@ -51,6 +56,11 @@ pub fn gamma_ns_qcd(
 ///
 /// Returns an array of shape `(MAX_ORDER_QCD - 2, 2, 2)`. Only the first `order_qcd`
 /// entries along the outer axis are filled; remaining slots are zero.
+///
+/// # Available perturbative orders:
+///
+/// - $a_s^1$ [\[Gluck:1995yr\]][crate::bib::Gluck1995yr]
+/// - $a_s^2$ [\[Gluck:1995yr\]][crate::bib::Gluck1995yr] [\[Moch:2004pa\]][crate::bib::Moch2004pa]
 pub fn gamma_singlet_qcd(
     order_qcd: usize,
     cache: &mut Cache,
--- a/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
+++ b/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
@@ -17,6 +17,10 @@ mod as2;
 ///
 /// - $a_s^1$ [\[Moch:2004pa\]][crate::bib::Moch2004pa]
 /// - $a_s^2$ [\[Moch:2004pa\]][crate::bib::Moch2004pa]
+///
+/// # Panics
+///
+/// Panics if `order_qcd` is larger than the currently available order or `mode` is not a valid |PID|.
 pub fn gamma_ns_qcd(
     order_qcd: usize,
     mode: u16,
@@ -61,6 +65,10 @@ pub fn gamma_ns_qcd(
 ///
 /// - $a_s^1$ [\[Gluck:1995yr\]][crate::bib::Gluck1995yr]
 /// - $a_s^2$ [\[Gluck:1995yr\]][crate::bib::Gluck1995yr] [\[Moch:2004pa\]][crate::bib::Moch2004pa]
+///
+/// # Panics
+///
+/// Panics if `order_qcd` is larger than the currently available order.
 pub fn gamma_singlet_qcd(
     order_qcd: usize,
     cache: &mut Cache,
--- a/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
+++ b/crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
@@ -1,4 +1,4 @@
-//! The polarized, space-like anomalous dimensions at various couplings power.
+//! The polarized, space-like anomalous dimensions.
 
 use crate::constants::{MAX_ORDER_QCD, PID_NSM, PID_NSP, PID_NSV};
 use crate::harmonics::cache::Cache;
@@ -8,14 +8,14 @@ mod as1;
 mod as2;
 // pub mod as3;
 
-/// Compute the tower of the non-singlet anomalous dimensions.
+/// Compute the tower of the polarized, space-like non-singlet anomalous dimensions.
 ///
 /// Returns an array of shape `(MAX_ORDER_QCD - 2,)`. Only the first `order_qcd` entries
 /// are filled; remaining slots are zero.
 pub fn gamma_ns_qcd(
     order_qcd: usize,
     mode: u16,
-    c: &mut Cache,
+    cache: &mut Cache,
     nf: u8,
     _n3lo_variation: [u8; 3],
 ) -> [Complex<f64>; MAX_ORDER_QCD - 2] {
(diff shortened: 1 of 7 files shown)


crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs
```diff
@@ -1,4 +1,4 @@
-//! The unpolarized, space-like anomalous dimensions at various couplings power.
+//! The unpolarized, space-like anomalous dimensions.
```
felixhekhorn, line 1: Let's be short (and to the point), also we don't expose the individual files any longer

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs
```diff
@@ -15,52 +15,65 @@ pub(crate) mod as2;
 mod as3;
 mod as4;
 
-/// Compute the tower of the non-singlet anomalous dimensions.
+/// Compute the tower of the unpolarized, space-like non-singlet anomalous dimensions.
```
felixhekhorn, line 18: Specify the full path

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs
```diff
@@ -15,52 +15,65 @@ pub(crate) mod as2;
...
 ///
-/// `n3lo_variation = (ns_p, ns_m, ns_v)` is a list indicating which variation should
-/// be used. `variation = 1,2` is the upper/lower bound, while any other value
+/// This function computes the first `order_qcd` entries for the |PID| `mode` at the Mellin variable set in `cache` using `nf` light flavors.
+/// `n3lo_variation = (ns_p, ns_m, ns_v)` is a list with three entries indicating which |N3LO| variation should
+/// be used (if requested) for the given `mode`, where values `1` or `2` indicate the upper or lower bound respectively, while any other value
```
felixhekhorn, line 22: We need to mention all arguments. I also slightly rephrased the N3LO sentence, which hopefully is clearer now. This can be copied to a large degree I suspect.

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs
```diff
@@ -15,52 +15,65 @@ pub(crate) mod as2;
...
 /// Returns an array of shape `(MAX_ORDER_QCD,)`. Only the first `order_qcd` entries
 /// are filled; remaining slots are zero.
+///
+/// # Available perturbative orders:
+///
+/// - $a_s^1$ [\[Moch:2004pa\]][crate::bib::Moch2004pa] [\[Vogt:2004mw\]](crate::bib::Vogt2004mw)
+/// - $a_{em}^1$ [\[Carrazza:2015dea\]][crate::bib::Carrazza2015dea]
+/// - $a_s^2$ [\[Moch:2004pa\]][crate::bib::Moch2004pa] [\[Vogt:2004mw\]](crate::bib::Vogt2004mw)
+/// - $a_{em}^2$ [\[deFlorian:2016gvk\]][crate::bib::deFlorian2016gvk]
+/// - $a_s^3$ [\[Moch:2004pa\]][crate::bib::Moch2004pa] [\[Vogt:2004mw\]](crate::bib::Vogt2004mw)
```
felixhekhorn, line 32: Please complete the list. Copy all references used in the respective modules here (dropping the reference to equations)

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs
```diff
@@ -15,52 +15,65 @@ pub(crate) mod as2;
...
+///
+/// # Panics
+///
+/// Panics if `order_qcd` is larger than the currently available order or `mode` is not a valid |PID|.
```
felixhekhorn, line 37: This can be copied to a large degree I suspect

crates/ekore/src/anomalous_dimensions/unpolarized/spacelike.rs
```diff
@@ -15,52 +15,65 @@ pub(crate) mod as2;
...
 pub fn gamma_ns_qcd(
     order_qcd: usize,
     mode: u16,
-    c: &mut Cache,
+    cache: &mut Cache,
```
felixhekhorn, line 41: Let's use telling names

felixhekhorn (commented):
@AkshatRai07 Please complete the example function and then then propagate it to all the others in the system

felixhekhorn:
grafik

@AkshatRai07  if you have some time maybe you can also look into the `<abbr>` hack in `doc-header.html` and see why it does not work for everything (e.g. here |DGLAP| worked once, but not twice)

AkshatRai07:
One doubt, some functions in `ad_ps` call the functions in `ad_us`, so I need to add those refs of `ad_us` in `ad_ps` too right?

felixhekhorn:
> One doubt, some functions in `ad_ps` call the functions in `ad_us`, so I need to add those refs of `ad_us` in `ad_ps` too right?

ehm ... good question, but I would say no, no need to add them. I'd say the first reference is good enough and people can from there find the dependencies ... in this particular case the dependencies are known since a long time and they don't deserve special attention. In other context this can be a more complicated questions however ... (I can tell you the story at some point in person :see_no_evil: )

AkshatRai07:
Should we also write the relevant refs for `Cache` in harmonics? They are there in `g_functions` and `polygamma`

crates/ekore/src/operator_matrix_elements/unpolarized/spacelike.rs
```diff
@@ -1,18 +1,31 @@
...
 use num::complex::Complex;
 mod as1;
 mod as2;
 
-/// Compute the tower of the singlet |OME|.
+/// Compute the tower of the unpolarized, space-like singlet |OME|.
+///
+/// This function computes the first `matching_order_qcd` entries at the Mellin variable set in `cache`
+/// using `nf` light flavors and `L`, logarithm of the squared ratio of factorization and renormalization
+/// scale $\ln(\mu_F^2 / m^2)$.
```
felixhekhorn:
```suggestion
/// using `nf` light flavors and `L`, logarithm of the squared ratio of matching scale and mass $\ln(\mu^2 / m^2)$.
```
`fmt` may want to add a line break there, so please check ...

same below

crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
```diff
@@ -8,67 +8,96 @@ mod as1;
...
+///
+/// Panics if `order_qcd` is larger than the currently available order or `mode` is not a valid |PID|.
 pub fn gamma_ns_qcd(
     order_qcd: usize,
     mode: u16,
-    c: &mut Cache,
+    cache: &mut Cache,
     nf: u8,
     _n3lo_variation: [u8; 3],
 ) -> [Complex<f64>; MAX_ORDER_QCD - 2] {
```
felixhekhorn, line 32: actually, polarized will not depend in the midterm future on `_n3lo_variation` (we are even missing the implementation of N2LO at the moment) - should we drop the argument? in #537 this would break our repetition (the macros) I think since then all functions might be different again, right? so ... do we rather have a dead parameter or almost repeated code? or can we adjust the macros to account for that and get the best of both worlds?

felixhekhorn:
> Should we also write the relevant refs for `Cache` in harmonics? They are there in `g_functions` and `polygamma`

Maybe yes. Since we don't expose the individuals, maybe you can just add something like "We use implementations from [...]"

crates/ekore/src/anomalous_dimensions/polarized/spacelike.rs
AkshatRai07: >  polarized will not depend in the midterm future on _n3lo_variation

A small doubt, on the python side, `ad_ps` does not have `n3lo_variation` as a parameter, why is it here then? It's not about mid term future the function doesn't need N3LO Var at all. About the macros in `ekore_capi`, it won't be much of a issue, I tested and not much code duplication is required to handle this.

felixhekhorn: before https://github.com/NNPDF/eko/commit/64ce2fd86c6a96d05f17bfcef92eb919100d6764 the called function (e.g. `let gamma_ns_qcd = match ...`) was determined first and then afterwards called with a unique set of arguments; after that commit we call each function with its own arguments (e.g. https://github.com/NNPDF/eko/blob/597593ae17e2dd6e3f000bcf544e314999c2678c/crates/eko/src/lib.rs#L223 ), so we can drop the argument now ...

crates/ekore/src/harmonics/cache.rs
```diff
@@ -79,6 +79,9 @@ impl K {
 const CACHE_SIZE: usize = 20;
 
 /// Hold all cached values.
+///
+/// We use implementations from [\[Vogt:2004ns\]](crate::bib::Vogt2004ns) [\[KOLBIG1972221\]](crate::bib::KOLBIG1972221)
+/// [CERNlib](http://cernlib.web.cern.ch/cernlib/), name from [\[MuselliPhD\]](crate::bib::MuselliPhD)
```
felixhekhorn:
```suggestion
/// [CERNlib](http://cernlib.web.cern.ch/cernlib/)
```
the name is not publicly known right?

Issue #452, opened 2025-03-20
Caching data in workflows is not working


felixhekhorn:
> Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.

from https://github.com/NNPDF/eko/actions/runs/13971535800/job/39114365372#step:25:2

this is because this https://github.com/NNPDF/eko/blob/c69a7b964a68342cfac728627be1a2c1084287ed/.github/workflows/python-poetry-tests-with-data.yml#L82
is just wrong

This holds in all places where the cache is used:
- Python unit tests
- PyPI
- Rust unit tests

The solution will imply a series of steps, which include at least:
- [ ] make the path a parameter in the Python workflow files - for Rust it is hard coded
- [ ] do something against linking the file for Rust from Python. Currently the link is created when downloading, this will break when we only download once. Solution might be to permanently link the directory

felixhekhorn:
When resolving the data cache we can also improve the _independent_ poetry cache by adjusting the Python action- see https://github.com/actions/setup-python?tab=readme-ov-file#caching-packages-dependencies . Thanks to @alecandido for pointing out.

felixhekhorn:
Further improvement suggested by @Radonirinaunimi [in PineAPPL](https://github.com/NNPDF/pineappl/pull/344#issuecomment-2841606909)

> But I just realized that one can extract the caching logic into a separate reusable workflow and simply call:
> 
> ```
> jobs:
>   download-test-data:
>     uses: ./.github/workflows/cache-test-data.yml
> ```

felixhekhorn:
As @AkshatRai07 found out in #554 [it might be more complicated](https://github.com/NNPDF/eko/pull/554#discussion_r3863248674)

PR #554, opened 2026-07-21
Add ekore_py

AkshatRai07:
A step for #519
In this PR we:
- `ekore_py` crate, which exposes `ekore` crate using `PyO3` and build the lib using `maturin`.
- Expose anomalous_dimensions and operator_matrix_elements as Python functions.

### TODO

- [x] Expose necessary functions and constants.
- [x] Add guards.
- [x] Add tests.
- [x] Add CI.
- [x] Add CD.
- [x] Add doc-strings.
- [x] Explore removal of duplicate code.
- [x] Add Changelog entry.


AkshatRai07:
Doubt, what should we name this python library?

felixhekhorn:
> Doubt, what should we name this python library?

I guess there are two options: `ekore_py` (as in PineAPPL https://github.com/NNPDF/pineappl/tree/master/pineappl_py ) or `ekore_pyapi` (as in NeoPDF https://github.com/QCDLab/neopdf/tree/master/neopdf_pyapi) - I'm fine with either ... maybe the latter is slightly more consistent?

AkshatRai07:
The crate name will be `ekore_py` only, I was asking the name of the python library. For eg. it's the `eko` crate but the python project is `eko-rs` (hence pip install eko-rs) and library is `ekors` (hence import ekors). For `neopdf_pyapi` it's `neopdf-hep` and `neopdf`, for `pineappl_py` it is `pineappl` and `pineappl` (both are same).

felixhekhorn:
> The crate name will be `ekore_py` only, I was asking the name of the python library. For eg. it's the `eko` crate but the python project is `eko-rs` (hence pip install eko-rs) and library is `ekors` (hence import ekors). For `neopdf_pyapi` it's `neopdf-hep` and `neopdf`, for `pineappl_py` it is `pineappl` and `pineappl` (both are same).

I see, mmm ... I've no strong opinion ... `-` can not be in a Python module name of course, that's why it has to be `ekors` ... should we harmonize that (for eko) or we don't care? should we then follow that naming convention here, i.e. `pip install ekore-rs` and `import ekorers`? I think we should avoid the plain `ekore` since that is already taken by the existing eko package ... also `ekore_py` is just fine to me (e.g. also `numpy` as an explicit "py" in the name)

AkshatRai07:
So in my opinion we should not add the `-py` prefix, someone new to the ecosystem may get confused between `ekore` and `ekorepy` as which is the correct one and the difference between the two. In my opinion we should name the project `ekore-rs` and the library `ekore_rs`. This name conveys it is the Rust version of the library, and the underscore in library name makes it less confusing compared to `ekorers`. Works?

.github/workflows/unittests-pyapi.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          python-version: ${{ vars.PYTHON_VERSION_WORKFLOWS }}
+      - name: Setup Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+      - name: Create virtual environment
+        run: python -m venv .venv
+      - name: Add venv to PATH
+        run: echo "$PWD/.venv/bin" >> "$GITHUB_PATH"
+      - name: Install task runner and test deps
+        run: pip install maturin numpy poethepoet pytest
```
felixhekhorn: - do we need `venv`? can't we rely on the global environment?
- should we pin the versions here?
- if the answer to the last is yes: we should cache the environment #452

crates/ekore_py/src/macros.rs
```diff
@@ -0,0 +1,216 @@
...
+    (
+        $py:expr, $order_qcd:expr, $mode:expr, $cache:expr, $nf:expr,
+        $bound:expr, $path:path
+    ) => {{
+        gamma_ns_qcd_body!(@check $order_qcd, $mode, $bound);
+        let mut cache = $cache.borrow_mut();
+        let gamma = $path($order_qcd, $mode, &mut cache.inner, $nf);
+        gamma_ns_qcd_body!(@collect $py, $order_qcd, gamma)
+    }};
+    (@check $order_qcd:expr, $mode:expr, $bound:expr) => {
```
felixhekhorn, line 27:
```suggestion
    // sanity checks
    (@check $order_qcd:expr, $mode:expr, $bound:expr) => {
```
maybe add a few more comments, these macro expansions look a bit scary 🙃

crates/ekore_py/src/macros.rs
```diff
@@ -0,0 +1,216 @@
...
+            )));
+        }
+        if !matches!($mode, $crate::constants::PID_NSP | $crate::constants::PID_NSM | $crate::constants::PID_NSV) {
+            return Err(PyValueError::new_err(format!(
+                "invalid non-singlet mode: {}",
+                $mode
+            )));
+        }
+    };
+    (@collect $py:expr, $order_qcd:expr, $gamma:expr) => {{
```
felixhekhorn, line 42:
```suggestion
    // transform for return
    (@collect $py:expr, $order_qcd:expr, $gamma:expr) => {{
```

crates/ekore_py/src/macros.rs
```diff
@@ -0,0 +1,216 @@
+//! Shared bodies for the `#[pyfunction]`s defined throughout this crate.
+
+/// Body for a non-singlet tower function returning shape `(order_qcd,)`.
```
felixhekhorn:
```suggestion
/// Body for a non-singlet tower function.
/// Returning shape: `(order_qcd,)`.
```
maybe?

crates/ekore_py/src/macros.rs
```diff
@@ -0,0 +1,216 @@
...
+macro_rules! gamma_ns_qcd_body {
+    // With an `n3lo_variation` argument.
```
felixhekhorn:
```suggestion
    // Compute with an `n3lo_variation` argument.
```

felixhekhorn:
I checked locally and it seems to work

.github/workflows/unittests-pyapi.yml
AkshatRai07: Tried caching using the way it is written in GitHub docs, but it is creating dependency mismatch error in Ubuntu and python mismatch error in MacOS. The easier thing to do is to create a venv and then cache it, which is what I've done now.

felixhekhorn: I'm surprised that the recommended way does not work, but I guess it might be here because we have maturin in the game, which is clearly non-standard. Let's keep it at this for the moment.

felixhekhorn (approved):

PR #557, opened 2026-08-11
Fix/536 tutorial path

### LINKED ISSUE #536: `EKO.read` in tutorial broken
The tutorial, [e.g. here](https://eko.readthedocs.io/en/latest/overview/tutorials/output.html), reads
```py
with eko.EKO.read("./myeko.tar") as evolution_operator:
```
which is explicitly failing now `AttributeError: 'str' object has no attribute 'resolve'` (from [here](https://github.com/NNPDF/eko/blob/060467cd18dc7f00c8b234be0dd3ffd80db44db9/src/eko/io/struct.py#L390)), as side effect of https://github.com/NNPDF/eko/pull/449 (before we were most likely silently passing by the type check)

The solution is to use either a) enforce `pathlib.Path` everywhere in the tutorial or b) cast inside the function; a) should be preferred in my opinion ...

lol782:
Closes #536 

Before you review my PR ,I wanted to let you know that i committed twice because i accidently committed the docker file in **342bef6** and **98b5d45** is the latest commit cleaning my mistake.
As for the research what i find is that #449 tightened the type handling - previously a str was passing silently. so option (a) is the best approach rather than using 'EKO.read' this helps keep consistent with #449 and also matches with other tutorials (eg dglap.ipynb).
Please let me know  if any changes are required.

### DIFF
--- a/tests/eko/io/test_struct.py
+++ b/tests/eko/io/test_struct.py
@@ -7,7 +7,6 @@ import pytest
 import yaml
 from packaging.version import parse
 
-import eko
 from eko import EKO, interpolation
 from eko.io import struct
 from eko.io.items import Target
@@ -216,16 +215,3 @@ class TestEKO:
             and version.micro == 0
             and version.is_postrelease
         )
-    def test_read_path(self, tmp_path):
-        """Test that read can accept both str and Path for the path argument."""
-        from ekobox.cards import example
-
-        path = tmp_path / "myeko.tar"
-        eko.solve(example.theory(), example.operator(), path)
-
-        #str path
-        with EKO.read(str(path)) as eko_str:
-            assert isinstance(eko_str, EKO)
-        #Path path
-        with EKO.read(path) as eko_path:
-            assert isinstance(eko_path, EKO)
--- a/tests/eko/io/test_struct.py
+++ b/tests/eko/io/test_struct.py
@@ -7,6 +7,7 @@ import pytest
 import yaml
 from packaging.version import parse
 
+import eko
 from eko import EKO, interpolation
 from eko.io import struct
 from eko.io.items import Target
@@ -215,3 +216,16 @@ class TestEKO:
             and version.micro == 0
             and version.is_postrelease
         )
+    def test_read_path(self, tmp_path):
+        """Test that read can accept both str and Path for the path argument."""
+        from ekobox.cards import example
+
+        path = tmp_path / "myeko.tar"
+        eko.solve(example.theory(), example.operator(), path)
+
+        #str path
+        with EKO.read(str(path)) as eko_str:
+            assert isinstance(eko_str, EKO)
+        #Path path
+        with EKO.read(path) as eko_path:
+            assert isinstance(eko_path, EKO)
--- a/doc/source/overview/tutorials/output.ipynb
+++ b/doc/source/overview/tutorials/output.ipynb
@@ -19,14 +19,18 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": null,
    "id": "51dbf608-ccfb-4fb5-aca3-723a5f280f65",
    "metadata": {
     "tags": []
    },
    "outputs": [],
    "source": [
-    "import eko"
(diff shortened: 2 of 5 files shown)


felixhekhorn:
> i accidently committed the docker file in **[342bef6](https://github.com/NNPDF/eko/commit/342bef613054de71366c84e3ab851362d5b840b6)** and **[98b5d45](https://github.com/NNPDF/eko/commit/98b5d4526578817b3687cd8f6682d52d082bc4ea)** is the latest commit cleaning my mistake.

maybe you can squash the two commits, since it was a genuine git error?

According to your experience as Python developer: should we still do in addition option b) (i.e. casting)? My reasoning would be:
- CON: casting is un-pythonic a.k.a. "we're adults here" a.k.a. duck typing
- PRO: it ensures backward compatibility and makes it easier for users

lol782:
I squashed the commits thanks for the info and also you know the intent behind #449 better than I do, so I'll defer to you on that.
My take: option (a) already fixes the reported bug and since it only touches the tutorials it doesn't interfere with the strict typing from #449 at all. So (a) alone is a complete fix.
 
Whether to *also* do (b) is really a question about the engine's public contract (a metaphor)  not the tutorials:
 
- If #449 tightened the type on purpose (i.e. `EKO.read` should require a `Path`) then (a) alone is cleaner and we keep the strict contract.
- If we'd rather be user-friendly and accept path like input then adding a `Path(path)` cast in `EKO.read` matches how the stdlib treats paths (`open()` etc accept both `str` and `Path`) That's (b) and it lives inside the engine.

They don't conflict (a) fixes the docs, (b) makes the function tolerant so we could do both. But (b) means editing `src/`, so I didn't want to expand the PR without your go ahead. If you want (b), I'm happy to add the cast plus a small test in this same PR. Just let me know.

felixhekhorn:
>     * If #449 tightened the type on purpose

I think the answer to this is no - which is essentially the same as saying "backward compatibility". Thus I suggest to do also option b) However, let's do the minimal thing, so something like "if is `str` cast to `Path`"

lol782:
here's what i did,
- **(a)** `output.ipynb`: pass a `pathlib.Path` to `EKO.read` instead of
  a raw string (this one is already done).
- **(b)** `src/eko/io/struct.py`: added a minimal cast in `EKO.read` so it
  accepts both `str` and `Path` (`if isinstance(path, str): path = Path(path)`),
  and widened the type hint to `Union[str, Path]`. This keeps backward
  compatibility and matches how the stdlib treats path-like arguments.
  `EKO.edit` inherits this since it calls `read`.
- Added a test in `tests/eko/io/test_struct.py` that reads an EKO from a
  `str` path and i also attached a screen shot that test work.
image

tests/eko/io/test_struct.py
```diff
@@ -215,3 +216,16 @@ def test_version(self, tmp_path: pathlib.Path, eko_factory: EKOFactory):
...
+
+        path = tmp_path / "myeko.tar"
+        eko.solve(example.theory(), example.operator(), path)
+
+        #str path
+        with EKO.read(str(path)) as eko_str:
+            assert isinstance(eko_str, EKO)
+        #Path path
+        with EKO.read(path) as eko_path:
+            assert isinstance(eko_path, EKO)
```
felixhekhorn: actually, calling `eko.solve` is (potentially) expensive (in runtime), thus I would avoid adding an additional call here. Instead, I suggest to add the check to https://github.com/NNPDF/eko/blob/master/tests/eko/runner/test_managed.py where we already have a call

felixhekhorn (changes requested):
I think the remaining changes are just fine.

Please, remember to run pre-commit. (I think the changes in `struct.py` will pass, but the test gave you away 🙃  )

lol782:
Added a str-path assertion to the existing `test_raw` in`tests/eko/runner/test_managed.py`, reusing its `solve` call to avoid an extra (expensive) computation.

felixhekhorn (approved):
Thanks for your contribution!

lol782:
thanks for the review, happy to pick up more issues.

PR #558, opened 2026-08-22
Bug in as1aem1 splitting kernel + minor things

gstagnit:
While debugging EKO evolution against my development version of eMELA, I realised that in the gluon-initiated mixed QED-QCD splitting kernels there is an additional factor of NC, that according to eq.(29)-(31) of 1512.00612 I think should not be there. By removing it, I find closer agreement to my numbers. Then I also fixed a typo in one tutorial notebook and I think that the argument of j12 in n3lo_decompose_expanded should be beta0 and not nf.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,9 @@ Items without prefix refer to a global change.
 
 ## [Unreleased](https://github.com/NNPDF/eko/compare/v0.15.5...HEAD)
 
+### Fixed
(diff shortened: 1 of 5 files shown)


felixhekhorn (approved):
Thank you for your fixes!
- the typo `orders` -> `order`: how did that ever work? 🙈 
- $j_{12}$: a clear bug as can be seen from other invocations (thankfully that function is never used in practice)
- $N_C$ factor: comparing gg to `Eq. (31) of :cite:deFlorian:2015ujt` I can justify the 4 (coming from the normalization of both $\alpha_s$ and $\alpha_{em}$), and I think in the diagram for phg (which is just a quark loop) there is no room for any other color factor apart from $T_R$ (all the other diagrams must then have the same color).

PR #559, opened 2026-08-26
Update Architecture.md

AkshatRai07:
A step for #519.
In this PR we:
- Update the `architecture.md` file with the crates `dekoder`, `ekore_py`, and `ekore_capi`.

PR #561, opened 2026-08-28
Refactor GitHub workflows

### LINKED ISSUE #560: Improve github workflow
- Should we copy this? https://github.com/NNPDF/pineappl/commit/c710fea9ea178121a9c3c8a60bf7da42062eb4e7
- We should rename `.github/workflows/crates.yml` to `release-crates.yml` to be more consistent. Maybe even `pypi.yml` ?

AkshatRai07:
Fixes #560.
In this PR we:
- Adopted a consistent `<category>-<target>.yml` convention (`unittests-*`, `release-*`, `bench-*`):
  - `unittests.yml` $\rightarrow$ `unittests-python.yml`
  - `pypi.yml` $\rightarrow$ `release-python.yml`
  - `crates.yml` $\rightarrow$ `release-crates.yml`
  - `maturin.yml` $\rightarrow$ `release-ekors.yml`
  - `isobench.yml` $\rightarrow$ `bench-iso.yml`
  - `lha_bot.yml` $\rightarrow$ `bench-lha.yml`
  - `lha_bot_rust.yml` $\rightarrow$ `bench-lha-rust.yml`
- Enhanced release & manual publishing:
  - Restricted release tag triggers to semantic version tags (`'v[0-9]+*'`).
  - Added `workflow_dispatch` with a `tag` input across all release workflows to allow completing or retrying partially failed releases.
  - Implemented tag fallback logic (`TAG: ${{ github.event.inputs.tag || github.ref_name }}`) for version bumping and artifact packaging.
  - Updated job conditionals (`if: startsWith(github.ref, 'refs/tags/') || github.event.inputs.tag != ''`) so manual runs with a tag can publish, while tagless runs perform dry builds.
- Standardized Python environment setup with `actions/setup-python@v5` across release version-bumping steps.
- Updated  `@v2` actions to `@v4` (`actions/checkout@v4` and `codecov/codecov-action@v4`) in benchmark workflows.
- Pointed the CI test badge in `README.md` to `unittests-python.yml`.


scarlehoff:
Perhaps this is a good time to change to trusted publishing for eko (instead of creating a new token for the new package).

For now I added the trusted publisher pointing to `release-python.yml` to the `eko` package https://docs.pypi.org/trusted-publishers/using-a-publisher/

 I've also pre-created the `ekore-rs` package pointing instead to `release-ekors.yml` so the package should be created automagically when you try to deploy it.

.github/workflows/release-crates.yml
```diff
@@ -18,9 +26,11 @@ jobs:
       - name: Install task runner
         run: pip install poethepoet
       - name: Bump versions
+        env:
+          TAG: ${{ github.event.inputs.tag || github.ref_name }}
         run: |
           poetry install --only version
-          poe bump-version
+          python crates/bump-versions.py ${{ env.TAG }}
```
felixhekhorn: I guess you changed this to make it more consistent across workflows, right? But wouldn't it make sense to do it the other way round, i.e. make everybody use `poe`? the idea being that this is potentially a complicated task (and not just a simple Python call). However, if you disagree, then we should drop the command from poe to have a unique command.

.github/workflows/release-python.yml
```diff
@@ -0,0 +1,25 @@
...
+          Existing release tag to (re)publish. Use this to complete a release
+          whose tag-triggered run failed partway.
+        required: false
+
+jobs:
+  publish:
+    if: startsWith(github.ref, 'refs/tags/') || github.event.inputs.tag != ''
+    uses: ./.github/workflows/python-poetry-pypi-with-data.yml
+    secrets:
+      PYPI_TOKEN: ${{ secrets.PYPI_TOKEN }}
```
felixhekhorn, line 21: If I understand OIDC correctly, this can be dropped as a consequence, right?

felixhekhorn (commented):
thanks for doing a more extensive review of the workflow files

.github/workflows/release-python.yml
scarlehoff: Indeed.

.github/workflows/release-crates.yml
felixhekhorn: propagate the decision, if necessary, also to `Architecture.md` (e.g. in #559 )

felixhekhorn:
Also, remember to update https://github.com/NNPDF/eko/blob/f131c1ba4b67fa7689c1474655e64b90925717ce/extras/architecture.md?plain=1#L370

.github/workflows/release-crates.yml
AkshatRai07: I have reverted it to be `poe bump-version` in all the files.

AkshatRai07:
> I've also pre-created the ekore-rs package pointing instead to release-ekors.yml

There's an issue in this.
- `release-ekors.yml` builds `crates/eko/Cargo.toml`, which publishes `eko-rs`.
- `release-pyapi.yml` builds `crates/ekore_py/Cargo.toml`, which publishes `ekore-rs`.

Please update the permissions respectively.

scarlehoff:
Ah, my bad, 

> release-pyapi.yml builds crates/ekore_py/Cargo.toml, which publishes ekore-rs.

Now `ekore-rs` points to `release-pyapi.yml`

felixhekhorn (approved):

Issue #371, opened 2024-05-15
Couplings method is called with array instead of float


felixhekhorn:
Can you please have a look to the deprecation warning from numpy [here](https://github.com/NNPDF/eko/actions/runs/9096504564/job/25002064814#step:11:134)

specifically all the
```
/home/runner/work/eko/eko/src/eko/couplings.py:691: DeprecationWarning: Conversion of an array with ndim > 0 to a scalar is deprecated, and will error in future. Ensure you extract a single element from your array before performing this operation. (Deprecated NumPy 1.25.)
    key = (float(a_ref[0]), float(a_ref[1]), nf, nl, scale_from, float(scale_to))
```
look very suspicious - someone is calling the method with `scale_to` not being a float, which is clearly wrong

It might be a badly written test.

PR #563, opened 2026-08-31
read metadata and cards from tar without extracting multi-dim arrays

### LINKED ISSUE #533: Interface to extract only the *.yaml metadata
So that the whole eko doesn't need to be extracted.

lol782:
closes #533

Implements the cheap card-reading we discussed in #533.

## Approach (following your instructions )
- Added `from_raw` to `TheoryCard`, `OperatorCard`, and `Metadata` (in runcards.py / metadata.py). Each takes raw yaml, applies the v1/v2 upgrades, then calls `from_dict`. This is the upgrade logic lifted out of the `EKO.theory_card` / `EKO.operator_card` properties.
- Added `load_meta_and_cards_from_tar(path)` in runcards.py: opens the archive and pulls only metadata.yaml / theory.yaml / operator.yaml via `tarfile.extractfile` — the operators are never extracted. Order is metadata → theory → operator (operator upgrade needs the theory dict).
- Refactored the two EKO properties to reuse `from_raw`, so folder-reading and tar-reading share one code path.

## Points I'd like your view on
1. **Metadata legacy upgrade.** `v1/v2.update_metadata` needs the folder (`paths`), which the tar path doesn't have. So I kept that legacy 0.13/0.14 upgrade inside `Metadata.load` and made `from_raw` do the plain build. Consequence: reading a *very old* EKO straight from tar won't auto-upgrade its metadata. Fine for modern files — is that an acceptable boundary, or do you want it handled differently?
2. **Function placement.** Put `load_meta_and_cards_from_tar` in runcards.py as you suggested. 
3. **Test depth.** Added a test that reads the cards from a real archive and checks types + a few values. Let me know if you want stronger assertions.

Existing io tests pass locally and below is the new one.
Screenshot 2026-08-31 180517

## AI disclosure
[see below](https://github.com/NNPDF/eko/pull/563#issuecomment-5684160747)


src/eko/io/metadata.py
```diff
@@ -77,6 +77,11 @@ def load(cls, path: os.PathLike):
         content = cls.from_dict(raw)
         content._path = path
         return content
+    
+    @classmethod
+    def from_raw(cls, raw: dict) -> "Metadata":
```
felixhekhorn, line 82: > Metadata legacy upgrade

yeah, seems that the `update_metadata` s take `paths` - but if you look to the function body they don't do anything with it so I'd say we can just remove them from there, because we definitely need patches here

src/eko/io/runcards.py
```diff
@@ -343,3 +367,32 @@ def masses(theory: TheoryCard, evmeth: EvolutionMethod) -> List[SquaredScale]:
...
+    """
+    for member in tar.getmembers():
+        if pathlib.Path(member.name).name == filename:
+            extracted = tar.extractfile(member)
+            if extracted is not None:
+                return yaml.safe_load(extracted.read())
+    raise KeyError(f"'{filename}' not found in {tar.name}")
+
+
+def load_meta_and_cards_from_tar(path):
```
felixhekhorn, line 384: > Function placement

fine for me - any clever suggestion for a shorter function name? 🙃  being explicit is also fine

tests/eko/io/test_struct.py
```diff
@@ -197,6 +199,19 @@ def test_load_opened(self, tmp_path: pathlib.Path, eko_factory: EKOFactory):
 
         assert read_closed.metadata == read_opened.metadata
 
+    def test_load_meta_and_cards_from_tar(self, eko_factory: EKOFactory):
+        """Load metadata and both YAML cards from a real EKO archive."""
+        eko = eko_factory.get()
```
felixhekhorn, line 204: > Test depth

this is only the current data version - can we please add also all the older as in https://github.com/NNPDF/eko/blob/ab85bf7144a51489a2dfdea975d1ef92f154cce9/tests/eko/io/test_legacy.py#L22 ? Consider putting this list into a fixture so we can reuse it here and there

src/eko/io/runcards.py
```diff
@@ -169,6 +184,15 @@ def evolgrid(self) -> List[EPoint]:
     def pids(self):
         """Internal flavor basis, used for computation."""
         return np.array(br.flavor_basis_pids)
+    
+    @classmethod
+    def from_raw(cls, raw: dict, data_version: int, theory_raw)-> "OperatorCard":
+        """Build an operator card from a raw dictionary."""
+        if data_version == 1:
+            raw =v1.update_operator(raw)
```
felixhekhorn, line 192: since you are not testing on the full set of versions you did not realize that this is wrong - look:
https://github.com/NNPDF/eko/blob/ab85bf7144a51489a2dfdea975d1ef92f154cce9/src/eko/io/v1.py#L65
there is a reason I put the argument there 🙃

felixhekhorn (changes requested):
Remember to run pre-commit

src/eko/io/runcards.py
```diff
@@ -343,3 +367,32 @@ def masses(theory: TheoryCard, evmeth: EvolutionMethod) -> List[SquaredScale]:
         return [mq.value**2 for mq in theory.heavy.masses]
 
     raise ValueError(f"Unknown mass scheme '{theory.heavy.masses_scheme}'")
+
+def _read_yaml_member(tar: tarfile.TarFile, filename: str) -> dict:
+    """Extract and parse a single yaml file from the archive by name.
+
+    Only this member's bytes are read; nothing else is extracted.
+    """
+    for member in tar.getmembers():
```
scarlehoff, line 376: Is this "costless". If it is, then this is fine.

Otherwise, this function could take a list of filenames so that one can extract all of the files we need in one go.

src/eko/io/runcards.py
scarlehoff, line 384:
```suggestion
def load_meta_and_cards_from_tar(eko_path):
```

if we are being explicit, better to be explicit here (so that it is clear to the user that this is the "eko path"

lol782:
fix the required changes, Apologies for not remembering pre-commit.

- Operator card bug — from_raw wasn't forwarding the theory dict to v1/v2.update_operator. Fixed (this was the empty-operator failure on legacy versions).
- Metadata legacy upgrade — removed the unused paths argument from update_metadata in v1.py/v2.py, and moved the version patching into Metadata.from_raw (which was where the legacy case was failing). load now just reads the file and calls from_raw.
- Tests — added a fixture with the legacy + current archives (v1-0.13.tar, v1-0.14.tar, v3.tar) so both test_legacy.py and the new test reuse it. The new function is now tested across all versions, not just current. Both pass.
- Runcards — Changed the function logic to improve performance. Previously, it scanned the member list up to three times; now, the number of scans is hardcoded to three.

and also changed the name from **load_meta_and_cards_from_tar → read_eko_cards** and for test function name too.

src/eko/io/metadata.py
```diff
@@ -69,18 +69,25 @@ def load(cls, path: os.PathLike):
         # patch if necessary
         if data_version == 1:
             if version.major == 0 and version.minor == 13:
-                raw = v1.update_metadata(paths, raw)
+                raw = v1.update_metadata(raw)
             elif version.major == 0 and version.minor == 14:
-                raw = v2.update_metadata(paths, raw)
+                raw = v2.update_metadata(raw)
```
felixhekhorn, line 74: We want to reuse the new function of course

src/eko/io/runcards.py
```diff
@@ -343,3 +367,37 @@ def masses(theory: TheoryCard, evmeth: EvolutionMethod) -> List[SquaredScale]:
         return [mq.value**2 for mq in theory.heavy.masses]
 
     raise ValueError(f"Unknown mass scheme '{theory.heavy.masses_scheme}'")
+
+
+def _read_yaml_members(tar, filenames):
```
felixhekhorn: In the last commit you removed types, e.g. `-> dict`, but please keep them (and consider adding more)

tests/conftest.py
```diff
@@ -68,6 +68,11 @@ def operator_card():
     return card
 
 
+@pytest.fixture
+def legacy_eko_filenames():
+    return ["v1-0.13.tar", "v1-0.14.tar", "v3.tar"]
```
felixhekhorn: 1. The "legacy" in the name is not correct since it contains both legacy and current
2. rather indicate (some way) that they are actual real objects (and not mocked)
3. of course the path should be here (otherwise it is repeated)

felixhekhorn (changes requested):
- Please mark conversation which you believe to have addressed as resolved, so we know more easily which problems still need to be addressed
- A simple way for me (and everybody else 🙃 ) to see you are human and not a bot is by seeing you thinking. For example https://github.com/NNPDF/eko/pull/563/changes/e2b63665f6b381cbd0da994688ff060e747a2c19 could have been 4 commits - doing in turn what you describe [here](https://github.com/NNPDF/eko/pull/563#issuecomment-5553963672). Making small self-contained commits is not easy and I openly admit I'm not doing it myself all the time, but I try - and this way you can see me thinking.
- this will also help with writing more informative commit messages as "fix the required changes" [is not very helpful in the long run](https://chris.beams.io/git-commit)

lol782:
wait you're still thinking i am a bot 🥲.

felixhekhorn:
> wait you're still thinking i am a bot 🥲.

nono, of course not - but I want to help you to avoid that impression

PS: and to teach you how to write better commits along the way :see_no_evil:

src/eko/io/metadata.py
lol782: can you explain it more? i don't understand what you mean ,we are using the updated function here its not new 🦆?right?

lol782:
Sorry for the mess! I forgot to run pre-commit before pushing those changes.

lol782:
Also added the type hints for the newly added functions ,
The only thing left is now is the 1st change ,can you please elaborate it if i guessed it you want me to make me merge the Metadata.from_raw  and the Metadata.load😵‍💫?

src/eko/io/metadata.py
```diff
@@ -69,15 +69,27 @@ def load(cls, path: os.PathLike):
...
 
         # now we are ready
         content = cls.from_dict(raw)
```
felixhekhorn:
```suggestion
        content = cls.from_raw(raw)
```

> merge the Metadata.from_raw and the Metadata.load

well, more reuse as I said 🙃  I guess this is a classic case of "miss the forest for the trees" 😇 don't worry

felixhekhorn (changes requested):
last missing item: please add an entry to the Changelog

src/eko/io/metadata.py
felixhekhorn: ehm the github interface does not show properly what I meant, so just to be sure: replace line 67-77 with a single call (<- look what a nice simplification)

lol782: wow thanks for the explanation i was indeed looking at trees 🤣.

lol782:
You probably saw it 😅 here's what happened, so you know the story:
I got to the changelog and noticed mine differed from the one on GitHub. Turned out my master had never been updated since I forked, so I added an upstream remote and rebased onto it. That's when I learned rebasing rewrites the commit hashes, so a normal push was rejected ,I used --force-with-lease to update the branch, which I gather is the expected step after a rebase.

End result is the six separate commits above, one per change, plus the changelog entry. Learned a fair bit about _**rebase/upstream**_ in the process 🙂. Happy to adjust anything,but I'd lean toward keeping them as separate commits since that matches what you asked for earlier. Let me know.

src/eko/io/metadata.py
```diff
@@ -64,19 +64,21 @@ def load(cls, path: os.PathLike):
         paths = InternalPaths(path)
         # read raw file first to catch version
         raw = yaml.safe_load(paths.metadata.read_text(encoding="utf-8"))
+        content = cls.from_raw(raw)
+        content._path = path
+        return content
+
+    @classmethod
+    def from_raw(cls, raw: dict) -> "Metadata":
```
felixhekhorn, line 72:
```suggestion
    def from_raw(cls, raw: dict) -> Self:
```
we can use [`Self`](https://docs.python.org/3/library/typing.html#typing.Self) here since a) we are already [requiring py3.11](https://github.com/NNPDF/eko/blob/8ca897dfb897e39ee84dd581f49cb21b00a0034a/pyproject.toml#L43) and b) we return an instance of `cls` and not `Metadata` explicitly (see also [PEP 673](https://peps.python.org/pep-0673/#use-in-classmethod-signatures)).

change here and also `runcards.py`

felixhekhorn (changes requested):
Sorry another minor thing 🙃 

aaand I'm learning something myself by reading PEP673 more carefully and understanding that `Self` is not always correct, but sometimes you need the literal name (however, not here, if I understood correctly 🙈 )

felixhekhorn:
> keeping them as separate commits since that matches what you asked for earlier.

Indeed, I prefer separate commits since this way you can see the problem evolve and you can discard or retain individual steps if necessary

felixhekhorn:
In case this applies to you: note https://github.com/NNPDF/eko/commit/e5d00508c4eea77f6fb9d7460640dd96dc62608b

lol782:
> In case this applies to you: note [e5d0050](https://github.com/NNPDF/eko/commit/e5d00508c4eea77f6fb9d7460640dd96dc62608b)

Sure, I used Claude , via its chat interface, throughout this PR to understand the code and the issue, discuss the design, draft and refine code, and help debug. I reviewed, ran, and tested all changes myself and take full responsibility for the final contribution 😊.

lol782:
By the way since its coming to close, I’d love to keep the momentum going! so can you suggest what i should do next ? 🫠

felixhekhorn (approved):
Thanks for your work!

felixhekhorn:
> By the way since its coming to close, I’d love to keep the momentum going! so can you suggest what i should do next ? 🫠

first I'm happy to hear you like to continue :+1: then we can think of a new project for you. With thinking only 5 min I can come up with two proposals, but which are both not as "simple" as this. Please tell me if you like either of them and in case neither (which is also fine) we can think more than 5 min for another proposal.
1. Resolve https://github.com/NNPDF/eko/issues/371 - this has slightly more physics involved, but maybe you can also do without ...
2. looking to your profile and repos again I can see that you have not been in contact with Rust so far - would you like to get involved? We could try and see if there is a way to replicate this exact PR on the rusty side of eko, a.k.a. [`dekoder`](https://github.com/NNPDF/eko/tree/master/crates/dekoder). This proposal is clearly more speculative than 1.

Let me know what you think.

lol782:
Thanks — both look interesting, but I'd like to take you up on option 2. I haven't touched Rust before, so this feels like a good chance to get into it, especially since I already understand the problem from the Python PR ,I'd mainly be learning how it's expressed in Rust rather than what to build.

But since it's new to me, I'll need a few days first to read through dekoder and understand what the Rust side is actually doing before I can say whether the replication is feasible. I'll treat it as "explore and report back" .😙

felixhekhorn:
fine :+1: however, keep in mind that the `dekoder` crate is very much still in a proof of concept state and has not much features yet (only the most basic) and e.g. there are no metadata, theory card or operator card structs (yet). Thus what I would want to know as a first step (which I guess you can find out by reading the suitable documentation) is, if there is a way in Rust to implement the core of this PR: i.e. extract just a given file without extracting all the rest. If the answer is yes, we can in this first step just extract the raw yaml without any casting to a struct and without any patching for backward compatibility (both would come later). Keep this first big question in mind.

felixhekhorn:
> if there is a way in Rust to implement the core of this PR: i.e. extract just a given file without extracting all the rest.

> you can find out by reading the suitable documentation

I got distracted while cleaning stuff and found out the answer myself: yes, with https://docs.rs/tar/latest/tar/struct.Entry.html#method.unpack_in . So @lol782 if you wish (!), I can give you a more concrete task: can you provide a short PR which adds a function `read_eko_cards` to the `dekoder` crate, which returns three strings? (we can cast to yaml in a second step)

> I'll need a few days first to read through dekoder and understand what the Rust side is actually doing

Aaand I haven't forgotten that. Please take your time.

lol782:
>can you provide a short PR which adds a function `read_eko_cards` to the `dekoder` crate, which returns three strings? 

Sorry for the late reply! I've been diving into Rust and haven't had a chance to look through **_dekoder_** yet, but I'd be happy to give this a try.

lol782:
By the way, I have a question why you guys use the Rust here is it because of its fast nature or resemblance to python(syntax wise) 🐵?

felixhekhorn:
> By the way, I have a question why you guys use the Rust here is it because of its fast nature or resemblance to python(syntax wise) 🐵?

neither, but more pragmatic :upside_down_face: one of our colleagues wrote the [PineAPPL](https://github.com/NNPDF/pineappl) library a few years ago in Rust for his own reasons (one of them being he was tired of C++ :see_no_evil: ). This introduced Rust into the NNPDF universe and so we just followed that choice. And indeed, eventually, `dekoder` will be the interface between eko and PineAPPL.

scarlehoff:
The fact that it can be interfaced easily both to python and c/fortran + packaged into pypi compatible wheels also helped immensely.

PR #570, opened 2026-09-09
Pass plain ints instead of enums to numba kernels

### LINKED ISSUE #524: Use of `Enum` and `numba`
@kamillaurent reports:
```
2026-05-28 03:20:21,201 eko.evolution_operator/INFO: Evolution: computing operators - 193/196 took: 0.909104 s
2026-05-28 03:20:22,105 eko.evolution_operator/INFO: Evolution: computing operators - 194/196 took: 0.903717 s
2026-05-28 03:20:23,390 eko.evolution_operator/INFO: Evolution: computing operators - 195/196 took: 0.897556 s
2026-05-28 03:20:24,109 eko.evolution_operator/INFO: Evolution: computing operators - 196/196 took: 0.717515 s
2026-05-28 03:20:25,765 eko.evolution_operator/INFO: Evolution: Total time 4963.983533 s
[... after some time ...]
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/bin/evolven3fit", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/data/theorie/klaurent/nnpdf/n3fit/src/n3fit/scripts/evolven3fit.py", line 187, in main
    solve(tcard, opcard, args.dump)
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/eko/runner/managed.py", line 29, in solve
    eko.parts[recipe] = parts.evolve(eko, recipe)
                        ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/eko/runner/parts.py", line 76, in evolve
    op.compute()
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/eko/evolution_operator/__init__.py", line 981, in compute
    self.integrate()
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/eko/evolution_operator/__init__.py", line 1001, in integrate
    for j, row in enumerate(res):
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/eko/evolution_operator/__init__.py", line 902, in run_op_integration
    res = integrate.quad(
          ^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/scipy/integrate/_quadpack_py.py", line 479, in quad
    retval = _quad(func, a, b, args, full_output, epsabs, epsrel, limit,
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/scipy/integrate/_quadpack_py.py", line 626, in _quad
    return _quadpack._qagse(func,a,b,args,full_output,epsabs,epsrel,limit)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/numba/core/dispatcher.py", line 687, in typeof_pyval
    tp = typeof(val, Purpose.argument)
         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/numba/core/typing/typeof.py", line 33, in typeof
    ty = typeof_impl(val, c)
         ^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/numba/core/typing/typeof.py", line 230, in _typeof_enum
    clsty = typeof_impl(type(val), c)
            ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/functools.py", line 909, in wrapper
    return dispatch(args[0].__class__)(*args, **kw)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/numba/core/typing/typeof.py", line 249, in _typeof_enum_class
    return typecls(cls, dtypes.pop())
           ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/numba/core/types/abstract.py", line 68, in __call__
    return cls._intern(inst)
           ^^^^^^^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/numba/core/types/abstract.py", line 57, in _intern
    inst._code = _autoincr()
                 ^^^^^^^^^^^
  File "/data/theorie/klaurent/miniconda3/envs/nnpdf-dev/lib/python3.12/site-packages/numba/core/types/abstract.py", line 23, in _autoincr
    assert n < 2 ** 32, "Limited to 4 billion types"
AssertionError: Limited to 4 billion types
```
The problem seems to be (according to Claude run by @juanrojochacon )
```
The trigger is `_typeof_enum` → `_typeof_enum_class` in the chain: every time EKO calls a Numba-jitted integrand through `scipy.integrate.quad`, Numba inspects the arguments and, for an `Enum` argument, **creates a brand-new `EnumClass` type and interns it**. Numba does not cache enum types by identity properly here, so each call leaks a fresh type into the global registry. EKO's `run_op_integration` runs `quad` an enormous number of times (per grid point, per operator element, per Q² step), so the counter climbs until it overflows 2³².
```
and a supposed solutions (suggested by the same pair of "people")
```
The cleanest fix is to stop passing the enum into the jitted function and pass its `.value` (a plain int) instead, converting back inside or branching outside the hot loop
```

ashishsoni-ai:
### Summary

Fixes #524.

Every `scipy.integrate.quad` call type-inspects the arguments of the jitted integration kernel. For `IntEnum` arguments numba goes through `_typeof_enum` → `_typeof_enum_class` → `IntEnumClass(cls, dtype)` and **interns a fresh enum type in the global type registry each time** (its dispatcher signature cache does not prevent the registry growth). EKO's integration hot loops invoke the kernel once per integrand evaluation, so the numba type counter `_typecodes` grows without bound and a long evolution eventually trips:

```
File "numba/core/types/abstract.py", line 23, in _autoincr
    assert n < 2 ** 32, "Limited to 4 billion types"
AssertionError: Limited to 4 billion types
```

as reported by @kamillaurent.

### What this PR does

- Convert at the boundary, at the `functools.partial` call sites:
  - `Operator.quad_ker` (`ev_method`, `sv_mode`)
  - `OperatorMatrixElement.quad_ker` (`sv_mode`, `backward_method`)
  now pass `int(...)` of `EvoMethods` / `sv.Modes` / `MatchingMethods` members into the jitted kernels.
- Replace the two `is` comparisons inside jitted code with `==`:
  - `build_ome` (`MatchingMethods`)
  - `ns.dispatcher` (`EvoMethods.ORDERED_TRUNCATED`)
  
  This is required for the int contract to be sound: under `njit`, `x is SomeEnumMember` silently evaluates to `False` for a plain int argument (verified with numba 0.65.1), while `x == SomeEnumMember` compares by value as intended.
- Update the kernel docstrings to document the plain-int contract.
- Add regression tests:
  - call-site contract: the `quad_ker` partials pass plain ints, not `enum.Enum` instances
  - plain-int arguments are accepted end-to-end by `quad_ker`

### Verification

- Instrumented `numba.core.types.abstract.Type._intern`: 5000 int-arg calls → 1871 intern calls (bounded baseline), 5000 enum-arg calls → 10007 (~2 fresh types per call, linear growth). With the fix, the enum arguments never cross the boundary, so the growth is gone.
- All kernels / evolution_operator / runner test suites pass (numba 0.65.1, Python 3.12), and a JIT-enabled smoke test exercised the `BACKWARD_EXPANDED` and `BACKWARD_EXACT` branches of `build_ome` plus `ns.dispatcher` with plain ints, matching enum results exactly.

### AI usage disclosure

This PR was implemented with the assistance of an AI coding assistant
(Codebuff, harness) using LLMs (z-ai/glm-5.3-flash and
deepseek/deepseek-v4-flash) for code and text drafting. The author
reviewed, tested and verified all changes and takes full responsibility
for the final contribution.

### DIFF
--- a/src/eko/evolution_operator/__init__.py
+++ b/src/eko/evolution_operator/__init__.py
@@ -251,7 +251,10 @@ class Operator(sv.ScaleVariationModeMixin):
             order=self.order,
             mode0=label[0],
             mode1=label[1],
-            ev_method=self.ev_method,
+            # pass plain ints across the numba boundary, to avoid leaking
+            # fresh enum types into the numba type registry at each call
+            # (see https://github.com/NNPDF/eko/issues/524)
+            ev_method=int(self.ev_method),
             is_log=self.int_disp.log,
             logx=logx,
             areas=areas,
@@ -264,7 +267,7 @@ class Operator(sv.ScaleVariationModeMixin):
             Lsv=np.log(self.xif2),
             ev_op_iterations=self.config["ev_op_iterations"],
             ev_op_max_order=tuple(self.config["ev_op_max_order"]),
-            sv_mode=self.sv_mode,
+            sv_mode=int(self.sv_mode),
             is_threshold=self.is_threshold,
             n3lo_ad_variation=self.config["n3lo_ad_variation"],
             is_polarized=self.config["polarized"],
--- a/src/eko/evolution_operator/__init__.py
+++ b/src/eko/evolution_operator/__init__.py
@@ -251,9 +251,6 @@ class Operator(sv.ScaleVariationModeMixin):
             order=self.order,
             mode0=label[0],
             mode1=label[1],
-            # pass plain ints across the numba boundary, to avoid leaking
-            # fresh enum types into the numba type registry at each call
-            # (see https://github.com/NNPDF/eko/issues/524)
             ev_method=int(self.ev_method),
             is_log=self.int_disp.log,
             logx=logx,
--- a/src/eko/evolution_operator/operator_matrix_element.py
+++ b/src/eko/evolution_operator/operator_matrix_element.py
@@ -163,9 +163,12 @@ class OperatorMatrixElement(Operator):
             a_s=self.a_s,
             nf=self.nf,
             L=self.L,
-            sv_mode=self.sv_mode,
+            # pass plain int across the numba boundary, to avoid leaking
+            # a fresh enum type into the numba type registry at each call
+            # (see https://github.com/NNPDF/eko/issues/524)
+            sv_mode=int(self.sv_mode),
             Lsv=np.log(self.xif2),
-            backward_method=self.backward_method,
+            backward_method=int(self.backward_method),
             is_msbar=self.is_msbar,
             is_polarized=self.config["polarized"],
             is_time_like=self.config["time_like"],
--- a/src/eko/evolution_operator/operator_matrix_element.py
+++ b/src/eko/evolution_operator/operator_matrix_element.py
@@ -163,9 +163,6 @@ class OperatorMatrixElement(Operator):
             a_s=self.a_s,
             nf=self.nf,
             L=self.L,
-            # pass plain int across the numba boundary, to avoid leaking
-            # a fresh enum type into the numba type registry at each call
-            # (see https://github.com/NNPDF/eko/issues/524)
             sv_mode=int(self.sv_mode),
             Lsv=np.log(self.xif2),
             backward_method=int(self.backward_method),
--- a/src/eko/evolution_operator/quad_ker.py
+++ b/src/eko/evolution_operator/quad_ker.py
@@ -129,8 +129,11 @@ def build_ome(A, matching_order, a_s, backward_method):
         perturbation matching order
     a_s : float
         strong coupling, needed only for the exact inverse
-    backward_method : MatchingMethods
-        empty or method for inverting the matching condition (exact or expanded)
+    backward_method : int
+        method for inverting the matching condition (exact or expanded),
+        as the plain int value of the corresponding :class:`MatchingMethods`
+        member (passing the member itself is still supported, but plain ints
+        avoid leaking enum types into the numba type registry)
 
     Returns
     -------
@@ -147,7 +150,7 @@ def build_ome(A, matching_order, a_s, backward_method):
     ome = np.eye(len(A[0]), dtype=np.complex128)
     A = A[:, :, :]
     A = np.ascontiguousarray(A)
-    if backward_method is MatchingMethods.BACKWARD_EXPANDED:
+    if backward_method == MatchingMethods.BACKWARD_EXPANDED:
         # expended inverse
         if matching_order[0] >= 1:
             ome -= a_s * A[0]
@@ -164,7 +167,7 @@ def build_ome(A, matching_order, a_s, backward_method):
         if matching_order[0] >= 3:
             ome += a_s**3 * A[2]
         # need inverse exact ?  so add the missing pieces
-        if backward_method is MatchingMethods.BACKWARD_EXACT:
+        if backward_method == MatchingMethods.BACKWARD_EXACT:
             ome = np.linalg.inv(ome)
     return ome
 
@@ -278,8 +281,9 @@ def quad_ker_ad(
         pid for first sector element
     mode1 : int
         pid for second sector element
-    ev_method : str
-        ev_method
+    ev_method : int
+        evolution method, as the plain int value of the corresponding
+        :class:`eko.kernels.EvoMethods` member
     is_log : boolean
         is a logarithmic interpolation
     logx : float
@@ -304,8 +308,9 @@ def quad_ker_ad(
         number of evolution steps
     ev_op_max_order : int
         perturbative expansion order of U
-    sv_mode: int, `enum.IntEnum`
-        scale variation mode, see `eko.scale_variations.Modes`
+    sv_mode : int
+        scale variation mode, as the plain int value of the corresponding
+        :class:`eko.scale_variations.Modes` member
     is_threshold : boolean
         is this an intermediate threshold operator?
     n3lo_ad_variation : tuple
@@ -404,8 +409,9 @@ def quad_ker_qcd(
         pid for first sector element
     mode1 : int
         pid for second sector element
-    ev_method : str
-        ev_method
+    ev_method : int
+        evolution method, as the plain int value of the corresponding
+        :class:`eko.kernels.EvoMethods` member
     as1 : float
         target coupling value
     as0 : float
@@ -418,8 +424,9 @@ def quad_ker_qcd(
         number of evolution steps
     ev_op_max_order : int
         perturbative expansion order of U
-    sv_mode: int, `enum.IntEnum`
-        scale variation mode, see `eko.scale_variations.Modes`
+    sv_mode : int
+        scale variation mode, as the plain int value of the corresponding
+        :class:`eko.scale_variations.Modes` member
     is_threshold : boolean
         is this an itermediate threshold operator?
     n3lo_ad_variation : tuple
@@ -528,8 +535,9 @@ def quad_ker_qed(
         pid for first sector element
     mode1 : int
         pid for second sector element
-    ev_method : str
-        ev_method
+    ev_method : int
+        evolution method, as the plain int value of the corresponding
+        :class:`eko.kernels.EvoMethods` member
     as1 : float
         target coupling value
     as0 : float
@@ -550,8 +558,9 @@ def quad_ker_qed(
         number of evolution steps
     ev_op_max_order : int
         perturbative expansion order of U
-    sv_mode: int, `enum.IntEnum`
-        scale variation mode, see `eko.scale_variations.Modes`
+    sv_mode : int
+        scale variation mode, as the plain int value of the corresponding
+        :class:`eko.scale_variations.Modes` member
     is_threshold : boolean
         is this an itermediate threshold operator?
     n3lo_ad_variation : tuple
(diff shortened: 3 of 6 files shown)


felixhekhorn:
This repository will adhere to the AI policy of NNPDF https://github.com/NNPDF/nnpdf/pull/2512 (note to myself: remember to put this in a prominent place). So please
- rebase the commit such that you are the only committer declaring full responsibility
- document above if and how you used AI

ashishsoni-ai:
Hi @felixhekhorn, thanks for pointing to the LLM guidelines (NNPDF/nnpdf#2512) — both points are addressed:

1. **Sole committer declaring full responsibility**: the branch was rebased to a single commit `64bf7237` authored and committed by `ashishtransfirm` only; no co-author trailer remains.
2. **AI usage documented in the PR**: I updated the *AI usage disclosure* section above to name the harness (Codebuff) and the LLMs used (z-ai/glm-5.3-flash, deepseek/deepseek-v4-flash).

felixhekhorn:
> 1. **Sole committer declaring full responsibility**: the branch was rebased to a single commit `64bf7237` authored and committed by `ashishtransfirm` only; no co-author trailer remains.

Why is the author @ashishtransfirm and not @ashishsoni-ai ? We care about responsibility and being able to track changes. Now, the former account is (quasi) a bot, where as the latter looks more reasonable like a human. Naively I expect the former to be abandoned at any time, where as the latter hopefully not. It is impossible to see that the two accounts belong (hopefully) together - there is no way to find that out. Please change it. (Also please remember to answer my questions on the issue.)

felixhekhorn:
@lol782 can you please take a look at these changes and tell me if they make sense?

lol782:
> @lol782 can you please take a look at these changes and tell me if they make sense?

sure, i will try my best.

src/eko/evolution_operator/__init__.py
```diff
@@ -251,7 +251,10 @@ def quad_ker(self, label, logx, areas):
             order=self.order,
             mode0=label[0],
             mode1=label[1],
-            ev_method=self.ev_method,
+            # pass plain ints across the numba boundary, to avoid leaking
+            # fresh enum types into the numba type registry at each call
+            # (see https://github.com/NNPDF/eko/issues/524)
+            ev_method=int(self.ev_method),
```
felixhekhorn:
```suggestion
            ev_method=int(self.ev_method), # cast to int - see https://github.com/NNPDF/eko/issues/524
```

src/eko/evolution_operator/operator_matrix_element.py
```diff
@@ -163,9 +163,12 @@ def quad_ker(self, label, logx, areas):
             a_s=self.a_s,
             nf=self.nf,
             L=self.L,
-            sv_mode=self.sv_mode,
+            # pass plain int across the numba boundary, to avoid leaking
+            # a fresh enum type into the numba type registry at each call
+            # (see https://github.com/NNPDF/eko/issues/524)
+            sv_mode=int(self.sv_mode),
```
felixhekhorn:
```suggestion
            sv_mode=int(self.sv_mode), # cast to int - see https://github.com/NNPDF/eko/issues/524
```

src/eko/evolution_operator/quad_ker.py
```diff
@@ -129,8 +129,11 @@ def build_ome(A, matching_order, a_s, backward_method):
         perturbation matching order
     a_s : float
         strong coupling, needed only for the exact inverse
-    backward_method : MatchingMethods
-        empty or method for inverting the matching condition (exact or expanded)
+    backward_method : int
+        method for inverting the matching condition (exact or expanded),
+        as the plain int value of the corresponding :class:`MatchingMethods`
+        member (passing the member itself is still supported, but plain ints
+        avoid leaking enum types into the numba type registry)
```
felixhekhorn:
```suggestion
        member
```
We don't want to encourage wrong behavior

tests/eko/evolution_operator/test_ome.py
```diff
@@ -71,6 +73,29 @@ def test_build_ome_nlo():
...
+    o = 1
+    aNS = A_non_singlet((o, 0), N, nf, L)
+    aS = A_singlet((o, 0), N, nf, L, is_msbar)
+    for a in [aNS, aS]:
+        for method in MatchingMethods:
+            ome_enum = build_ome(a, (o, 0), a_s, method)
+            ome_int = build_ome(a, (o, 0), a_s, int(method))
+            np.testing.assert_allclose(ome_enum, ome_int, err_msg=method)
+
+
```
felixhekhorn: As said above: let's not encourage the use of enum if we want int - so this should be deleted, right?

felixhekhorn (commented):
I'm fine with the provided solution, so genuine question: what about [`IntEnum`](https://docs.python.org/3/library/enum.html#enum.IntEnum) ? would it work the same as solution or has it the same problem?

felixhekhorn:
> I'm fine with the provided solution, so genuine question: what about [`IntEnum`](https://docs.python.org/3/library/enum.html#enum.IntEnum) ? would it work the same as solution or has it the same problem?

Sorry, reading #524 I can answer myself: the problem remains. (right?) I should have read the issue before reviewing here, my bad.

lol782:
hey @felixhekhorn , so i gone through the PR and you're thoughts too i would say this is a pretty different area from anything I've touched, so correct me if am wrong and here's my understanding 🦆:

The bug is that passing the mode enums (ev_method, sv_mode, backward_method) into the jitted kernels makes numba intern a fresh enum type on each call numba keys on the argument's type, and an enum member's type is the enum class, so it leaks. Casting to int() unwraps it to a plain int (a type numba already knows), so nothing new gets interned. The is→== changes follow because the kernels now receive ints, and int is EnumMember would be False while == compares by value. That all makes sense to me somehow.

On the tests, as I read them: one asserts the quad_ker partials actually carry ints and not enums (conversion happens), one runs the full EvoMethods × Modes product and checks enum vs int give identical results (fix doesn't change the math), and build_ome is checked with an int method (the is→== swap). The monkeypatching is just stubbing the heavy physics functions so the kernel plumbing can run without a full evolution.

Two things I'd raise, in line with your comments: the docstrings saying enums are "still supported" work against the int contract, so I'd drop that wording; and the enum-looping test similarly encourages the thing we're moving away from. 

On your IntEnum question , I think it'd hit the same problem, since an IntEnum member's type is still the enum class, so numba would intern it the same way.
i know you already knows these things but what is it you need me to do here 🤔?

felixhekhorn:
@lol782 my hope was that [since you also wanted to work on the issue](https://github.com/NNPDF/eko/issues/524#issuecomment-5370000741) you would tell me "yeah, I would have done the same" or maybe not. Anyway thanks for the review, it seems we all agree this is ok.

felixhekhorn:
@ashishsoni-ai do you agree with [my (mostly) cosmetic suggestions above?](https://github.com/NNPDF/eko/pull/570#pullrequestreview-5168396437) If yes, let's apply them and then we can merge.

lol782:
> @lol782 my hope was that [since you also wanted to work on the issue](https://github.com/NNPDF/eko/issues/524#issuecomment-5370000741) you would tell me "yeah, I would have done the same" or maybe not. Anyway thanks for the review, it seems we all agree this is ok.

Honestly, I don't think I'd have gotten there the same way I wouldn't have known numba interns a type per enum, so I'd have been stuck at the traceback for a while before working back to the enum boundary. Reading this taught me the mechanism.
Given what I know now, the int approach is the one that makes sense to me I can't think of a cleaner way that doesn't touch numba itself 😓.

felixhekhorn:
> Reading this taught me the mechanism.

you learned something - so this is what you wanted, no? :innocent: this is why we work together. As one of my professors always says: if you know the solution, everything is easy.

PS: this is also why I asked to @ashishsoni-ai to explain more

ashishsoni-ai:
Yes, absolutely. I agree with your suggestions and have applied them in `8783dea2` and pushed the changes to the branch.

I’ve removed the redundant comments, updated the docstring so it doesn’t encourage enum inputs, and removed the tests that were encouraging enum usage. The actual regression tests and the plain-int boundary fix are still intact.

Thanks for the review and for pointing these things out 🙂

felixhekhorn (approved):

